Issue #27, opened 2019-10-22
Add CI/CD


RiccardoM:
Currently there is no CI/CD setup into the repo. We should use CircleCI as well as add proper unit tests

PR #29, opened 2019-10-23
Keeper improvements

### LINKED ISSUE #19: Double return value in Keeper methods
Currently there are a bunch of methods inside `keeper.go` that have double return values.  
As examples, we can take a look at

https://github.com/desmos-labs/desmos/blob/18c63f5e7f4b61ac43b705c70a789a12b2dee698/x/magpie/keeper.go#L30

https://github.com/desmos-labs/desmos/blob/18c63f5e7f4b61ac43b705c70a789a12b2dee698/x/magpie/keeper.go#L53

https://github.com/desmos-labs/desmos/blob/18c63f5e7f4b61ac43b705c70a789a12b2dee698/x/magpie/keeper.go#L84

I personally think that having a double return value with a `bool` value that is `true` when there was no error is useless. It's probably better and more efficient to have a single `error` value which is `nil` if everything went ok and then simply having to check that value without having to do the double `err != nil && success` check.

### LINKED ISSUE #20: Improving the getter methods inside the Keeper
Currently the `GetX` methods inside the `keeper.go` file have a single return value. As examples we can take a look at 

https://github.com/desmos-labs/desmos/blob/18c63f5e7f4b61ac43b705c70a789a12b2dee698/x/magpie/keeper.go#L41

https://github.com/desmos-labs/desmos/blob/18c63f5e7f4b61ac43b705c70a789a12b2dee698/x/magpie/keeper.go#L102

I personally think that in this case having a single value may make the "not found" situation hard to handle. 

Let's consider the first case (`GetPost`). As this returns a default `Post` when none is found, what should we check to make sure that is was not found indeed? 

I think that having a double return value such as the following might be more useful: 

```golang
func (k Keeper) GetPost(ctx sdk.Context, id string) (post Post, found bool) {
    // ...
} 
```

Having a method like this one, we could then perform a simple check like this one 

```golang 
if post, found := keeper.GetPost(ctx, id); found {
  // do something with post 
}
```

Or something like this 

```golang 
if !found {
  // throw an error 
}

// do something with post
```

RiccardoM:
## Changes
- Removed double return values from the Keeper setters (fixes #19) 
- Added double return values to the Keeper getters (fixes #20) 
- Added some Keeper tests 
- Added the Circle CI configuration (works towards #27)


kwunyeung:
tAck

PR #34, opened 2019-10-29
Split the posts and magpie module apart

### LINKED ISSUE #20: Improving the getter methods inside the Keeper

RiccardoM:
## Description
Split the `magpie` module into `magpie` and `posts` (closes #20)

## Checklist
- [ ] Wrote tests
- [ ] Re-reviewed `Files changed` in the Github PR explorer


x/magpie/client/cli/tx.go
```diff
@@ -19,124 +19,24 @@ import (
...
-		},
-	}
-}
-
 // GetCmdCreateSession is the CLI command for creating a session for create post
 func GetCmdCreateSession(cdc *codec.Codec) *cobra.Command {
 	return &cobra.Command{
 		Use:   "create-session [namespace] [external address] [pubkey] [external signer signature]",
-		Short: "record a session for external service to post a magpie",
+		Short: "Creates a session for an external service to post a magpie",
```
kwunyeung: Should not read "post a magpie" here as now `Magpie` will just be the bridge. Maybe just `Creates a session for an external service to post`

x/magpie/internal/keeper/handler.go
```diff
@@ -162,33 +38,50 @@ func handleMsgCreateSession(ctx sdk.Context, keeper Keeper, msg types.MsgCreateS
...
 	copy(pkBytes33[:], pkBytes)
 	pubkey := secp256k1.PubKeySecp256k1(pkBytes33)
 
-	message := fmt.Sprintf(`{"account_number":"0","chain_id":"%s","fee":{"amount":[],"gas":"200000"},"memo":"","msgs":[{"type":"desmos/MsgCreateSession","value":{"created":"%s","external_owner":"%s","namespace":"%s","owner":"%s","pubkey":"%s","signature":null}}],"sequence":"0"}`,
-		ctx.ChainID(), msg.Created.Format(time.RFC3339Nano), msg.ExternalOwner, msg.Namespace, msg.Owner.String(), msg.Pubkey)
+	// Create the StdSignDoc by using the given message data, with an empty string
+	signedMsg := msg
+	signedMsg.Signature = ""
+
+	stdSignDoc := auth.StdSignDoc{
```
kwunyeung: Looks good. Would like to know if test pass.

x/posts/client/cli/query.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/kwunyeung/desmos/x/posts/internal/types"
+	"github.com/spf13/cobra"
+)
+
+// GetQueryCmd adds the query commands
+func GetQueryCmd(storeKey string, cdc *codec.Codec) *cobra.Command {
+	magpieQueryCmd := &cobra.Command{
```
kwunyeung: May use a proper variable name in `Post` module.

x/posts/client/cli/tx.go
```diff
@@ -0,0 +1,131 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/kwunyeung/desmos/x/posts/internal/types"
+)
+
+// GetTxCmd set the tx commands
+func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
+	magpieTxCmd := &cobra.Command{
```
kwunyeung: Same for variable name.

x/magpie/internal/types/msgs_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+var testOwner, _ = sdk.AccAddressFromBech32("cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns")
+var timeZone, _ = time.LoadLocation("UTC")
+var msgShareDocumentSchema = types.MsgCreateSession{
+	Owner:         testOwner,
+	Created:       time.Date(2019, 10, 31, 9, 42, 0, 0, timeZone),
+	Namespace:     "cosmos",
+	ExternalOwner: "cosmos1njrqah832yfdv8yhxnrskerzxhj5zj9e563uge",
+	PubKey:        "cosmospub1addwnpepqf06gxm8tf4u9af99zsuphr2jmqvr2t956me5rcx9kywmrtg6jewy8gjtcs",
+	Signature:     "QmZh...===",
```
kwunyeung: Does this work?

kwunyeung:
`go mode vendor`

RiccardoM:
> will test soon

It's still WIP. I want to add some more tests and later fix the issues before tagging it R4R. I'll request your review when done. 

> `go mod vendor`

I avoid using `vendor` and prefer `download` instead. Will fix the problems soon btw

RiccardoM:
Closing this as it's getting very dirty

PR #37, opened 2019-11-06
Various improvements

### LINKED ISSUE #23: Magpie Session created and expiration should be replaced with block height
The current demo is generating a created time and expiry time on each session. This can't be verified and not able to make consensus. It has to be changed to block height.

### LINKED ISSUE #26: Split the magpie module
Currently all the features are present inside the `magpie` module. I personally think that we should split those features into different modules. They might be the following ones:

- `magpie` for cross-chain sessions
- `posts` for posts and likes
- `discussions` for discussions proposed into #13 
- `polls` for polls proposed into #14

RiccardoM:
## Description
- Split the `magpie` module into `magpie` and `posts` (closes #26)
- Replaced the timestamps with block heights inside post creation time, session creation time and like time (closes #23)

## Checklist
- [x] Wrote tests


x/magpie/internal/keeper/handler.go
```diff
@@ -162,33 +37,48 @@ func handleMsgCreateSession(ctx sdk.Context, keeper Keeper, msg types.MsgCreateS
...
 	}
 
+	// Create the testSession
 	session := types.Session{
-		SessionID:     keeper.GetLastSessionId(ctx).Next(),
-		Created:       msg.Created,
-		Expiry:        msg.Created.Add(time.Minute * 14400),
+		SessionID:     keeper.GetLastSessionID(ctx).Next(),
+		Created:       ctx.BlockHeight(),
+		Expiry:        ctx.BlockHeight() + 240, // 24 hours, counting a 6 secs block interval
```
kwunyeung: May put this in genesis params later.

x/posts/internal/keeper/handler.go
```diff
@@ -0,0 +1,172 @@
...
+	return func(ctx sdk.Context, msg sdk.Msg) sdk.Result {
+		switch msg := msg.(type) {
+		case types.MsgCreatePost:
+			return handleMsgCreatePost(ctx, keeper, msg)
+		case types.MsgEditPost:
+			return handleMsgEditPost(ctx, keeper, msg)
+		case types.MsgLikePost:
+			return handleMsgLike(ctx, keeper, msg)
+		default:
+			errMsg := fmt.Sprintf("Unrecognized Magpie Msg type: %v", msg.Type())
```
kwunyeung: This may have to be updated.

x/magpie/internal/keeper/querier.go
```diff
@@ -5,89 +5,46 @@ import (
...
-	"github.com/kwunyeung/desmos/x/magpie/internal/types"
+	"github.com/desmos-labs/desmos/x/magpie/internal/types"
 	abci "github.com/tendermint/tendermint/abci/types"
 )
 
 // query endpoints supported by the magpie Querier
 const (
-	QueryPost = "post"
-	QueryLike = "like"
+	QuerySession = "testSession"
```
kwunyeung: It looks this would give error when query `sessions/` endpoint.

x/posts/internal/types/post.go
```diff
@@ -43,12 +42,11 @@ func ParsePostId(value string) (PostId, error) {
...
-	ParentID      PostId         `json:"parent_id"`
+	PostID        PostID         `json:"id"`
+	ParentID      PostID         `json:"parent_id"`
 	Message       string         `json:"message"`
-	Created       time.Time      `json:"created"`
-	Modified      time.Time      `json:"modified"`
-	Likes         uint           `json:"likes"`
+	Created       int64          `json:"created"`     // Block height at which the post has been created
+	LastEdited    int64          `json:"last_edited"` // Block height at which the post has been edited the last time
 	Owner         sdk.AccAddress `json:"owner"`
```
kwunyeung: The `Likes` is deleted here. It was there to have a quick reference on how many likes have been casted to the post. If it doesn't exist, would it be easy to get this number? I doubt. LevelDB is simple key value pair data structure and seems these aggregation feature doesn't exist?

x/magpie/internal/keeper/handler.go
RiccardoM: Opened #38 to have a better discussion there,

x/posts/internal/keeper/handler.go
RiccardoM: Done

x/magpie/internal/keeper/querier.go
RiccardoM: Fixed, thanks

x/posts/internal/types/post.go
RiccardoM: The problem is that currently we are storing likes in the wrong way. 
Right now when a Like object is saved inside the local database, it is saved using it's id as the key as the whole object itself as the value.  

I personally think we should store the different likes for a post using the post id as the key and the array of likes as the value.   
In this way we could easily retrieve the array of likes that a post has associated to itself (having the id of the post) and count how many likes the post has. This could be done inside the keeper's `GetPost` method or, even better, inside the `querier` method that handles the reading of posts or a specific post's details. 

Also, with this new method of storing likes when a user wants to unlike a post we will have to go through just the likes objects associated with that post and find the one for that user.  
The same would happen when a user tries to like a post and we need to make sure he hasn't liked it yet. 
Right now, those operations are performed in `O(n)` where `n` is the overall number of likes to all the posts. With the new method they could be performed in `O(m)` with `m` being the number of the likes to the specific post. Since we have `m <= n` for each `n` chosen, I think we should change how we store the likes in that way. 

Even when retrieving the number of likes per post we can do it in `O(1)` as we have the post id and can easily get the array of likes, so even in this case we do not have any particular problem.

kwunyeung: Make sense. You are right. Actually I did the same way in a previous webapp using MongoDB before. I have overlooked this issue here.

PR #40, opened 2019-11-15
Post and Like structure revist

### LINKED ISSUE #39: Avoid double liking
**Describe the bug**
Currently when storing the like of a user to a post, we do not check if the user has already liked the post itself. This leads to the possibility of double-liking. 

**To Reproduce**
1. Create a post
2. Like the post 
3. Like the post again

**Expected behavior**
An error should be returned when trying to like the post twice.

RiccardoM:
## Description
This PR implements various `Post` and `Like` structure improvements. 
Also, the various messages have been revisited in order to delete useless time fields. 
Tests have been added and documentation has been written.

Closes #39 

## Checklist
- [x] Wrote tests

### DIFF
--- a/x/magpie/internal/keeper/handler.go
+++ b/x/magpie/internal/keeper/handler.go
@@ -82,7 +82,13 @@ func handleMsgCreateSession(ctx sdk.Context, keeper Keeper, msg types.MsgCreateS
 		Signature:     msg.Signature,
 	}
 
-	if err := keeper.CreateSession(ctx, session); err != nil {
+	// Check for any previously existing session
+	if _, found := keeper.GetSession(ctx, session.SessionID); found {
+		return sdk.ErrUnknownRequest(fmt.Sprintf("Session with id %s already exists", session.SessionID)).Result()
+	}
+
+	// Save the session
+	if err := keeper.SaveSession(ctx, session); err != nil {
 		return err.Result()
 	}
 
--- a/x/posts/internal/keeper/keeper.go
+++ b/x/posts/internal/keeper/keeper.go
@@ -68,7 +68,7 @@ func (k Keeper) GetPost(ctx sdk.Context, id types.PostID) (post types.Post, foun
 
 	key := k.getPostStoreKey(id)
 	if !store.Has(key) {
-		return types.NewPost(), false
+		return types.Post{}, false
 	}
 
 	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &post)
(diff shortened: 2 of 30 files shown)


kwunyeung:
Let's target merging it this weekend and launch testnet.

x/magpie/internal/keeper/handler.go
```diff
@@ -82,7 +82,13 @@ func handleMsgCreateSession(ctx sdk.Context, keeper Keeper, msg types.MsgCreateS
 		Signature:     msg.Signature,
 	}
 
-	if err := keeper.CreateSession(ctx, session); err != nil {
+	// Check for any previously existing session
+	if _, found := keeper.GetSession(ctx, session.SessionID); found {
+		return sdk.ErrUnknownRequest(fmt.Sprintf("Session with id %s already exists", session.SessionID)).Result()
```
kwunyeung: Not a big deal but looks strange if a function is named `GetSession` where the function is to check whether the session exists.

x/posts/internal/keeper/keeper.go
```diff
@@ -104,84 +88,42 @@ func (k Keeper) GetPosts(ctx sdk.Context) []types.Post {
...
-	like.PostID = post.PostID
+	// Get the existent likes
+	var likes types.Likes
+	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &likes)
 
-	// Store the like and update the last like id
-	if err := k.SaveLike(ctx, like); err != nil {
-		return err
+	// Check for double likes
+	if likes.ContainsOwnerLike(like.Owner) {
```
kwunyeung: Can this be done in `ValidateBasic()` or `ValidateBasic()` can't access this data?

kwunyeung: Possible to query the state before broadcasting the message?

RiccardoM: Unluckily, not. The `ValidateBasic` method has no access to the state.

On the other hand, we can move this to the handler. Do you think it would be better?

x/magpie/internal/keeper/handler.go
RiccardoM: The `GetSession` here is used to check if a session already exists, but elsewhere (i.e inside the `querier`) the same method is used to get a session using its ID.

Here we use the second returned value, which allows for faster and easier comparison than the first one. It's just a convinient way to keep the code DRY and avoiding having multiple methods with very similar functionality

PR #59, opened 2019-12-17
Posts improvements

### LINKED ISSUE #46: Support for post subspaces and topics
## Context
Ideally, what we want to achieve with Desmos is to enable any developer to create a messaging-based application that is completely decentralized and censorship resistant. 

To do this, we need to find a way to distinguish the messages of an application from the ones of other applications built on Desmos. 

## Subspace implementation proposal 
To solve this problem, we can implement the concept of *subspace* and *topic*. 

A **subspace** will identify all the messages coming from the same application. For example, all messages coming from BigDipper will have `subspace = "bigdipper"`. 

A **topic**, on the other hand, allows applications to arbitrarily decide the topic of a specific post. For example, inside the `bigdipper` namespace we might want to have `proposal-discussion`, `private-message` and other topics that let the clients easily fetch the messages related to a specific discussion subject. 

### Storage implementation 
The post object should be changed adding two new fields: 

```golang
type Post struct {
    PostID     PostID         `json:"id"`
    ParentID   PostID         `json:"parent_id"`
    Message    string         `json:"message"`
    Created    int64          `json:"created"`
    LastEdited int64          `json:"last_edited"`
    Owner      sdk.AccAddress `json:"owner"`

    Subspace   string         `json:"subspace"`
    Topic      string         `json:"topic"`
}
```

The following data should also be stored:

- The associated list of post IDs that are part of a specific `subspace`.
- The list of `topic`s associated to a specific `subspace` 
- The list of post IDs that are part of a specific `topic`. 

In this way we will be able to fetch in constant time the following data: 

- The list of messages for a given subspace
- The list of topics for a given subspace
- The list of messages for a given topic
- The subspace of a post 
- The topic of a post

### LINKED ISSUE #47: Support for post reactions
## Context 
In the era of current social networks, more and more applications are allowing users to add more than simple likes to posts. As an example, we've seen Facebook reactions: 

screenshot-www facebook com-2019 11

Another one is the ability on Slack to react with emojis: 

Screenshot from 2019-11-20 11-33-16

I think we should support this feature too as it make Desmos even more attractive to applications developers.

## Post reaction implementation proposal
Currently we represent the like object using the following data structure: 

```golang
type Like struct {
    Created sdk.Int        `json:"created"` // Block height at which the like was created
    Owner   sdk.AccAddress `json:"owner"`
}
```

I propose to expand and rename this structure to support other reaction types as follows: 

```golang
type Reaction struct {
    Created sdk.Int        `json:"created"`
    Owner   sdk.AccAddress `json:"owner"`
    Value   string         `json:"value"`
}
```

The only change is the new `Value` field which allows application to insert any kind of text inside the reaction. This might be a simple `like` string, or even emojis. 

### Storage changes
No storage changes should be required.

### LINKED ISSUE #52: Support for arbitrary data
## Context
As Desmos will be a public chain potentially used by many different applications as their backend storage, we might want to allow such applications to store arbitrary data inside posts. 

Some examples can be the following. 

* BigDipper might want to store a reference to the chain-id for which the message has been posted and/or the address of the original poster;
* applications having a local storage might want to store a locally-generated reference to a post in order to easily update its status when its uploaded to the chain. 

## Arbitrary data support proposal
In order to solve both issues in one single run, I suggest we implement the support for arbitrary data by adding a new field into the `Post` type.

For sake of simplicity I recommend we only support `String -> String` associations, and not any king of value type that might be used inappropriately. 

With this, any developer can integrate whatever data he wants into the post itself, without having to add any overhead to the overall post structure. 

### Storage implementation
In order to implement such feature, we can change how `Post` objects are made, adding a new `map[string]string` field: 

```golang
type Post struct {
    PostID            PostID            `json:"id"`
    ParentID          PostID            `json:"parent_id"`
    Message           string            `json:"message"`
    Created           sdk.Int           `json:"created"`
    LastEdited        sdk.Int           `json:"last_edited"`
    AllowsComments    bool              `json:"allows_comments"`
    ExternalReference string            `json:"external_reference"`
    Owner             sdk.AccAddress    `json:"owner"`

    OptionalData      map[string]string `json:"optional_data"` // Contains any kind of arbitrary data
}
```
I also think that we might want to move the existing `ExternalReference` field into such map to make the overall object more simple and having less fields leading thus to a smaller size on disk: 

```golang
type Post struct {
    PostID            PostID            `json:"id"`
    ParentID          PostID            `json:"parent_id"`
    Message           string            `json:"message"`
    Created           sdk.Int           `json:"created"`
    LastEdited        sdk.Int           `json:"last_edited"`
    AllowsComments    bool              `json:"allows_comments"`
    Owner             sdk.AccAddress    `json:"owner"`

    OptionalData      map[string]string `json:"optional_data"` // Contains any kind of arbitrary data
}
```

### LINKED ISSUE #66: Limit the number and length of arbitrary data
## Context
With #52 we introduced the ability to put arbitrary data inside each post. This is particularly useful to developers when wanting to attach any external data that might be later helpful to a post. 

However, while highly useful, it also might become a problem if some attacker starts using this field to put crazily long data inside it to increase the chain storage enormously. 

## Solution
What I have thought is to limit the **number** of arbitrary data that can be stored per post as well as the **length** of each data stored. The limitations I have thought are:

- **maximum 10 fields** per post 
- **maximum 200 chars** per field

This will lead to at most 2000 bytes-length arbitrary data fields, which IMO will be sufficient for most developers as well as reasonably low to store on chain. 

In the future, if needed, we can increase both limits independently. 

What do you guys think? @kwunyeung @bragaz

RiccardoM:
## Description
- Implemented the support for arbitrary data inside a post (closes #52) 
- Implemented the support for posts reactions (closes #47) 
- Implemented the support for posts subspaces (closes #46)
- Implemented max arbitrary data length (closes #66)
- Implemented the migration to v0.2.0 about posts and reactions

## Checklist
- [x] Wrote tests


x/posts/internal/types/reaction.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ContainsReactionFrom returns true if the reactions slice contain
+// a reaction from the given user having the given value, false otherwise
+func (reactions Reactions) ContainsReactionFrom(user sdk.AccAddress, value string) bool {
+	return reactions.IndexOfByUserAndValue(user, value) != -1
+}
+
+// IndexOfByUserAndValue returns the index of the reaction from the
+// given user with the specified value inside the reactions slice.
+func (reactions Reactions) IndexOfByUserAndValue(owner sdk.AccAddress, value string) int {
```
golangcibot: `owner` can be `github.com/cosmos/cosmos-sdk/types.Address` (from `interfacer`)

x/posts/client/rest/rest.go
```diff
@@ -18,19 +18,20 @@ import (
 func RegisterRoutes(cliCtx context.CLIContext, r *mux.Router, storeName string) {
 	r.HandleFunc("/posts", createPostHandler(cliCtx)).Methods("POST")
 	r.HandleFunc("/posts/{postID}", getPostHandler(cliCtx, storeName)).Methods("GET")
-	r.HandleFunc("/posts/likes", likePostHandler(cliCtx)).Methods("POST")
+	r.HandleFunc("/posts/reactions", addReactionToPostHandler(cliCtx)).Methods("POST")
 	r.HandleFunc("/posts/likes", unlikePostHandler(cliCtx)).Methods("DELETE")
```
kwunyeung: Should we change this endpoint to `/reactions` as well?

RiccardoM: Thanks, also renamed the handler

x/posts/internal/keeper/post_response.go
```diff
@@ -10,15 +10,15 @@ import (
...
 type PostQueryResponse struct {
 	types.Post
-	Likes    types.Likes   `json:"likes"`
-	Children types.PostIDs `json:"children"`
+	Reactions types.Reactions `json:"reactions"`
+	Children  types.PostIDs   `json:"children"`
 }
 
-func NewPostResponse(post types.Post, likes types.Likes, children types.PostIDs) PostQueryResponse {
+func NewPostResponse(post types.Post, reactions types.Reactions, children types.PostIDs) PostQueryResponse {
```
kwunyeung: I guess the query from CLI is not getting this post response as it's still querying the post only.

https://github.com/desmos-labs/desmos/blob/4f43ef8ae3dfea83de9bbeeba3c8cd3fbc43683b/x/posts/client/cli/query.go#L39

x/posts/internal/types/msgs.go
```diff
@@ -41,8 +53,24 @@ func (msg MsgCreatePost) ValidateBasic() sdk.Error {
...
+
+	if len(strings.TrimSpace(msg.Message)) == 0 {
+		return sdk.ErrUnknownRequest("Post message cannot be empty nor blank")
+	}
+
+	if len(strings.TrimSpace(msg.Subspace)) == 0 {
+		return sdk.ErrUnknownRequest("Post subspace cannot be empty nor blank")
+	}
+
+	if len(msg.OptionalData) > 10 {
```
kwunyeung: Better store the limits in constants rather than hardcoding here.

x/posts/internal/types/post_response_test.go
```diff
@@ -1,12 +1,11 @@
-package keeper_test
+package types_test
 
 import (
 	"encoding/json"
+	"github.com/desmos-labs/desmos/x/posts/internal/types"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

x/posts/internal/types/msgs.go
RiccardoM: Done

x/posts/internal/keeper/post_response.go
RiccardoM: Fixed this

Issue #14, opened 2019-09-16
Support for polls


kwunyeung:
`Poll` is very common in communications. A `Poll` will have poll options and aggregated results.

RiccardoM:
## Context
If we browse some of the most known poll websites such as [EasyPolls.net](https://www.easypolls.net/), [Doodle](https://doodle.com/) and [StrawPoll](https://www.strawpoll.me/) we can see that all of them have the same concept of _poll_. 

By abstracting the common parts, we can see that a _poll_ to be defined such, must: 

- Have one question in a textual form 
- Provide two or more answers to the question, in a textual form
- Allow (or not) the user to select multiple answers (decided during the creation of the poll)

## Poll Post implementation proposal
After seeing what are the requirements of a _poll_, here is my implementation proposal for a `PollPost` type. 

```golang
// PollPost is a post that represents a poll
type PollPost struct {
    StdPost
    Open                  bool         `json:"open"`                    // Tells if the poll is still accepting answers
    EndDate               int64        `json:"end_date"`                // Block height at which the poll will no longer accept new answers
    ProvidedAnswers       []PollAnswer `json:"answers"`                 // Lists of answers provided by the creator
    AllowsMultipleAnswers bool         `json:"allows_multiple_answers"` // Tells if the poll is a single or multiple answers one
    AllowsAnswerEdits     bool         `json:"allows_answer_edits"`     // Tells whether the user can edit an answer he already inserted
}

// PollAnswer contains the data of a single poll answer inserted by the creator
type PollAnswer struct {
    Id   uint64 `json:"id"`   // Unique id inside the post
    Text string `json:"text"` // Text of the answer
}
```

Other than implementing all the requirements above, by adding the `EndDate` field we can provide the creator the ability to specify a time range after which the poll will be automatically be closed and not accept any more answers. 

### Operations to provide
The operations to provide to the users are the following. 

1. Create a poll  
   Requires to input the text of the poll, the answers to be provided, whether or not it accepts multiple answers and whether or not it allows answer edits.

2. Close a poll  
   Requires to provide a `types.PostID` and sets the `Open` field to `false` 

3. Answer to a poll
   Requires to provide a `types.PostID` and a slice of `uint64` representing the answers ids. It adds the answers to the poll or edits any existing ones.  
   If the poll does not accept multiple answers from the same user, the slice must be of length one.  
   If the poll does not accept edits and the user tries to override his answer, an error must be thrown.

### Storage
[...]

PR #72, opened 2020-01-08
Media Post Implementation

### LINKED ISSUE #36: Support for media posts
Propose adding a `Media` and `MediaType` field in `Post` module for storing a media file in a post. Similar to an image of a Twitter tweet or an image post of Instagram.

This field is optional. 
If it exists, validate if the `Media` value is a valid url. `MediaType` should be an option from a list of mime types.

bragaz:
## Description
This PR implements `MediaPost` and relative methods. 

TODO:
- [x] Implement types;
- [x] Implement message;
- [x] Update codecs;
- [x] Wrote CLI commands.

## Checklist
- [x] Wrote tests.


RiccardoM:
@bragaz I've noticed that currently you've implemented the `Provider` support and `MimeType` checks. 

I personally think we should follow what @kwunyeung suggested inside his [comment](https://github.com/desmos-labs/desmos/issues/36#issuecomment-555832136) on #72: 

> We cannot and not necessary to verify the mime types on-chain. It depends on how the applications work and specify it as the uploading and rendering parts won't be done on-chain. 

And 

> [...] Storing the mime types are for the applications to know how the files should be rendered. So I don't think the Provider is necessary in any cases.

This means that there is no need for on-chain verification of mime types verification nor provider support at all. 

Do you think they should be kept? If so, why?

bragaz:
Yes I was referring to your comment in #36 :
> Such lists can be used as reference to know what providers and MIME types are supported.
They can also be used by the governance to add a new provider or a new MIME type to the supported list

I Miss @kwunyeung 's one and he's totally right, application should take care of that. This will also keep chain cleaner and clients could implement as many providers they want.

x/posts/internal/types/media_post_test.go
```diff
@@ -0,0 +1,952 @@
+package types_test
+
+import (
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"fmt"
	"testing"

```

bragaz:
I've been struggling all day with `MarshalJSON()` `UnmarshalJSON()` methods of MediaPost and I wasn't able to find any good implementation, @RiccardoM do you have any suggestions on how to do these method starting by `String()`?

RiccardoM:
> I've been struggling all day with `MarshalJSON()` `UnmarshalJSON()` methods of `MediaPost` and I wasn't able to find any good implementation, @RiccardoM do you have any suggestions on how to do these method starting by `String()`?

You are doing the exact opposite of what should be done. If you take a look at the `TextPost`, I have: 

1. Overridden the `MashalJson` with a custom implementation
2. Overridden the `String` method with a custom implementation that calls the `MarshalJson` method

You are overriding the `MarshalJson` calling the `String` method, which will not work at all as the `String` method returns just the class name by default. 

So the good implementation should be something like

```golang
func (mp MediaPost) MarshalJson() ([]byte, err) {
  type mediaPostJson MediaPost
  return json.Marshal(mediaPostJson(mp))
}

func (mp MediaPost) String() string {
  json, err := json.Marshal(&mp) 
  if err != nil {  
    panic(err)
  }
  return string(json)
}
```

Actually, it's not needed to use the `json.Marshal` from the `String` method and is actually not the best option to be honest. 

We should change it inside `TextPost` too, so that the `String` method returns less data in a plain format, much like it does for the `Posts` type. It should be better to open a new issue for this.

bragaz:
I was doing that way since I discovered through the `String()` test that somehow, it was serializing the media post without the `Medias`, so just the first object of `MediaPost`. After that, I thought that it was because it was an array and needed to be handled differently...and so i tried to implement everything like you saw.

RiccardoM:
> I was doing that way since I discovered through the `String()` test that somehow, it was serializing the media post without the `Medias`, so just the first object of `MediaPost`. After that, I thought that it was because it was an array and needed to be handled differently...and so i tried to implement everything like you saw.

I would prefer we stick to an implementation like the one I've suggested, and we later discuss any problem that might raise individually once we have the tests setup and failing. 

Please don't change the implementations completely just because tests are failing, and do not remove the tests too. Keep them as it might be useful to understand what's wrong.

bragaz:
Besides the conflicts, we need to understand why the marshalling of `MediaPost` don't include/overwrite the `medias` part. All failing tests are about it.
`MediaPost` `TextPost` value is marshalled correctly, but it's the only one.
E.g.
With the `String()` method implemented as follow:
```golang
// String implements fmt.Stringer
func (pm PostMedia) String() string {
	bytes, err := json.Marshal(&pm)
	if err != nil {
		panic(err)
	}

	return string(bytes)
}
```
The result of the function's test is this:
```
{
  "id": "2",
  "parent_id": "0",
  "message": "media Post",
  "created": "0",
  "last_edited": "0",
  "allows_comments": false,
  "subspace": "desmos",
  "creator": "cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns"
}
```
And it's wrong because I think we need to get something like:
```
{
  "post": {
    "id": "2",
    "parent_id": "0",
    "message": "media Post",
    "created": "0",
    "last_edited": "0",
    "allows_comments": false,
    "subspace": "desmos",
    "creator": "cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns"
  },
  "medias": [
    {
      "provider": "ipfs",
      "uri": "uri",
      "mime_Type": "text/plain"
    }
  ]
}
```

I've tried to implement `String()` as follow:
```golang
func (mp MediaPost) String() string {
	bytes, err := mp.TextPost.MarshalJSON()
	bytes2, err := json.Marshal(&mp.Medias)
	bb := append(bytes[:], bytes2[:]...)
	if err != nil {
		panic(err)
	}
	return string(bb)
}
```
It isn't printing a beautiful result but at least print medias...I really don't understand why this is happening..maybe I'm missing something! Let me know.

This problem is the same in `PostQueryResponse`.

RiccardoM:
@bragaz Please pull from master to solve the existing conflicts

RiccardoM:
@bragaz I had the time to look deeper into this, and I have found something really interesting. 

### How JSON encoding works and how to fix the problem
Currently we have the following type inheritances: 

```
  Post 
   ^
TextPost
   ^
MediaPost
```

Both `TextPost` and `MediaPost` implement `MarshalJSON`. What happens is that during the call of `json.Marshal(m)`, Go performs a lookup of the `MarshalJson` method in a top-down fashion starting from `Post`. 

What does this mean? It means that it stops as soon as one type implements `MarshalJson` and uses that implementation. In our cases, it stopped at `TextPost`, which resulted in every `MediaPost` being serialized as a `TextPost` object instead. 

By removing the implementation of both `MarshalJSON` and `UnmarshalJSON` from `TextPost`, it starts working properly again. The results are: 


```golang
var m MediaPost = ...
json.Marshal(m)

// {"id":"2","parent_id":"0","message":"media Post","created":"2020-01-01T15:15:00Z","last_edited":"0001-01-01T00:00:00Z","allows_comments":false,"subspace":"desmos","creator":"cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns","medias":[{"uri":"uri","provider":"provider","mime_Type":"text/plain"}]}
```

It is completely fine to remove both `MarshalJSON` and `UnmarshalJSON` from `TextPost` as it has no custom implementation that Amino does not handle properly. 

However, we need to implement those methods into `MediaPost` as it encapsulates `TextPost` and Amino does not handle very well encapsulation. 

### Changes in the serialization fashion 
Looking at this, it made me think about the best way we can handle all of this properly. What we could do is the following. 

JSON serialized posts that extend `TextPost` should add to it the new fields. This means overriding the `MarshalJSON` and `UnmarshalJSON` for each extending post type. 

This will result in `MediaPost`s being serialized as follows: 

```json
{
  "id": "2",
  "parent_id": "0",
  "message": "media Post",
  "created": "2020-01-01T15:15:00Z",
  "last_edited": "0001-01-01T00:00:00Z",
  "allows_comments": false,
  "subspace": "desmos",
  "creator": "cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns",
  "medias": [
    {
      "uri": "uri",
      "provider": "provider",
      "mime_Type": "text/plain"
    }
  ]
}
```

As you can see, we do **not** encapsulate the `TextPost` into a `"post"` object, but we still have the new `"medias"` array. 

This results in a more clean serialization overall. 

#### Problems with this approach 
Thinking ahead of time, this solution might however bring some new problems when we will start to implement poll posts (#14) too. 

What would happen if a users wants to create a poll having an image inside? 

This should result in a hybrid type that inherits from both `MediaPost` and `PollPost`, but as they will have both overridden the `MarshalJSON` method, one of the two will be chosen instead of the new one. 

However, we can think this thought later, but it is still important to now it now.

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+func (mp MediaPost) GetID() PostID {
+	return mp.PostID
+}
+
+// GetParentID implements Post GetParentID
+func (mp MediaPost) GetParentID() PostID {
+	return mp.ParentID
+}
+
+func (mp MediaPost) SetMessage(message string) Post {
```
RiccardoM: Please add a method description using inline comments (`\\`) to this and all the below methods

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+func (mp MediaPost) Equals(other Post) bool {
+	// Cast and delegate
+	if otherMp, ok := other.(MediaPost); ok {
+		return checkMediaPostEquals(mp, otherMp)
+	}
+
+	return false
+}
+
+// Equals implements Post Equals
```
RiccardoM: This method description does not fit the name of the method

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+
+	if len(first.Medias) != len(second.Medias) {
+		return false
+	}
+
+	for index, media := range first.Medias {
+		if media != second.Medias[index] {
+			return false
+		}
+	}
```
RiccardoM: I would prefer we review this putting all the `Medias` equality checking inside the `Medias` type instead. 

This would allows to simplify this method to be

```golang
return first.TextPost.Equals(second.TextPost) && first.Medias.Equals(second.Medias)
```

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+func (pms PostMedias) String() string {
+	bytes, err := json.Marshal(&pms)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bytes)
+}
+func (pms PostMedias) Equals(other PostMedias) bool {
+	for index, postMedia := range pms {
```
RiccardoM: This is now checking for lengths differences. 

For example, if we have two arrays like 

```
A1 = [M1, M2]
A2 = [M2]
```

This would return true while it should be false. 

Please fix this and add a test case to show that this fails.

x/posts/internal/types/msgs.go
```diff
@@ -94,15 +116,65 @@ func (msg MsgCreatePost) ValidateBasic() sdk.Error {
...
+func (msg MsgCreateTextPost) GetSigners() []sdk.AccAddress {
 	return []sdk.AccAddress{msg.Creator}
 }
 
+// ----------------------
+// --- MsgCreateMediaPost
+// ----------------------
+
+// MsgCreateMediaPost defines a CreateMediaPost message
+type MsgCreateMediaPost struct {
```
RiccardoM: Do we really want a new type just to add medias? 

Maybe we should go with encapsulations instead, transforming the original `MsgCreatePost` into 


```golang
type MsgCreatePost struct {
  // All post fields
  Medias        PostMedias        `json:"post_medias,omitempty"`
}
```

We also should have a new method called `NewMsgCreatePostWithMedia`: 

```golang
func NewMsgCreatePostWithMedia(<all post data>, medias PostMedias) MsgCreatePost {
  // ...
}
```

x/posts/internal/types/post_response.go
```diff
@@ -7,13 +7,34 @@ import (
...
 // that is returned to user upon a query
 type PostQueryResponse struct {
-	Post
+	Type      string    `json:"type"`
+	Post      Post      `json:"post"`
 	Reactions Reactions `json:"reactions"`
 	Children  PostIDs   `json:"children"`
 }
 
+func (response PostQueryResponse) String() string {
```
RiccardoM: Why is this method necessary?

x/posts/internal/types/post_response.go
```diff
@@ -7,13 +7,34 @@ import (
...
+	bytes, err := json.Marshal(&response)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bytes)
+}
+
 func NewPostResponse(post Post, reactions Reactions, children PostIDs) PostQueryResponse {
+	var ttype string
```
RiccardoM: I would prefer we rename this something like `msgType` 

```suggestion
	var msgType string
```

Also, we should initialize it to `TextPost` by default, and override the initial value later

x/posts/internal/types/strings.go
```diff
@@ -0,0 +1,39 @@
+package types
+
+// Strings represents a slice of strings.
+type Strings []string
```
RiccardoM: Where is this needed?

x/posts/legacy/v0.1.0/types.go
```diff
@@ -15,16 +15,16 @@ type GenesisState struct {
 // PostID represents a unique post id
 type PostID uint64
 
-// Post is a struct of a post
+// TextPost is a struct of a post
```
RiccardoM: Refactoring error, I think this should be `Post` instead

```suggestion
// Post is a struct of a post
```

x/posts/legacy/v0.1.0/types.go
```diff
@@ -15,16 +15,16 @@ type GenesisState struct {
...
 type Post struct {
 	PostID            PostID         `json:"id"`                 // Unique id
-	ParentID          PostID         `json:"parent_id"`          // Post of which this one is a comment
+	ParentID          PostID         `json:"parent_id"`          // TextPost of which this one is a comment
```
RiccardoM: Refactoring error, I think this should be `Post` instead

```suggestion
	ParentID          PostID         `json:"parent_id"`          // Post of which this one is a comment
```

x/posts/legacy/v0.1.0/types.go
```diff
@@ -15,16 +15,16 @@ type GenesisState struct {
...
 	Message           string         `json:"message"`            // Message contained inside the post
 	Created           sdk.Int        `json:"created"`            // Block height at which the post has been created
 	LastEdited        sdk.Int        `json:"last_edited"`        // Block height at which the post has been edited the last time
 	AllowsComments    bool           `json:"allows_comments"`    // Tells if users can reference this PostID as the parent
 	ExternalReference string         `json:"external_reference"` // Used to know when to display this post
-	Owner             sdk.AccAddress `json:"owner"`              // Creator of the Post
+	Owner             sdk.AccAddress `json:"owner"`              // Creator of the TextPost
```
RiccardoM: Refactoring error, I think this should be `Post` instead

```suggestion
	Owner             sdk.AccAddress `json:"owner"`              // Creator of the Post
```

x/posts/alias.go
```diff
@@ -48,12 +50,20 @@ type (
 	PostIDs      = types.PostIDs
 	Post         = types.Post
 	Posts        = types.Posts
+	TextPost     = types.TextPost
+	TextPosts    = types.TextPosts
```
RiccardoM: Do we really need this type and alias?

x/posts/alias.go
```diff
@@ -48,12 +50,20 @@ type (
...
+	MediaPost    = types.MediaPost
+	MediaPosts   = types.MediaPosts
```
RiccardoM: Do we really need this type and alias?

x/posts/internal/types/codec.go
```diff
@@ -13,8 +13,13 @@ func init() {
...
 // RegisterCodec registers concrete types on the Amino codec
 func RegisterCodec(cdc *codec.Codec) {
-	cdc.RegisterConcrete(MsgCreatePost{}, "desmos/MsgCreatePost", nil)
+	cdc.RegisterInterface((*MsgCreatePost)(nil), nil)
+	cdc.RegisterConcrete(MsgCreateTextPost{}, "desmos/MsgCreateTextPost", nil)
+	cdc.RegisterConcrete(MsgCreateMediaPost{}, "desmos/MsgCreateMediaPost", nil)
 	cdc.RegisterConcrete(MsgEditPost{}, "desmos/MsgEditPost", nil)
 	cdc.RegisterConcrete(MsgAddPostReaction{}, "desmos/MsgAddPostReaction", nil)
 	cdc.RegisterConcrete(MsgRemovePostReaction{}, "desmos/MsgRemovePostReaction", nil)
+	cdc.RegisterInterface((*Post)(nil), nil)
```
RiccardoM: What happens if you don't register the interface or concrete types? Can we test this?

x/posts/client/cli/tx.go
```diff
@@ -67,11 +68,25 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+				for i := 3; i < len(args); i++ {
+					arg := strings.Split(args[i], ",")
+					media := types.NewPostMedia(arg[0], arg[1], arg[2])
+					medias = append(medias, media)
+				}
+				msg = types.NewMsgCreateMediaPost(args[1], parentID, allowsComments, args[0], map[string]string{}, from, time.Now().UTC(), medias)
+			} else {
+				msg = types.NewMsgCreateTextPost(args[1], parentID, allowsComments, args[0], map[string]string{}, from, time.Now().UTC())
+			}
+
```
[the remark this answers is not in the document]
bragaz: we can't initialize `msg` like that cause type wont match then inside `NewMsgCreateMediaPost`.
We may do a typecheck like this:
```golang
var msg types.MsgCreatePost
msg = types.NewMsgCreateTextPost(args[1], parentID, allowsComments, args[0], map[string]string{}, from, time.Now().UTC())

//if there's some medias
if len(args) > 3 {
    var medias types.PostMedias
    //read each media and add it to the medias if valid
    for i := 3; i < len(args); i++ {
	arg := strings.Split(args[i], ",")
	media := types.NewPostMedia(arg[0], arg[1], arg[2])
	medias = append(medias, media)
    }
    if textMsg, ok := msg.(types.MsgCreateTextPost); ok {
	 msg = types.NewMsgCreateMediaPost(textMsg, medias)
    } 
}
```

should I take this way?

x/posts/internal/types/codec.go
bragaz: How can I test such a thing? building custom codecs without it?

x/posts/client/cli/tx.go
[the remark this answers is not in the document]
RiccardoM: @bragaz Yes, please. Go this way :100:

x/posts/internal/types/codec.go
RiccardoM: @bragaz Just delete those lines and run the tests. If they pass, it means they are useless.

x/posts/client/cli/tx.go
```diff
@@ -67,11 +76,30 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+			if len(args) > 3 {
+				var medias types.PostMedias
+				// Read each media and add it to the medias if valid
+				for i := 3; i < len(args); i++ {
+					arg := strings.Split(args[i], ",")
+					if len(arg) == 3 {
+						media := types.NewPostMedia(arg[0], arg[1], arg[2])
+						medias = append(medias, media)
+					} else {
+						return sdk.ErrUnknownRequest("If medias are present, you should specify uri, provider and mime type")
```
RiccardoM: I would improve this error message adding an example of how this can be done, or demanding the user to use the `--help` flag

x/posts/client/cli/tx.go
```diff
@@ -43,9 +44,17 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
 // GetCmdCreatePost is the CLI command for creating a post
 func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create [subspace] [message] [allows-comments]",
+		Use:   "create [subspace] [message] [allows-comments] [[[uri],[provider],[mime-type]]...]",
 		Short: "Create a new post",
-		Args:  cobra.ExactArgs(3),
+		Long: fmt.Sprintf(`
```
RiccardoM: I think we can improve the command description making it more clear as follows. 

>>>
Create a new post, specifying the subspace, message and whether or not it will allow for comments. 

Optional media attachments are also supported. 
If you with to add one or more media attachment, you have to specify a uri, a provider and a mime type for each. You can do so by concatenating them together separated by a comma (,). 

Example usages: 

%s tx posts create "desmos" "Hello world!" true
%s tx posts create "demos" "A post with media" true "https://example.com,text/plain"
%s tx posts create "desmos" "A post with multiple medias" false "https://example.com/media1,text/plain" "https://example.com/media2,application/json"
>>>

x/posts/client/rest/tx.go
```diff
@@ -44,7 +44,16 @@ func createPostHandler(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
-		msg := types.NewMsgCreatePost(req.Message, parentID, req.AllowsComments, req.Subspace, req.OptionalData, addr, req.CreationTime)
+		var msg types.MsgCreatePost
+		msg = types.NewMsgCreateTextPost(req.Message, parentID, req.AllowsComments, req.Subspace, req.OptionalData,
```
RiccardoM: If this `msg` variable initialization exceeds the line limit, can we put each value on its line? I think it would make it more easy to read.

x/posts/client/rest/tx.go
```diff
@@ -44,7 +44,16 @@ func createPostHandler(cliCtx context.CLIContext) http.HandlerFunc {
...
+			addr, req.CreationTime)
+
+		if req.Medias != nil && len(req.Medias) != 0 {
```
RiccardoM: In Go, a `nil` array will have `len == 0` so the first check can be omitted without any problem, we can simply check for `len(req.Medias)`

x/posts/legacy/v0.1.0/types.go
RiccardoM: This is still here :stuck_out_tongue_winking_eye:

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+func (mp MediaPost) String() string {
+	bytes, err := json.Marshal(&mp)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bytes)
+}
+
+// GetID implements Post GetID
```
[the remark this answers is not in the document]
RiccardoM: Still here

x/posts/internal/types/media_post.go
```diff
@@ -0,0 +1,260 @@
...
+	return true
+}
+
+// ---------------
+// --- PostMedia
+// ---------------
+
+type PostMedia struct {
+	URI      string `json:"uri"`
+	Provider string `json:"provider"`
```
RiccardoM: @bragaz Is this `Provider` useful in some place? If not, I think we should remove it.

x/posts/internal/types/codec.go
RiccardoM: @bragaz Have you tried deleting them? What is the result?

x/posts/internal/types/media_post.go
bragaz: mmmh, i don't think so..it's a more client-related thing IMO.

bragaz:
Removing 
```golang
cdc.RegisterInterface((*Post)(nil), nil)
``` 
Will result in some errors in tests and handler because it's an unregistered interface.

x/posts/internal/types/media_post.go
bragaz: If you want I can remove that

x/posts/client/cli/tx.go
```diff
@@ -67,11 +79,38 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+
+			// If there are some medias
+			if len(args) > 3 {
+				var medias types.PostMedias
+				// Read each media and add it to the medias if valid
+				for i := 3; i < len(args); i++ {
+					arg := strings.Split(args[i], ",")
+					if len(arg) == 3 {
+						media := types.NewPostMedia(arg[0], arg[1], arg[2])
+						medias = append(medias, media)
```
RiccardoM: What would happen if the user specifies two times the same media data? Should we use something like `appendIfMissing` instead of `append` here?

x/posts/client/cli/tx.go
```diff
@@ -67,11 +79,38 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+					} else {
+						return sdk.ErrUnknownRequest("If medias are present, you should specify uri, provider and mime type, if you are confused, please use the --help flag")
+					}
+				}
+				if textMsg, ok := msg.(types.MsgCreateTextPost); ok {
```
RiccardoM: This check will always succeed, can we simply delete it and replace it with something like the following?

```golang
msg = types.NewMsgCreateMediaPost(MsgCreateTextPost(msg), medias))
```

x/posts/client/rest/tx.go
```diff
@@ -44,7 +44,16 @@ func createPostHandler(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
-		msg := types.NewMsgCreatePost(req.Message, parentID, req.AllowsComments, req.Subspace, req.OptionalData, addr, req.CreationTime)
+		var msg types.MsgCreatePost
+		msg = types.NewMsgCreateTextPost(req.Message, parentID, req.AllowsComments, req.Subspace, req.OptionalData,
+			addr, req.CreationTime)
+
+		if len(req.Medias) != 0 {
+			if textMsg, ok := msg.(types.MsgCreateTextPost); ok {
```
RiccardoM: This check will always succeed. Can we replace it with something like the following? 

```golang
msg = types.NewMsgCreateMediaPost(MsgCreateTextPost(msg), req.Medias)
```

x/posts/client/cli/tx.go
bragaz: I was thinking about that when I was implementing it...it is good to forbid user to post the same media more times? I mean, I can post the same GIF two times,  but that will not be the same for a video or a PDF...so yeah, we should use `appendIfMissing` instead.

x/posts/internal/keeper/handler.go
```diff
@@ -83,29 +105,29 @@ func handleMsgEditPost(ctx sdk.Context, keeper Keeper, msg types.MsgEditPost) sd
 	}
 
 	// Checks if the the msg sender is the same as the current owner
-	if !msg.Editor.Equals(existing.Creator) {
+	if !msg.Editor.Equals(existing.Owner()) {
```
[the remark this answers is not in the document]
bragaz: Just for not having the same name to avoid confusion, I can switch it with `Creator` if you prefer that

x/posts/internal/keeper/keeper.go
```diff
@@ -126,27 +126,27 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
 		// match creation time if valid height
 		if params.CreationTime != nil {
-			matchCreationTime = params.CreationTime.Equal(p.Created)
+			matchCreationTime = params.CreationTime.Equal(p.CreationTime())
 		}
 
 		// match allows comments
 		if params.AllowsComments != nil {
-			matchAllowsComments = *params.AllowsComments == p.AllowsComments
+			matchAllowsComments = *params.AllowsComments == p.CanComment()
```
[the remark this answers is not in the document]
bragaz: Same reason of `Creator`

x/posts/internal/types/events.go
```diff
@@ -3,6 +3,7 @@ package types
 // Magpie module event types
 const (
 	EventTypePostCreated         = "post_created"
+	EventTypeMediaPostCreated    = "media_post_created"
```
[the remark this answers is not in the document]
bragaz: Nope, just put here in case of, should I remove it?

x/posts/client/cli/tx.go
RiccardoM: @bragaz Using `appendIfMissing` here means that we are not allowing to use the same media inside the same post. E.g., the user will not be able to create a post having the same exact GIF, PDF, or file in it. 

This is correct, and does not refer to creating two posts with the same media inside. This means

- Alice creating post P1 with media M1 will be allowed
- Alice creating post P1 with media M1 and M2 will be allowed
- Alice creating post P1 with media M1 and M1 (again) will **not** be allowed
- Alice creating post P1 with media M1 and post P2 with the same media M1 will be allowed

x/posts/internal/keeper/handler.go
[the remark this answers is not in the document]
bragaz: If they have the same name an error occurs

x/posts/internal/types/strings.go
bragaz: I implemented this at the beginning but it result useless and forgot about it, I will remove it.

x/posts/internal/types/post_response.go
bragaz: to make a correct query output in `query.go` at line 55 	
```golang
return cliCtx.PrintOutput(out)
 ```

**Comment by bragaz** [comment]
Following the discussion made about Interfaces and types, I'm closing this PR in order to open another one with the following changes:
- Remove all the interfaces
- A single `Post` implementation as follow:
```golang
type Post struct {
  // All TextPost's field
  Medias []Media
}
type MsgCreatePost {
  // All TextPost's field
  Medias []Media
}
```

PR #81, opened 2020-01-20
Media post implementation

### LINKED ISSUE #36: Support for media posts

bragaz:
## Description
This PR (follows #72 ) implements `MediaPost` and relative methods. Closes #36. 

TODO:
- [x] Implement types;
- [x] Implement message;
- [x] Update codecs;
- [x] Wrote CLI commands.

## Checklist
- [x] Wrote tests.


RiccardoM:
@bragaz Please fix the title and reference the closed PR inside the description too.

x/posts/internal/types/post_test.go
```diff
@@ -339,12 +349,47 @@ func TestPost_Validate(t *testing.T) {
...
 				date,
 				owner,
+				medias,
 			),
 			expError: "post optional data values cannot exceed 200 characters. key1 of post with id 1 is longer than this",
 		},
 		{
-			post:     types.NewPost(types.PostID(1), types.PostID(0), "Message", true, "Desmos", map[string]string{}, date, owner),
-			expError: "",
+			post: types.NewPost(
```
RiccardoM: I think this test and the following ones should be put inside another function as they verify the behavior of `PostMedias.Validate()` only

x/posts/internal/types/post_test.go
```diff
@@ -608,6 +676,40 @@ func TestPost_Equals(t *testing.T) {
 				Subspace:       "desmos",
 				OptionalData:   map[string]string{},
 				Creator:        otherOwner,
+				Medias:         medias,
+			},
+			expEquals: false,
+		},
+		{
+			name: "Different medias",
```
RiccardoM: I think this test and the below one should be put inside another functions as they verify only the `PostMedias.Equals` implementation

x/posts/internal/keeper/handler.go
```diff
@@ -11,14 +11,14 @@ import (
...
 		switch msg := msg.(type) {
-		case types.MsgCreatePost:
-			return handleMsgCreatePost(ctx, keeper, msg)
 		case types.MsgEditPost:
 			return handleMsgEditPost(ctx, keeper, msg)
 		case types.MsgAddPostReaction:
 			return handleMsgAddPostReaction(ctx, keeper, msg)
 		case types.MsgRemovePostReaction:
 			return handleMsgRemovePostReaction(ctx, keeper, msg)
+		case types.MsgCreatePost:
```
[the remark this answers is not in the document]
bragaz: Yes it was there only because of the previous implementation

x/posts/internal/keeper/handler_test.go
```diff
@@ -116,19 +182,19 @@ func Test_handleMsgEditPost(t *testing.T) {
 		{
 			name:       "Post not found",
 			storedPost: nil,
-			msg:        types.NewMsgEditPost(types.PostID(0), "Edited message", testPostOwner, testPost.Created),
+			msg:        types.NewMsgEditPost(types.PostID(0), "Edited message", testPostOwner, testPost.LastEdited),
```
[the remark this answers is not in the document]
bragaz: my fault while fixing tests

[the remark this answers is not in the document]
bragaz: There's also a test where at lines 55 where a post with media is stored correctly

x/posts/internal/keeper/querier_test.go
```diff
@@ -122,18 +122,32 @@ func Test_queryPosts(t *testing.T) {
...
 				types.NewPostResponse(
-					types.NewPost(types.PostID(2), types.PostID(1), "Child", false, "", map[string]string{}, testPost.Created, creator),
+					types.NewPost(types.PostID(2), types.PostID(1), "Child", false, "", map[string]string{}, testPost.Created, creator, testPost.Medias),
+					types.Reactions{},
+					types.PostIDs{},
+				),
+			},
+		},
+		{
+			name: "Empty params returns all posts",
```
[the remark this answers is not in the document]
bragaz: sure

x/posts/internal/types/post.go
```diff
@@ -133,21 +135,7 @@ func NewPost(id, parentID PostID, message string, allowsComments bool, subspace
 		Subspace:       subspace,
 		OptionalData:   optionalData,
 		Creator:        creator,
-	}
-}
-
-func NewPostComplete(id, parentID PostID, message string, created, lastEdited time.Time, allowsComments bool,
```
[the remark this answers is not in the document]
bragaz: I removed this because it is unused

x/posts/internal/types/post_test.go
bragaz: similar tests are already inside post_media_test.go so if you want I can remove these two

x/posts/client/cli/tx.go
```diff
@@ -67,11 +80,39 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+
+			// If there are some medias
+			if len(args) > 3 {
+				var medias types.PostMedias
+				// Read each media and add it to the medias if valid
+				for i := 3; i < len(args); i++ {
+					arg := strings.Split(args[i], ",")
+					if len(arg) != 2 {
+						return fmt.Errorf("if medias are specified, they shouldn't have empty fields, please use the " +
+							"--help flag to know more")
```
RiccardoM: Probably we can have this on the same line of the first part of the message, even if it exceeds the limit

x/posts/internal/keeper/common_test.go
```diff
@@ -46,6 +46,11 @@ func testCodec() *codec.Codec {
 var testPostOwner, _ = sdk.AccAddressFromBech32("cosmos1y54exmx84cqtasvjnskf9f63djuuj68p7hqf47")
 var timeZone, _ = time.LoadLocation("UTC")
 var testPostCreationDate = time.Date(2020, 1, 1, 15, 15, 00, 000, timeZone)
+var medias = types.PostMedia{
```
RiccardoM: No need to have this either, we can simply put it directly inside the `testPost` initialization

```golang
var testPost = types.NewPost(
	//...
       types.PostMedias{types.PostMedia{...}}
)
```

Also, I would advice using a `NewPostMedia` function instead of the type constructor

x/posts/internal/types/genesis_test.go
```diff
@@ -29,10 +29,8 @@ func TestValidateGenesis(t *testing.T) {
 		{
 			name: "Genesis with invalid reaction errors",
 			genesis: types.GenesisState{
-				Posts: types.Posts{},
-				Reactions: map[string]types.Reactions{
-					"1": {types.Reaction{Owner: nil}},
-				},
+				Posts:     types.Posts{},
+				Reactions: map[string]types.Reactions{"1": {types.Reaction{Owner: nil}}},
```
RiccardoM: Please revert these changes as they appear on `master`

x/posts/internal/types/post.go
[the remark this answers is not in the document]
RiccardoM: Ok, so we can re-remove it I guess

x/posts/internal/types/post_media.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+// ---------------
+// --- PostMedias
+// ---------------
+
+type PostMedias []PostMedia
+
+// String implements stringer interface
```
RiccardoM:
```suggestion
// String implements fmt.Stringer
```

x/posts/internal/types/post_media.go
```diff
@@ -0,0 +1,120 @@
...
+func (pms PostMedias) String() string {
+	out := "medias - [URI] [Mime-Type]\n"
```
RiccardoM:
```suggestion
	out := "Medias - [URI] [Mime-Type]\n"
```

x/posts/internal/types/post_media.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+// String implements fmt.Stringer
+func (pm PostMedia) String() string {
+	out := "Media - "
+
+	out += fmt.Sprintf(" URI - [%s] ; Mime-Type - [%s] \n", pm.URI, pm.MimeType)
+
+	return strings.TrimSpace(out)
```
RiccardoM: I would advise not splitting this into three different like, but doing all into one line 

```golang
return fmt.Sprintf("%s (%s)", pm.URI, pm.MimeType)
```

x/posts/internal/types/post_media.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Equals allows to check whether the contents of pm are the same of other
+func (pm PostMedia) Equals(other PostMedia) bool {
+	return pm.URI == other.URI && pm.MimeType == other.MimeType
+}
+
+// ParseURI checks if the given uri string is well-formed according to the regExp and return and error otherwise
+func ParseURI(uri string) error {
+	rEx := regexp.MustCompile(
```
RiccardoM: Probably we can have this as a global `var` inside the `post_media.go` file instead. This would allows us to avoid re-build it each time we parse a URI

x/posts/internal/types/post_media.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+// Equals allows to check whether the contents of pm are the same of other
+func (pm PostMedia) Equals(other PostMedia) bool {
+	return pm.URI == other.URI && pm.MimeType == other.MimeType
+}
+
+// ParseURI checks if the given uri string is well-formed according to the regExp and return and error otherwise
+func ParseURI(uri string) error {
```
RiccardoM: I would advise renaming this method to something more like `ValidateURI` or `IsURIValid` (in the latter case, returning a `bool` instead of an `error). 

Usually we refer to methods called `ParseXXX` to the ones that actually return a new type from a string, which is not done here.

x/posts/internal/types/post_media.go
RiccardoM:
```suggestion
	out := "Medias - [URI] [Mime-Type]\n"
```

x/posts/client/cli/tx.go
```diff
@@ -67,11 +80,38 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+				from,
+				time.Now().UTC(),
+				nil,
+			)
+
+			// If there are some medias
+			if len(args) > 3 {
+				var medias types.PostMedias
+				// Read each media and add it to the medias if valid
+				for i := 3; i < len(args); i++ {
```
RiccardoM: Sorry, but I've seen now that this won't work. If we have more than one media, this will be executed multiple times and only the latest media will be used. 

It should be changed to: 

```golang
if len(args) > 3 {
	medias := types.PostMedias{}

	// Read each media and add it to the medias if valid
	for i := 3; i < len(args); i++ {
		arg := strings.Split(args[i], ",")
		if len(arg) != 2 {
			return fmt.Errorf("if medias are specified, they shouldn't have empty fields, please use the --help flag to know more")
		}

		media := types.NewPostMedia(arg[0], arg[1])
		medias = medias.AppendIfMissing(media)
	}

	msg.Medias = medias
}
```

We need to initialize the `medias` to avoid a nil pointer error at line 105, also we can delete the second returned parameter from `AppendIfMissing` as it isn't necessary anymore.

PR #84, opened 2020-01-21
Use a SHA-256 as posts' subspace values

### LINKED ISSUE #82: Subspace should be a hash value
`subspace` in `Post` is better in hash value so that it can be verified by the users by comparing the hash value instead of relying on comparing with a string.

Consider an imposter is trying to build an app and put `subspace` as `C0SMOS` to pretend to be an app discussing things on `COSMOS`.

But I don't have an idea of how the UX can be done to let normal endusers to verify this yet. @RiccardoM  @bragaz please comment.

RiccardoM:
## Description
- Modifies the posts to only support SHA-256 hashes as `susbspace` values (closes #82). 
- Implemented the migration from `v0.2.0` to `v0.3.0` to properly edit existing subspaces
- Fixed the documentation

## Checklist
- [x] Wrote tests.


kwunyeung:
Should this be done in the handler or you prefer the dapp developers resolve this before posting? 

I'm also thinking if we need an dapp registry for making sure the subspace won't collide. If there is such a registry, the developers should broadcast the message the the hash without questioning. This is like the appID and secret key of the social apps.

RiccardoM:
> Should this be done in the handler or you prefer the dapp developers resolve this before posting?

Due to the fact that the subspace can also be used to query all the posts that are part of it, I think it's better to ask dapp developers to do this without having the handler to perform additional steps.

Having the handler secretly hashing the subspace would mean that we would have to hash it again during the search of posts and other operations (i.e. the edit). I'm also not sure what benefits this would bring. 

> I'm also thinking if we need an dapp registry for making sure the subspace won't collide. If there is such a registry, the developers should broadcast the message the the hash without questioning. This is like the appID and secret key of the social apps.

I'm not sure I've understood correctly this part. I've fot a couple of questions:

- How would two SHA-256 hashes collide? 
- Where should the registry be kept? 
- How would developers interact with the registry?

kwunyeung:
The registry is in the state on-chain for sure. The hash won't collide but when two dapps want to use the same subspace it will mix all the messages if we don't have a mechanism for them to register a clean subspace (which work similar to AppID).

RiccardoM:
> The registry is in the state on-chain for sure. The hash won't collide but when two dapps want to use the same subspace it will mix all the messages if we don't have a mechanism for them to register a clean subspace (which work similar to AppID).

Ok, I got what you are saying but I also think this is going to be very problematic if we are going to handle it inside the chain. Let me explain. 

Suppose we have a public registry of subspaces as you said, and suppose also that we have the method that allows application developers to register for their own subspace giving them a secret string that they need to send to the chain to make sure they are the only one that can access such subspace. The problem is: **how would they send it**? 

The only thing we can handle are transactions (we need to support the Tenderming `/txs` endpoint) and messages inside them. If we require developers to put the secret string inside messages well... they wouldn't be secret anymore. There's no way anyone can give the chain a secret without anyone knowing it (that's a con of the blockchain technology). 

What I think should be done, instead, is the following. Let's assume I want to develop a new application based on Desmos and I do not want other people to write to my subspace. What I would do is: 

1. Make sure the subspace I want is not used by anyone else.  
    To do this, I can simply use the `/posts?subspace=XXXXXX` endpoint, or we can create a simple web interface that does this. 

2. Create a mid-layer centralized API system. This system will be used both when **sending transactions** and when **reading posts**. 

   1. Each time a signed transaction is created, the app will not send it directly to the `/txs` endpoint on the chain, but on the private API. This will forward it to the chain and **register** the signer as a valid user.
   2. When reading posts, instead of asking posts directly to the `/posts` endpoint on the chain, the app will query another private API that returns the list of posts **filtered** by removing all the posts from users that have not been registered previously. 

In short, what I want to do in my dApp is: 

1. Each time a user performs a transaction from my client, I register its address on my centralized API. 
2. Each time I read for posts, I filter all the posts that are not from registered users. 

This way, clients will display only messages created from the original clients, without any additional ones. 


As you can see, this approach works for each and every possible application (they can even add the support for secret keys inside their private API system) and produces the desired results without having any impact on the chain architecture.

kwunyeung:
Haha nice @RiccardoM ! Actually I have thought of issue of the secret key when I typed the last comment. I don't think we can have a solution at this stage. Maybe later we can think of a mechanism of using ZKP of proofing the client has the secret and the proof can be propagate to the user to send a tx. We can always come back to this issue later. 

I think what we can do now is to create a registry and let developers to "claim" the subspace with a certain fee. This definitely can be implemented later when we have a better solution. So for now, let's merge this PR first.

PR #85, opened 2020-01-22
Default magpie session length inside genesis

### LINKED ISSUE #38: Parametric session length
**Is your feature request related to a problem? Please describe.**
Currently with PR #37 we've switched from using `time.Time` values to `int64` when representing the duration time of `magpie` sessions. 

Taken from @kwunyeung's https://github.com/desmos-labs/desmos/pull/37#discussion_r343842132 we should think about having a parametric session duration time. 

**Describe the solution you'd like**
I suggest leaving the user the ability to specify a session length (in blocks) when creating it. This could be done by adding a new `duration` parameter to the `MsgCreateSession` type. 

I would also like to know more about what @kwunyeung meant with "We should add this to the Genesis params later" in order to know if we can implement such thing.

RiccardoM:
## Description
Added the support for `magpie` default session length inside the genesis (closes #38).

## Checklist
- [x] Wrote tests.


x/magpie/internal/keeper/keeper_test.go
```diff
@@ -1,12 +1,51 @@
...
+	"fmt"
+	"math"
 	"testing"
 
 	"github.com/desmos-labs/desmos/x/magpie/internal/types"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestKeeper_SetDefaultSessionLength(t *testing.T) {
+	tests := []int64{-1, 0, 1, math.MaxInt32}
```
bragaz: why testing invalid lengths such as -1 and 0?

x/magpie/internal/keeper/keeper_test.go
```diff
@@ -1,12 +1,51 @@
...
+			var stored int64
+			store := ctx.KVStore(k.StoreKey)
+			k.Cdc.MustUnmarshalBinaryBare(store.Get([]byte(types.SessionLengthKey)), &stored)
+			assert.Equal(t, length, stored)
+		})
+	}
+}
+
+func TestKeeper_GetDefaultSessionLength(t *testing.T) {
+	tests := []int64{-1, -2, 0, 1, 2}
```
bragaz: why testing invalid lengths such as -1, -2 and 0?

x/magpie/internal/keeper/keeper.go
```diff
@@ -20,6 +20,29 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey) Keeper {
 	}
 }
 
+// -------------
+// --- Params
+// -------------
+
+// SetDefaultSessionLength allows to set a default session length for new magpie sessions.
+// The specified length is intended to be in number of blocks.
+func (k Keeper) SetDefaultSessionLength(ctx sdk.Context, length int64) {
```
bragaz: Is here a specific reason why you choose `int64` instead of `uint64`?

x/magpie/internal/keeper/keeper.go
```diff
@@ -20,6 +20,29 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey) Keeper {
...
+	store := ctx.KVStore(k.StoreKey)
+	store.Set([]byte(types.SessionLengthKey), k.Cdc.MustMarshalBinaryBare(length))
+}
+
+// GetDefaultSessionLength returns the default session length in number of blocks.
+func (k Keeper) GetDefaultSessionLength(ctx sdk.Context) int64 {
+	store := ctx.KVStore(k.StoreKey)
+
+	length := int64(0)
```
bragaz: Isn't here better putting `int64(240)`? 
So if for some reasons the if below doesn't match the condition at least we return a valid `DefaultSessionLength`.

x/magpie/internal/keeper/keeper_test.go
RiccardoM: Valid observation. I've change the keeper method to return an error if the length is equals or below `0`.

x/magpie/internal/keeper/keeper_test.go
RiccardoM: Removed such cases, it's OK to not test them.

x/magpie/internal/keeper/keeper.go
RiccardoM: By using `int64` the method usage is simplified inside the `handler` (which is the only meaningful place where it's used) as `ctx.BlockHeight()` is an `int64`. 

Also, if we would have to use `uint64` we should perform additional checks later as we could set the default session length to a value up to `2^64`, which would not be supported inside the `expiration` field of the `Session` type (which is a `int64` that allows values up to `2^32`)

x/magpie/internal/keeper/keeper.go
RiccardoM: I put the default value to be `0` as this would make all sessions expire immediately upon being created. I think it's always better to return the default type values based on the operations they are going to be used in later (in this case the addition has the neutral value equals to `0`). 

If all the sessions expire immediately, we will notice it and fix it fast enough. Also, by putting a hardcoded `240` here would mean we would have to remember to change it here and inside the default genesis value if we would have to edit it. As we cannot import the genesis package inside the keeper (it would cause circular dependencies) I just thought `0` would be better overall.

x/magpie/internal/keeper/keeper.go
bragaz: That's make sense!

PR #86, opened 2020-01-22
Poll post implementation

### LINKED ISSUE #14: Support for polls

bragaz:
## Description
This PR implements the support to `Polls` inside `Post`. Closes #14 . 

TODO:
- [x] Implement types;
- [x] Implement message;
- [x] Update codecs;
- [x] Wrote CLI commands;
- [x] Wrote REST commands.

## Checklist
- [x] Wrote tests.


x/posts/internal/keeper/keeper.go
```diff
@@ -170,6 +171,60 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+	usersAnswersDetails := k.GetPostPollAnswersDetails(ctx, postID)
+
+	if usersAnswersDetails, appended := usersAnswersDetails.AppendIfMissingOrIfUsersEquals(userPollAnswers); appended {
+		store.Set(k.getAnswersStoreKey(postID), k.Cdc.MustMarshalBinaryBare(&usersAnswersDetails))
+	}
+
+}
+
+// GetAllPostPollsAnswers returns the list of all the post polls answers associated with the given postID that are stored into the current state.
+func (k Keeper) GetPostPollAnswersDetails(ctx sdk.Context, postID types.PostID) types.UsersAnswersDetails {
```
golangcibot: `postID` can be `fmt.Stringer` (from `interfacer`)

x/posts/internal/types/keys.go
```diff
@@ -17,18 +17,24 @@ const (
...
 	PostReactionsStorePrefix = "reactions:"
+	PollAnswersStorePrefix   = "poll_answers:"
 
 	MaxPostMessageLength            = 500
 	MaxOptionalDataFieldsNumber     = 10
 	MaxOptionalDataFieldValueLength = 200
 
 	ActionCreatePost         = "create_post"
 	ActionEditPost           = "edit_post"
+	ActionClosePollPost      = "close_poll_post"
```
RiccardoM: I think we can remove the trailing `post` here

```suggestion
	ActionClosePollPost      = "close_poll"
```

x/posts/internal/types/keys.go
```diff
@@ -17,18 +17,24 @@ const (
...
+	ActionAnswerPollPost     = "answer_poll_post"
```
RiccardoM: As above

```suggestion
	ActionAnswerPollPost     = "answer_poll"
```

x/posts/internal/types/msgs.go
```diff
@@ -24,11 +24,12 @@ type MsgCreatePost struct {
 	Creator        sdk.AccAddress    `json:"creator"`
 	CreationDate   time.Time         `json:"creation_date"`
 	Medias         PostMedias        `json:"post_medias,omitempty"`
```
RiccardoM: I think that even `PostMedias` can be `*PostMedias`

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// ---------------
+// --- PollData
+// ---------------
+
+// PollData contains the information of a poll that is associated to a post
+type PollData struct {
+	Title                 string      `json:"title"`                   // Describes what poll is about
```
RiccardoM: Can we change this with a more generic `Question`?

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+	ProvidedAnswers       PollAnswers `json:"provided_answers"`        // Lists of answers provided by the creator
+	EndDate               time.Time   `json:"end_date"`                // RFC3339 date at which the poll will no longer accept new answers
+	Open                  bool        `json:"open"`                    // Tells if the poll is still accepting answers
```
RiccardoM: I think a better json key would be `is_open` as it's a boolean value

```suggestion
	Open                  bool        `json:"is_open"`                    // Tells if the poll is still accepting answers
```

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+		ProvidedAnswers:       providedAnswers,
+		Open:                  open,
+		AllowsMultipleAnswers: allowMultipleAnswers,
+		AllowsAnswerEdits:     allowsAnswerEdits,
+	}
+}
+
+// String implements fmt.Stringer
+func (pd PollData) String() string {
+	bytes, err := json.Marshal(&pd)
```
RiccardoM: I would avoid using `json.Marshal` inside the `String` method as it may be counter intuitive in specific cases (like the CLI output). Instead I would prefer a more simple approach.

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bytes)
+}
+
+// MarshalJSON implements json.Marshaler as Amino does
+// not respect default json composition
+func (pd PollData) MarshalJSON() ([]byte, error) {
```
RiccardoM: We don't need this as the `PollData` type is pretty simple and does not encapsulate anything inside it, so the standard JSON implementation would do it.

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+	type temp PollData
+	return json.Marshal(temp(pd))
+}
+
+// UnmarshalJSON implements json.Unmarshaler as Amino does
+// not respect default json composition
+func (pd *PollData) UnmarshalJSON(data []byte) error {
```
RiccardoM: We don't need this as the `PollData` type is pretty simple and does not encapsulate anything inside it, so the standard JSON implementation would do it.

x/posts/internal/types/poll_data.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	return append(usersAD, ansDet), true
+}
+
+// ---------------
+// --- AnswersDetails
+// ---------------
+type AnswersDetails struct {
+	Answers []uint64       `json:"answers"`
```
RiccardoM: Why did we choose to go with `uint64`? Are we really sure we want to support up to `2^64` possible answers to a post? IMO this would be a bit overkill and we can easily go with a much more simpler `uint8` that allows up to 256 answers which are more than enough IMO. 

**Note**: if this is changed, all the parsing (e.g in the CLI) should be changed aswell.

x/posts/internal/types/post.go
```diff
@@ -120,10 +120,11 @@ type Post struct {
 	OptionalData   OptionalData   `json:"optional_data,omitempty"` // Arbitrary data that can be used from the developers
 	Creator        sdk.AccAddress `json:"creator"`                 // Creator of the Post
 	Medias         PostMedias     `json:"medias,omitempty"`        // Contains all the medias that are shared with the post
```
RiccardoM: Can we change `Medias` to be `*PostMedias` as well?

x/posts/internal/types/post.go
```diff
@@ -224,7 +230,7 @@ func (p Post) Equals(other Post) bool {
 		p.Subspace == other.Subspace &&
 		equalsOptionalData &&
 		p.Creator.Equals(other.Creator) &&
-		p.Medias.Equals(other.Medias)
+		p.Medias.Equals(other.Medias) && p.PollData.Equals(other.PollData)
```
RiccardoM: This line might throw an error if `p.PollData` is `nil`.

x/posts/client/cli/tx.go
```diff
@@ -47,19 +49,41 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
 				Usage examples:
 
-				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "Hello world!" true
-				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with media" true "https://example.com,text/plain"
-				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with multiple medias" false "https://example.com/media1,text/plain" "https://example.com/media2,application/json"
+				- tx posts create "desmos" "Hello world!" true
+                - tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with media" false \
+				  --medias "https://example.com/media1,text/plain"
+				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with multiple medias" false \
+                  --medias "https://example.com/media1,text/plain https://example.com/media2,application/json"
```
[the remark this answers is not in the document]
bragaz: No problem for the queries, after our last conversations about poll's implementation I removed all the code related to them and forgot to delete them!

For the CLI command actually we can post medias in both ways:
mine and as you explained here:
``` 
>   tx posts create "..." "A post with multiple medias" false \
>   --media "https://example.com/media1,text/plain"
>   --media "https://example.com/media2,application/json" 
```
For the 2) We could do like I've done for the poll details, I think it could work!

x/posts/client/cli/tx.go
```diff
@@ -47,19 +49,41 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
+                       where you specify a map of the following fields:
+						1) The title of the poll (key: title)
+						1) The end date of your poll after which no further answers will be accepted (key: date)
+						2) A boolean value that indicates the possibility of multiple answers from users (key: multiple-answers)
+						3) A boolean value that indicates the possibility to edit the answers in future (key: allows-answer-edits)
+					2. --poll-answer 
+                       where you specify a slice of answers that will be provided to the users once they want to take part in poll votations.
+						Each answer should is identified by the text of the answer itself
+				Usage examples:
+				-  tx posts create "desmos" "Post" true \
```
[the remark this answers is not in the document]
bragaz: it was just an error! there's only `--poll-answer`

x/posts/client/cli/tx.go
```diff
@@ -47,19 +49,41 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
+				- tx posts create "desmos" "Hello world!" true
+                - tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with media" false \
+				  --medias "https://example.com/media1,text/plain"
+				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "A post with multiple medias" false \
+                  --medias "https://example.com/media1,text/plain https://example.com/media2,application/json"
+				
+				If you want to add a poll to your post you need to specify it through two flags:
+					1. --poll-details 
+                       where you specify a map of the following fields:
+						1) The title of the poll (key: title)
```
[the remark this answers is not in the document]
bragaz: Sure, it makes sense.

x/posts/client/cli/tx.go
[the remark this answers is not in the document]
RiccardoM: I think it would be better to delete the `--medias` flag and just leave the `--media` one. Even better would be change the `--media` implementation using the map inside it

x/posts/client/rest/rest.go
```diff
@@ -25,6 +25,7 @@ type CreatePostReq struct {
 	OptionalData   map[string]string `json:"optional_data"`
 	CreationTime   time.Time         `json:"creation_time"`
 	Medias         types.PostMedias  `json:"medias,omitempty"`
+	PollData       types.PollData    `json:"poll_data,omitempty"`
```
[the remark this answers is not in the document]
bragaz: I think that `types.PostMedias` doesn't need that cause it's a slice and can be nullable without using pointers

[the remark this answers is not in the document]
RiccardoM: @bragaz Yeah, you're right

x/posts/client/rest/rest.go
```diff
@@ -40,3 +41,15 @@ type RemoveReactionReq struct {
 	PostID   string       `json:"post_id"`
 	Reaction string       `json:"reaction"`
 }
+
+type AnswerPollPostReq struct {
+	BaseReq rest.BaseReq `json:"base_req"`
+	PostID  string       `json:"post_id"`
```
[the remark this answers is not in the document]
bragaz: you mean by passing it with the request instead of the body right?

x/posts/internal/keeper/handler.go
```diff
@@ -172,3 +178,100 @@ func handleMsgRemovePostReaction(ctx sdk.Context, keeper Keeper, msg types.MsgRe
...
+	// check if post has a poll
+	if post.PollData == nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("No poll associated with this post ID: %s", msg.PostID))
+	}
+
+	// check if the post has already been closed.
+	if time.Now().UTC().After(post.PollData.EndDate) {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("The poll associated with this post ID is already closed"))
+	}
+
```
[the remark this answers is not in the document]
bragaz: I moved only the last two checks cause I need the post in both functions to perform other checks...unless you agree to return both `(post, error)`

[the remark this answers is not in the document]
bragaz:
```golang
// checkPostPollValid perform checks to ensure the validity of a poll
func checkPostPollValid(ctx sdk.Context, id types.PostID, keeper Keeper) (*types.Post, error) {
	// checks if post exists
	if !found {
		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("post with id %s doesn't exist", id))
	}

	// checks if post has a poll
	if post.PollData == nil {
		return &post, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("no poll associated with ID: %s", id))
	}

	// checks if the poll is already closed or not
	if time.Now().UTC().After(post.PollData.EndDate) {
		return &post, sdkerrors.Wrap(
			sdkerrors.ErrInvalidRequest, 
			fmt.Sprintf("the poll associated with ID %s was closed at %s", post.PostID, post.PollData.EndDate),
		)	
	}

	return &post, nil
}
```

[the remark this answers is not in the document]
RiccardoM: Yes, this is perfect!

x/posts/internal/types/poll_data.go
bragaz: We use that as the `answerID` but sure, `uint8` makes more sense 🤣

bragaz: Trying using `uint8` I found that it leads to more difficult parsing

RiccardoM: I think we can parse it as `uint64` and later cast it to `uint8`. This should return an error to the user if it's larger than the max `uint8` value, which makes sense

x/posts/internal/types/post.go
bragaz: I'm gonna add more tests

x/posts/client/rest/rest.go
```diff
@@ -40,3 +41,15 @@ type RemoveReactionReq struct {
...
+	Answers []uint64     `json:"answers"`
+}
+
+type ClosePollPostReq struct {
+	BaseReq rest.BaseReq `json:"base_req"`
+	TextMsg string       `json:"text_msg"`
+	PostID  string       `json:"post_id"`
```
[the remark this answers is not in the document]
RiccardoM: This is still open

x/posts/internal/keeper/handler.go
```diff
@@ -172,3 +178,112 @@ func handleMsgRemovePostReaction(ctx sdk.Context, keeper Keeper, msg types.MsgRe
 	}
 	return &result, nil
 }
+
+// checkPostPollValid perform checks to ensure the validity of a poll
```
RiccardoM:
```suggestion
// checkPostPollValid performs all the checks to ensure the post with the given id exists, contains a poll and such poll has not been closed
```

x/posts/internal/keeper/handler.go
```diff
@@ -172,3 +178,112 @@ func handleMsgRemovePostReaction(ctx sdk.Context, keeper Keeper, msg types.MsgRe
...
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("post with id %s doesn't exist", id))
+	}
+
+	// checks if post has a poll
+	if post.PollData == nil {
+		return &post, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("no poll associated with ID: %s", id))
+	}
+
+	// checks if the poll is already closed or not
+	if time.Now().UTC().After(post.PollData.EndDate) {
```
RiccardoM: As we have the `Open` field, I would suggest using that field to check whether the post is closed or not. This is due to the fact that time checks might have different results based on the node that is performing the check itself. 

Suppose we have a network that has two nodes, and poll that has closing time at 11:30:00. Now, this happens: 

* node A checks the end date at 11:29:59.999 (1 ms before it), and the check returns true 
* node B checks the end data at 11:30:00.001 (1 ms after it), and the check returns false 

In this case, node A and node B would have two incoherent states and will later most probably produce some bugs. 
Instead, if we use the `Open` field than this will not happen.

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +196,76 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
 	return filteredPosts
 }
 
+func (k Keeper) getAnswersStoreKey(postID types.PostID) []byte {
```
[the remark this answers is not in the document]
RiccardoM: This is still open

x/posts/internal/types/events.go
```diff
@@ -6,13 +6,18 @@ const (
...
+	EventTypeAnsweredPoll        = "post_poll_answered"
+	EventTypeClosePoll           = "post_poll_closed"
 
 	// Post attributes
 	AttributeKeyPostID       = "post_id"
 	AttributeKeyPostParentID = "post_parent_id"
 	AttributeKeyPostOwner    = "post_owner"
 	AttributeKeyPostEditTime = "post_edit_time"
 
+	//Poll attributes
```
RiccardoM:
```suggestion
	// Poll attributes
```

x/posts/internal/types/poll_data.go
RiccardoM: This is still open

bragaz: didn't see your previous answer, i'm on this

bragaz: `uint8` are serialized as chars so `"answers":"AQ=="` so we need to implement marshaler for uint8 as result...any thought?

bragaz: I think we can by pass these problems by making checks on `[]uint64` length so we don't need to implement marshaler and weird conversion such as `strconv.Itoa(int(answerID))`

RiccardoM: Do we have the same problems with `uint` (which has 32 bits)? If not, I would prefer using that one instead of `uint64` as it would allow us to save some space.

bragaz: Fine with `uint`

x/posts/client/cli/tx.go
```diff
@@ -47,19 +48,38 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
+						* A boolean value that indicates the possibility of multiple answers from users (key: multiple-answers)
+						* A boolean value that indicates the possibility to edit the answers in future (key: allows-answer-edits)
+					2. --poll-answer 
+                       where you specify a slice of answers that will be provided to the users once they want to take part in poll votations.
+						Each answer should is identified by the text of the answer itself
+				Usage examples:
+				- tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "Post" true \
+  					--poll-details question="Which dog do you prefer?",multiple-answers=false,allow-answer-edits=true,date=2020-01-01T15:00:00.000Z \
+  					--poll-answer "Beagle" \
+  					--poll-answer "Carlino" \
```
RiccardoM:
```suggestion
  					--poll-answer "Pug" \
```

x/posts/client/cli/tx.go
```diff
@@ -47,19 +48,38 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
+					--poll-answer "German Sheperd,Labrador"
```
RiccardoM:
```suggestion
					--poll-answer "German Sheperd"
```

x/posts/client/cli/tx.go
```diff
@@ -84,6 +104,80 @@ func GetCmdCreatePost(cdc *codec.Codec) *cobra.Command {
...
+
+			if len(pollDetailsMap) == 0 && len(pollAnswersSlice) > 0 {
+				return fmt.Errorf("poll answers specified but no poll details found. Please use %s to specify the poll details", flagPollDetails)
+			}
+
+			if len(pollDetailsMap) > 0 && len(pollAnswersSlice) == 0 {
+				return fmt.Errorf("poll details specified but answers are not. Please use the %s to specify one or more answer", flagPollAnswer)
+			}
+
+			var pollData *types.PollData = nil
```
RiccardoM: No need to initialize to `nil` as far as I know

```suggestion
			var pollData *types.PollData
```

x/posts/client/rest/rest.go
```diff
@@ -40,3 +41,14 @@ type RemoveReactionReq struct {
 	PostID   string       `json:"post_id"`
 	Reaction string       `json:"reaction"`
 }
+
+type AnswerPollPostReq struct {
+	BaseReq rest.BaseReq `json:"base_req"`
+	Answers []uint       `json:"answers"`
+}
+
+type ClosePollPostReq struct {
```
RiccardoM: I think we can remove this as poll are auto closed inside `EndBlocker` now

x/posts/client/rest/tx.go
```diff
@@ -16,6 +16,8 @@ func registerTxRoutes(cliCtx context.CLIContext, r *mux.Router) {
 
 	r.HandleFunc("/posts/reactions", addReactionToPostHandler(cliCtx)).Methods("POST")
 	r.HandleFunc("/posts/reactions", removeReactionToPostHandler(cliCtx)).Methods("DELETE")
+
```
RiccardoM:
```suggestion
```

x/posts/client/rest/rest.go
```diff
@@ -24,7 +24,8 @@ type CreatePostReq struct {
 	Subspace       string            `json:"subspace"`
 	OptionalData   map[string]string `json:"optional_data"`
 	CreationTime   time.Time         `json:"creation_time"`
-	Medias         types.PostMedias  `json:"medias,omitempty"`
+	Medias         *types.PostMedias `json:"medias,omitempty"`
```
RiccardoM: Do we need a pointer here even if it's an array as you stated previously?

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+	return []byte(types.PollAnswersStorePrefix + postID.String())
+}
+
+// Save the userAnswersDetails associated with the given postID inside the current context
+// It assumes that the post exists and has a Poll inside it.
+// If userAnswersDetails are already present, the old ones will be overridden.
+func (k Keeper) SavePollUserAnswers(ctx sdk.Context, postID types.PostID, userPollAnswers types.AnswersDetails) {
```
RiccardoM: I would rename this method to a more generic `SavePollAnswer`

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+	usersAnswersDetails := k.GetPostPollAnswersDetails(ctx, postID)
+
+	if usersAnswersDetails, appended := usersAnswersDetails.AppendIfMissingOrIfUsersEquals(userPollAnswers); appended {
+		store.Set(k.getAnswersStoreKey(postID), k.Cdc.MustMarshalBinaryBare(&usersAnswersDetails))
+	}
+
+}
+
+// GetAllPostPollsAnswers returns the list of all the post polls answers associated with the given postID that are stored into the current state.
+func (k Keeper) GetPostPollAnswersDetails(ctx sdk.Context, postID types.PostID) types.UsersAnswersDetails {
```
RiccardoM: I would rename this to `GetPollAnswers` instead to make it more easy to comprehend

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
 	return filteredPosts
 }
 
+func (k Keeper) getAnswersStoreKey(postID types.PostID) []byte {
+	return []byte(types.PollAnswersStorePrefix + postID.String())
+}
+
+// Save the userAnswersDetails associated with the given postID inside the current context
```
RiccardoM: Should start with `SavePollUserAnswers`...

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+
+	usersAnswersDetails := k.GetPostPollAnswersDetails(ctx, postID)
+
+	if usersAnswersDetails, appended := usersAnswersDetails.AppendIfMissingOrIfUsersEquals(userPollAnswers); appended {
+		store.Set(k.getAnswersStoreKey(postID), k.Cdc.MustMarshalBinaryBare(&usersAnswersDetails))
+	}
+
+}
+
+// GetAllPostPollsAnswers returns the list of all the post polls answers associated with the given postID that are stored into the current state.
```
RiccardoM: Wrong method name inside doc

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+
+	var usersAnswersDetails types.UsersAnswersDetails
+	answersBz := store.Get(k.getAnswersStoreKey(postID))
+
+	k.Cdc.MustUnmarshalBinaryBare(answersBz, &usersAnswersDetails)
+
+	return usersAnswersDetails
+}
+
+// GetReactions allows to returns the list of reactions that have been stored inside the given context
```
RiccardoM: Wrong method name inside doc

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+		k.Cdc.MustUnmarshalBinaryBare(iterator.Value(), &userAnswers)
+		postID, _ := types.ParsePostID(strings.TrimPrefix(string(iterator.Key()), types.PollAnswersStorePrefix))
+		usersAnswersData[postID] = userAnswers
+	}
+
+	return usersAnswersData
+}
+
+// GetPostPollAnswersByUser retrieves post poll answers associated to the given ID and filtered by user
+func (k Keeper) GetPostPollAnswersByUser(ctx sdk.Context, postID types.PostID, user sdk.AccAddress) []uint {
```
RiccardoM: I don't think we need this anymore

x/posts/internal/keeper/keeper.go
```diff
@@ -195,6 +195,72 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+		if user.Equals(postPollAnswers.User) {
+			return postPollAnswers.Answers
+		}
+	}
+	return nil
+}
+
+// ClosePollPost retrieves a previously opened poll post and close it.
+// It assumes that the post associated with the given ID exists and has a poll that's not already closed
+func (k Keeper) ClosePollPost(ctx sdk.Context, postID types.PostID) {
```
RiccardoM: Would rename this to a more generic `ClosePoll` instead

x/posts/internal/keeper/querier.go
```diff
@@ -90,3 +100,31 @@ func queryPosts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]byte,
...
+	}
+
+	_, found := keeper.GetPost(ctx, id)
+	if !found {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, fmt.Sprintf("Post with id %s not found", id))
+	}
+
+	pollAnswers := keeper.GetPostPollAnswersDetails(ctx, id)
+
+	pollAnswersResponse := types.PollAnswersQueryResponse{
```
RiccardoM: Should we return an error if the post does not have an associated poll?

x/posts/internal/types/msgs.go
```diff
@@ -173,6 +187,57 @@ func (msg MsgEditPost) GetSigners() []sdk.AccAddress {
 	return []sdk.AccAddress{msg.Editor}
 }
 
+// ----------------------
+// --- MsgAnswerPollPost
+// ----------------------
+
+// MsgAnswerPollPost defines the AnswerPollPost message
+type MsgAnswerPollPost struct {
```
RiccardoM: I would rename this to `MsgAnswerPoll`

x/posts/internal/types/keys.go
```diff
@@ -17,18 +17,22 @@ const (
...
 	PostReactionsStorePrefix = "reactions:"
+	PollAnswersStorePrefix   = "poll_answers:"
 
 	MaxPostMessageLength            = 500
 	MaxOptionalDataFieldsNumber     = 10
 	MaxOptionalDataFieldValueLength = 200
 
 	ActionCreatePost         = "create_post"
 	ActionEditPost           = "edit_post"
+	ActionClosePollPost      = "close_poll"
```
RiccardoM: I would change this to `ActionClosePoll`. Also, do we still need it?

x/posts/internal/types/keys.go
```diff
@@ -17,18 +17,22 @@ const (
...
+	ActionAnswerPollPost     = "answer_poll"
```
RiccardoM: I would change this to `ActionAnswerPoll`

x/posts/abci.go
```diff
@@ -0,0 +1,25 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/desmos-labs/desmos/x/posts/internal/types"
+)
+
+// EndBlocker called upon each block end to close expired polls
+func EndBlocker(ctx sdk.Context, keeper Keeper) {
+
+	posts := keeper.GetPosts(ctx)
+
+	for _, post := range posts {
```
RiccardoM: Instead of using a `for` loop on the entire list of posts, I would prefer we use an `Iterator` as this way the avoid going through all the posts two times (one to get the list and the other one to close the posts). 

Also, we should open up an issue reminding ourselves to improve this process by iterating only the posts that are still open, and not the ones that are already closed (we can do this by looking at the `gov` module `EndBlocker` inside the Cosmos SDK.

x/posts/client/rest/rest.go
bragaz: no we don't, as long as we can checks nullability over that `Medias`

x/posts/internal/keeper/keeper.go
bragaz: I wrote this method to keep the code more readable inside `handler.go` should I remove it?

RiccardoM: I've seen its usage now. No, we can skip it. I'll close this.

x/posts/internal/keeper/querier.go
```diff
@@ -90,3 +100,39 @@ func queryPosts(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) ([]byte,
...
+	}
+
+	if post.PollData == nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("Post with id %s has no poll associated", id))
+	}
+
+	pollAnswers := keeper.GetPollAnswers(ctx, id)
+
+	if pollAnswers == nil {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("Post with id %s has no answers to poll", id))
```
RiccardoM: Why are we throwing an error if a poll does not yet have any answer to it? I think we should simply return an empty slice instead.

x/posts/internal/types/msgs.go
RiccardoM: I think we can rename this to just `medias` 

```suggestion
	Medias         PostMedias        `json:"medias,omitempty"`
```

x/posts/internal/types/msgs.go
```diff
@@ -173,6 +187,57 @@ func (msg MsgEditPost) GetSigners() []sdk.AccAddress {
...
 }
 
+// ----------------------
+// --- MsgAnswerPoll
+// ----------------------
+
+// MsgAnswerPoll defines the AnswerPoll message
+type MsgAnswerPoll struct {
+	PostID      PostID         `json:"post_id"`
+	UserAnswers []uint         `json:"provided_answers"`
```
RiccardoM: I would rename this to be `answers` instead of `provided_answers` as the latter may be confused with the `provided_answer` field inside the `Post` object.

```suggestion
	UserAnswers []uint         `json:"answers"`
```

PR #93, opened 2020-02-03
Post contents check while saving a post

### LINKED ISSUE #92: Multiple posts from the same user with the same creation date are saved
## Bug description
Currently if the same user creates multiple posts having the same message and creation date, such messages are saved properly. I think we should check whether such user has already a message with that creation date and text, not saving it if so. 

## Steps to reproduce
1. Create a post having an `X` creation date
2. Post that message multiple times inside different transactions using the `/txs` REST endpoint. 

## Expected behavior
The post should not be saved multiple times, but the call should return 200 anyway.

RiccardoM:
## Description
This PR implements a better post contents check when saving a new post, in order to fix the bug that allowed a user to create multiple posts having the exact same contents (including the creation date) only because the generated ID would have been different. 
Closes #92. 

## Checklist
- [x] Wrote unit tests.


RiccardoM:
> What if two different users post the same content as simple as just a word `nice`. The 2nd post can't be saved?

This is ok. The bugfix consists of not allowing the same user to post the same exact content on the same exact date and time, which means creating to exactly same equals posts.

PR #98, opened 2020-02-14
Hashtags implementation 

### LINKED ISSUE #96: Supports with #hashtag topics
This is different from #46 which topic was removed from the struct. `#hashtag` as a topic has been a generic behaviour on social media. We should support this on Desmos and posts can be indexed and searched by `#hashtag`s.

If we support this, each post should be tagged by the `#hashtag` in the content itself. It should be done directly on chain if this will not generate too much overhead.

bragaz:
## Description
This PR implements hashtags that can be added now to messages and used as a filter when quering posts.
Closes #96.

## Checklist
- [x] Wrote unit tests.


x/posts/internal/keeper/keeper.go
```diff
@@ -41,6 +42,18 @@ func (k Keeper) GetLastPostID(ctx sdk.Context) types.PostID {
 	return id
 }
 
+func (k Keeper) GetPostHashtags(post types.Post) []string {
+	re := regexp.MustCompile(`(?:|^)#[A-Za-z0-9]+(?:|$)`)
```
RiccardoM: Can we move this `MustCompile` expression inside the `keys.go` file, maybe under the variables definition? I think we can compile it once and use it later to improve the efficiency

x/posts/internal/keeper/keeper.go
```diff
@@ -41,6 +42,18 @@ func (k Keeper) GetLastPostID(ctx sdk.Context) types.PostID {
...
+	hashtags := re.FindAllStringSubmatch(post.Message, -1)
+
+	var hts = []string{}
+	for _, hashtagSlice := range hashtags {
+		hts = append(hts, hashtagSlice[0])
+	}
```
RiccardoM: I think that this `for` loop might create a slice having duplicated items inside if the message contains the same hashtag multiple times. For example, considering the following message: 

_This is a #test message which has multiple #test hashtags inside_

Calling `GetPostHashtags` will return `[test, test]` instead of `[test]`.

Can we fix this? Also, it may be a good idea to add a test case like this where a message has multiple times the same hashtag inside.

x/posts/internal/keeper/keeper.go
```diff
@@ -139,7 +152,23 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+				sort.Strings(postHashtags)
+				sort.Strings(params.Hashtags)
+				for index, hashtag := range params.Hashtags {
+					if matchHashtags = postHashtags[index] == hashtag; !matchHashtags {
+						break
+					}
+				}
+			} else {
+				matchHashtags = false
+			}
```
RiccardoM: I think that this piece of code can be simplified using the `&&` operators for bool values and remembering that: 

- `true && false` = `false`
- `true && true` = `true`

```suggestion
			postHashtags := k.GetPostHashtags(post)
			matchHashtags = len(postHashtags) == len(params.Hashtags)
			
			sort.Strings(postHashtags)
			sort.Strings(params.Hashtags)
			for index, hashtag := range params.Hashtags {
				matchHashtags = postHashtags[index] == hashtag
			}
```

Considering that the hashtags number will never be _that_ large, I think that iterating over their values even if the length is different does not degrade the overall performance in a significant way.

x/posts/internal/keeper/keeper_test.go
```diff
@@ -47,6 +47,50 @@ func TestKeeper_GetLastPostId(t *testing.T) {
 	}
 }
 
+func TestKeeper_GetPostHashtags(t *testing.T) {
```
RiccardoM: Can we add a case to test identical hashtags used multiple times inside the same message to make sure that their values are extracted correctly? 

- Message: `#test post with multiple #test tags inside` 
- Expected result: [`test`]

x/posts/internal/keeper/keeper.go
```diff
@@ -41,6 +42,18 @@ func (k Keeper) GetLastPostID(ctx sdk.Context) types.PostID {
 	return id
 }
 
+func (k Keeper) GetPostHashtags(post types.Post) []string {
```
RiccardoM: Can we add the documentation to this method?

x/posts/internal/types/genesis.go
```diff
@@ -1,17 +1,21 @@
 package types
 
+import "fmt"
+
 // GenesisState contains the data of the genesis state for the posts module
 type GenesisState struct {
 	Posts       Posts                  `json:"posts"`
 	PollAnswers map[string]UserAnswers `json:"poll_answers_details"`
 	Reactions   map[string]Reactions   `json:"reactions"`
+	Hashtags    map[string]PostIDs     `json:"hashtags"`
```
RiccardoM: I think this can be removed from the genesis file as we are now reading this directly from the posts messages. Am I right?

x/posts/internal/types/keys.go
```diff
@@ -34,14 +35,22 @@ var (
...
 	return append(PostStorePrefix, []byte(id.String())...)
 }
 
+// HashtagStoreKey turn an hashtag to a key used to store a post into the posts store
+// nolint: interfacer
+func HashtagStoreKey(hashtag string) []byte {
+	keyword := strings.TrimPrefix(hashtag, "#")
+	return append(HashtagPrefix, []byte(keyword)...)
+}
+
```
RiccardoM: I think this can be removed safely as no longer needed

x/posts/internal/types/keys.go
```diff
@@ -34,14 +35,22 @@ var (
 	PostCommentsStorePrefix  = []byte("comments")
 	PostReactionsStorePrefix = []byte("reactions")
 	PollAnswersStorePrefix   = []byte("poll_answers")
+	HashtagPrefix            = []byte("hashtag:")
```
RiccardoM: This could be removed safely as it shouldn't be necessary anymore

x/posts/internal/types/msgs.go
```diff
@@ -44,7 +44,7 @@ func NewMsgCreatePost(message string, parentID PostID, allowsComments bool, subs
 	}
 }
 
-// Route should return the name of the module
+// Route should return the name of the modul
```
RiccardoM: Can we revert this? 

```suggestion
// Route should return the name of the module
```

x/posts/internal/types/post.go
```diff
@@ -92,6 +92,16 @@ func (ids PostIDs) Equals(other PostIDs) bool {
...
+// Contains returns true if ids contains the given postID, false otherwise.
+func (ids PostIDs) Contains(id PostID) bool {
+	for _, ele := range ids {
+		if ele.Equals(id) {
+			return true
+		}
+	}
+	return false
+}
+
```
RiccardoM: Can this be removed as it shouldn't be necessary anymore?

x/posts/internal/types/post.go
```diff
@@ -104,6 +114,19 @@ func (ids PostIDs) AppendIfMissing(id PostID) (PostIDs, bool) {
...
+		if ele.Equals(id) {
+			ids[index] = ids[len(ids)-1] // Copy last element to index i.
+			ids[len(ids)-1] = PostID(0)  // Erase last element (write zero value).
+			ids = ids[:len(ids)-1]       // Truncate slice.
+			return ids, true
+		}
+	}
+	return ids, false
+}
+
```
RiccardoM: Can this be removed?

x/posts/internal/keeper/keeper.go
```diff
@@ -139,7 +152,23 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
 
-		if matchParentID && matchCreationTime && matchAllowsComments && matchSubspace && matchCreator {
+		// match hashtags if provided
+		if len(params.Hashtags) > 0 {
+			postHashtags := k.GetPostHashtags(post)
+			if len(postHashtags) == len(params.Hashtags) {
+				sort.Strings(postHashtags)
+				sort.Strings(params.Hashtags)
+				for index, hashtag := range params.Hashtags {
+					if matchHashtags = postHashtags[index] == hashtag; !matchHashtags {
```
RiccardoM: There's something sketchy about this equal check and the `GetPostHashtags` implementation. 

Suppose we have: 

- Message: `This is a #test post` 
- Query: `posts?hashtags=test`

Now, `GetPostHashtags` returns the following array: `[#test]` so, `#test` is matched against `test`, which is not the same and thus false is returned. 

This means that the post is not included inside the list, which is incorrect. 

What I think should happen is that `GetPostHashtags` should return the list of hashtags **without** the leading `#`. This would make for easier comparison as well.

x/posts/internal/keeper/keeper.go
bragaz: I reviewed your solution as it got some problems and committed it in the latest commit

x/posts/client/cli/query.go
```diff
@@ -172,6 +176,7 @@ $ %s query posts posts --page=2 --limit=100
 	cmd.Flags().String(flagAllowsComments, "", "(optional) filter the posts allowing comments")
 	cmd.Flags().String(flagSubspace, "", "(optional) filter the posts part of the subspace")
 	cmd.Flags().String(flagCreator, "", "(optional) filter the posts created by creator")
+	cmd.Flags().StringSlice(flagHashtag, []string{}, "(optional) filter the posts that contains the specified hashtags")
```
RiccardoM:
```suggestion
	cmd.Flags().StringSlice(flagHashtag, []string{}, "(optional) filter the posts that contain the specified hashtags")
```

x/posts/internal/keeper/keeper.go
```diff
@@ -139,7 +139,20 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
+		if len(params.Hashtags) > 0 {
+			postHashtags := post.GetPostHashtags()
+			matchHashtags = len(postHashtags) == len(params.Hashtags)
+			sort.Strings(postHashtags)
+			sort.Strings(params.Hashtags)
+			if matchHashtags {
+				for index, hashtag := range params.Hashtags {
+					matchHashtags = postHashtags[index] == hashtag
+				}
+			}
```
RiccardoM: We need to keep track of all the iterations, and we can remove the first `if` and the number of hashtags will not be that large IMO: 

```suggestion
			for index, hashtag := range params.Hashtags {
					matchHashtags &= postHashtags[index] == hashtag
			}
```

x/posts/internal/types/post.go
```diff
@@ -260,6 +260,26 @@ func (p Post) ContentsEquals(other Post) bool {
...
 }
 
+// GetPostHashtags returns all the post's hashtags without duplicates
+func (p Post) GetPostHashtags() []string {
+	hashtags := HashtagRegEx.FindAllString(p.Message, -1)
+	return Unique(hashtags)
+}
+
+// Unique returns the given input slice without any duplicated value inside it
+func Unique(input []string) []string {
```
RiccardoM: Can we move this outside of here? Maybe inside some `common.go` file or something? This might be useful somewhere else as well

x/posts/internal/types/post_test.go
```diff
@@ -760,6 +760,93 @@ func TestPost_Equals(t *testing.T) {
 	}
 }
 
+func TestPost_Unique(t *testing.T) {
```
RiccardoM: We don't need this as it's going to be tested inside `TestPost_GetPostHashtags` as well. It's fine to delete it and shouldn't change the test coverage.

x/posts/internal/types/post_test.go
```diff
@@ -760,6 +760,93 @@ func TestPost_Equals(t *testing.T) {
...
+				types.PostID(0),
+				"Post with double #hashtag #hashtag",
+				false,
+				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
+				map[string]string{},
+				date,
+				owner,
+			),
+			expHashtags: []string{"#hashtag"},
+		},
```
RiccardoM: I think it's better to make so that `GetPostHashtags` returns the hashtags **without** the leading `#`.

```suggestion
		{
			name: "Hashtags in message extracted correctly",
			post: types.NewPost(types.PostID(1),
				types.PostID(0),
				"Post with #test #desmos",
				false,
				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
				map[string]string{},
				date,
				owner,
			),
			expHashtags: []string{"test", "desmos"},
		},
		{
			name: "No hashtags in message",
			post: types.NewPost(types.PostID(1),
				types.PostID(0),
				"Post with no hashtag",
				false,
				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
				map[string]string{},
				date,
				owner,
			),
			expHashtags: []string{},
		},
		{
			name: "No same hashtags inside string array",
			post: types.NewPost(types.PostID(1),
				types.PostID(0),
				"Post with double #hashtag #hashtag",
				false,
				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
				map[string]string{},
				date,
				owner,
			),
			expHashtags: []string{"hashtag"},
		},
```

x/posts/client/cli/query.go
RiccardoM: Still open

x/posts/internal/keeper/keeper.go
```diff
@@ -139,7 +140,20 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
-		if matchParentID && matchCreationTime && matchAllowsComments && matchSubspace && matchCreator {
+		// match hashtags if provided
+		if len(params.Hashtags) > 0 {
+			postHashtags := post.GetPostHashtags()
+			matchHashtags = len(postHashtags) == len(params.Hashtags)
+			sort.Strings(postHashtags)
+			sort.Strings(params.Hashtags)
+			if matchHashtags {
+				for index, hashtag := range params.Hashtags {
+					matchHashtags = postHashtags[index] == strings.Trim(hashtag, "#")
```
RiccardoM: Still need to keep track of iterations. Also, do we really need the trim? I think we can remove it

```suggestion
					matchHashtags &= postHashtags[index] == hashtag
```

x/posts/internal/types/post.go
```diff
@@ -260,6 +260,26 @@ func (p Post) ContentsEquals(other Post) bool {
...
+
+// UniqueHashtags returns the given input slice without any duplicated hashtag value inside it
+func UniqueHashtags(input []string) []string {
+	unique := make([]string, 0, len(input))
+	m := make(map[string]bool)
+
+	for _, val := range input {
+		if _, ok := m[val]; !ok {
+			m[val] = true
+			unique = append(unique, strings.Trim(val, "#"))
```
RiccardoM: It's probably better to use `TrimLeft`: 

```suggestion
			unique = append(unique, strings.TrimLeft(val, "#"))
```

x/posts/internal/types/poll_answer.go
```diff
@@ -57,7 +57,7 @@ func ParseAnswerID(value string) (AnswerID, error) {
 
 // PollAnswer contains the data of a single poll answer inserted by the creator
 type PollAnswer struct {
-	ID   AnswerID `json:"id"`   // Unique id inside the post, serialized as a string for Javascript compatibility
+	ID   AnswerID `json:"id"`   // UniqueHashtags id inside the post, serialized as a string for Javascript compatibility
```
RiccardoM: Revert 

```suggestion
	ID   AnswerID `json:"id"`   // Unique id inside the post, serialized as a string for Javascript compatibility
```

x/posts/legacy/v0.1.0/types.go
```diff
@@ -17,7 +17,7 @@ type PostID uint64
 
 // Post is a struct of a post
 type Post struct {
-	PostID            PostID         `json:"id"`                 // Unique id
+	PostID            PostID         `json:"id"`                 // UniqueHashtags id
```
RiccardoM: Revert

```suggestion
	PostID            PostID         `json:"id"`                 // Unique id
```

x/posts/legacy/v0.2.0/types.go
```diff
@@ -29,7 +29,7 @@ type OptionalData map[string]string
 
 // Post is a struct of a post
 type Post struct {
-	PostID         PostID         `json:"id"`                      // Unique id
+	PostID         PostID         `json:"id"`                      // UniqueHashtags id
```
RiccardoM: Revert

```suggestion
	PostID         PostID         `json:"id"`                      // Unique id
```

x/posts/legacy/v0.3.0/types.go
```diff
@@ -28,7 +28,7 @@ type OptionalData map[string]string
 
 // Post is a struct of a post
 type Post struct {
-	PostID         PostID         `json:"id"`                      // Unique id
+	PostID         PostID         `json:"id"`                      // UniqueHashtags id
```
RiccardoM: Revert

```suggestion
	PostID         PostID         `json:"id"`                      // Unique id
```

x/posts/client/cli/query.go
RiccardoM: Still open

x/posts/internal/keeper/keeper_test.go
```diff
@@ -539,6 +539,11 @@ func TestKeeper_GetPostsFiltered(t *testing.T) {
 			filter:   types.QueryPostsParams{Page: 1, Limit: 5, SortBy: types.PostSortByID, SortOrder: types.PostSortOrderDescending},
 			expected: types.Posts{posts[2], posts[1], posts[0]},
 		},
+		{
+			name:     "Sortin by hashtags works properly",
```
RiccardoM:
```suggestion
			name:     "Filtering by hashtags works properly",
```

cli_test/cli_posts_test.go
```diff
@@ -38,7 +38,7 @@ func TestDesmosCLIPostsCreateNoMediasNoPollData(t *testing.T) {
 	tests.WaitForNextNBlocksTM(1, f.Port)
 
 	// Make sure the post is saved
-	storedPosts := f.QueryPosts()
+	storedPosts := f.QueryPosts("--hashtag=#hashtag")
```
RiccardoM: I think the `#` here should be deleted in order to be similar to what happens with the REST APIs:

```suggestion
	storedPosts := f.QueryPosts("--hashtag=hashtag")
```

cli_test/cli_posts_test.go
```diff
@@ -26,7 +26,7 @@ func TestDesmosCLIPostsCreateNoMediasNoPollData(t *testing.T) {
 
 	// Later usage variables
 	subspace := "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e"
-	message := "message"
+	message := "message#hashtag"
```
RiccardoM: I don't think this message is valid. It should probably be something like `"message #hashtag"`. Of course, we have the problem that CLI simulation tests do not allow for spaces in between the arguments :thinking:

x/posts/internal/types/keys.go
```diff
@@ -28,6 +28,7 @@ const (
 
 var (
 	SubspaceRegEx = regexp.MustCompile("^[a-fA-F0-9]{64}$")
+	HashtagRegEx  = regexp.MustCompile(`(?:|^)#[A-Za-z0-9]+(?:|$)`)
```
RiccardoM: Still, If we look at Twitter, this is what happens: 

| String | Hashtags found |
| :------ | :------------------| 
| `#first#second` | `[]` |
| `#first #second` | `[first, second]`
| `#first_#second` | `[]` |

So it's considered an hashtag only if it's at the beginning of the message or is after a white space character (space, newline, etc). 

That being said, I think the RegEx should be changed to be

```suggestion
	HashtagRegEx  = regexp.MustCompile(`(\s|^)#[A-Za-z0-9]+(\s|$)`)
```

bragaz: Trying this regex and it seems not working

x/commons.go
```diff
@@ -0,0 +1,15 @@
+package x
```
RiccardoM: Can we move this file inside something like `x/commons`?

x/posts/internal/types/keys.go
RiccardoM: Probably the leading `/` and trailing `/g` must be removed: 

```
\s([#][a-zA-Z_]+)
```

x/posts/internal/types/post.go
```diff
@@ -260,6 +261,18 @@ func (p Post) ContentsEquals(other Post) bool {
...
 }
 
+// GetPostHashtags returns all the post's hashtags without duplicates
+func (p Post) GetPostHashtags() []string {
+	hashtags := HashtagRegEx.FindAllString(p.Message, -1)
+	uniqueHashtags := commons.Unique(hashtags)
+	withoutHashtag := make([]string, len(uniqueHashtags))
+
+	for index, hashtag := range uniqueHashtags {
+		withoutHashtag[index] = strings.TrimLeft(hashtag, "#")
```
RiccardoM: Better to add an additional `TrimSpace` if the regex takes also those: 

```suggestion
		withoutHashtag[index] = strings.TrimSpace(strings.TrimLeft(hashtag, "#"))
```

bragaz: I don't know why but, with the suggested regEx, trim doesn't work and it don't remove the #

RiccardoM: > I don't know why but, with the suggested regEx, trim doesn't work and it don't remove the #

I think we first need to remove the spaces and then the `#` so it should be the opposite: 

```suggestion
		withoutHashtag[index] = strings.TrimLeft(strings.TrimSpace(hashtag), "#")
```

x/posts/internal/types/keys.go
bragaz: Since even the latest regEx doesn't match all the cases you mentioned before, I surfed on the web and combined a solution that match all the cases (including numeric hashtags).
Let me know what do you think about it. I've also added more testcases to make sure that it's correct!

PR #120, opened 2020-03-17
Reaction registration implementation

### LINKED ISSUE #94: Post reaction registration and limiting
## Context
Currently when a user adds a reaction to a post, he has the freedom of adding whatever reaction he prefers best. This allows a freedom of expression, and results in post having different kinds of reactions: 

- single emoji ones (e.g. "👍:", [here](https://morpheus.desmos.network/transactions/1D5BBCB4F933DBB4164500EDF8D2047FE46A9BF0948BE6EEC32735C43C1D22DA)) 
- the ones containing multiple emojis (e.g."😃😃😃" like [this one](https://morpheus.desmos.network/transactions/2CFCFB1B58ABB749729AE5A1D3E6337CA099CC2079C44EF2C74AA7822E2D2AB4)) 
- the ones containing a text (e.g. "like", [here](https://morpheus.desmos.network/transactions/B749322F8691CE12C2C1BA89980466BB4632F29FB73D32E93C57D71EC67A7789))

While this allows users to better express themselves, it makes reactions representation a lot harded inside mobile application, which will most certainly be the way users will be able to use them. 

For example, in order to present and allow users to easily use post reactions, Mooncake currently only shows single Unicode characters emojis, thus filtering out the second and third type described above.

The problem with such approach is that it causes an incoherent visualization of post reactions across the different platform. To make it more clear why, let's suppose that the following reaction have been added to a post: 

1. "👍"
2. "😃😃😃"
3. "like"

What would happen inside the explorer is that all of the reaction are going to be visualized properly. However, inside Mooncake only the first one will be shown, resulting in a conflicting and most probably confusing situation for the end user that sees two different things on two different platforms. 

To solve this problem all over the Desmos client implementation, a unified approach should be taken.

## Implementation proposal
After a brief discussion between me and @kwunyeung we found a possible solution that heavily imitates what currently happens inside Github, Slack, Discord and other channels: using emojis shortcodes.

According to this approach, all reactions will be identified by a string that has the following format: 

```
:<value>:
```

Such string, also referenced as "emoji short code", is supported by most platforms and uniquely identifies an emoji. Examples of such shortcodes are the following: 

* `:+1:` -> :+1: 
* `:small_airplane:` -> :small_airplane: 
* `:smile:` -> :smile: 

In order to allow users to express their opinions the best way the can, we have thought of also implementing **custom shotcodes** to allow for custom reactions registrations. 

In order to implement such things, a list of supported shortcodes should be kept on-chain, and each time a new reaction is added to a post, that reaction should be a registered shortcode. If it is not, the adding should not proceed. 

### Implementation details
To implement this the best way possible, I suggest we create a new module called `reactions` which contains the following messaged handling and data storing logic. 

#### Types
The current `Reaction` type should be renamed to `PostReaction`, while a new `Reaction` should be created: 

```golang
// Reaction represents a registered reaction that can be referenced 
// by its shortcode inside post reactions
type Reaction struct { 
    Shortcode string
    Value string
}
````

Such type should also have a `Validate` method that checks the following constraints: 

* `Shortcode` must be a string starting and ending with a colon (`:`) and not containing any white space, thus satisfying the following regex: `:[a-z]([a-z\d_])*:`
* `Value` must be either a single unicode emoji or a URL starting with `http(s)`. This is to allow users to add gifs and custom pics as reactions too

#### Keeper
All the current methods regarding post reactions should be kept. 

The following new methods should be created as well: 

```golang
// RegisterReaction allows to register a new reaction for later reference
func (k Keeper) RegisterReaction(ctx sdk.Context, reaction types.Reaction) {}

// DoesReactionForShortcodeExist checks whether a reaction already exists for the given shortcode, returning it if it does.
func (k Keeper) DoesReactionForShortcodeExist(ctx sdk.Context, shortcode string) (types.Reaction, bool) {}

// ListReactions returns all the registered reactions
func (k Keeper) ListReactions(ctx sdk.Context) []types.Reaction {}
```

#### Messages
The current `MsgAddReaction` and `MsgRemoveReaction` types should be moved here. 

An additional `MsgRegisterReaction` should be created: 

```golang
// MsgRegisterReaction represents the message that must be used when wanting 
// to register a new reaction shortcode and the associated value
type MsgRegisterReaction struct {
    Shortcode string `json:"shortcode"`
    Value string     `json:"value"`
}
```
   
Such message should have the following constraints: 

* `Shortcode` must be a string starting and ending with a column (`:`) and not containing any white space, thus satisfying the following regex: `:(\S)*:`
* `Value` must be either a single unicode emoji or a URL starting with `http(s)`. This is to allow users to add gifs and custom pics as reactions too

#### Handler
The currently existing handler methods that deal with post reactions should be moved inside the `reactions` module handler. 

A new method handling `MsgRegisterPostReaction` should be created. Such method should act as follows: 

1. Checks whether a reaction already exists for the specified `msg.Shortcode`. 
2. If the reaction exists, returns an error. 
3. If the reaction does not exist, registers the new reaction properly. 

#### Querier
The currently present post reactions-related queries should be moved inside this new querier. 

A new query should also be implemented, allowing a user to get all the registered reactions. 

## Conclusion
I think that with such implementation we could fix the problem described above, maintaining a nice on-chain list of all the supported reactions. As always, I'd like your thoughts on this too @kwunyeung @bragaz

bragaz:
## Description
This PR added the possibility to register a reaction to use it later when reacting to a post.
Closes #94 

## Checklist
- [x] Wrote unit tests.


x/posts/client/rest/query.go
```diff
@@ -31,6 +31,7 @@ func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/posts/{postID}", queryPostHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/posts", queryPostsWithParameterHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/posts/{postID}/poll-answers", queryPostPollAnswersHandlerFn(cliCtx)).Methods("GET")
+	r.HandleFunc("/reactions", queryRegisteredReactions(cliCtx)).Methods("GET")
```
RiccardoM: Probably to make it more similar to the CLI command, we should change this to 

```suggestion
	r.HandleFunc("/posts/registeredReactions", queryRegisteredReactions(cliCtx)).Methods("GET")
```

x/posts/client/rest/tx.go
```diff
@@ -16,6 +16,7 @@ func registerTxRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/posts/reactions", addReactionToPostHandler(cliCtx)).Methods("POST")
 	r.HandleFunc("/posts/reactions", removeReactionToPostHandler(cliCtx)).Methods("DELETE")
 	r.HandleFunc("/posts/{postID}/answers", addAnswerToPostPollHandler(cliCtx)).Methods("POST")
+	r.HandleFunc("/reaction", registerReactionHandler(cliCtx)).Methods("POST")
```
RiccardoM: Probably we should change it to be more similar to the query command: 

```suggestion
	r.HandleFunc("/posts/registeredReactions", registerReactionHandler(cliCtx)).Methods("POST")
```

x/posts/internal/keeper/handler.go
```diff
@@ -132,17 +134,17 @@ func handleMsgAddPostReaction(ctx sdk.Context, keeper Keeper, msg types.MsgAddPo
 	}
 
 	// Create and store the reaction
-	reaction := types.NewReaction(msg.Value, msg.User)
-	if err := keeper.SaveReaction(ctx, post.PostID, reaction); err != nil {
+	reaction := types.NewPostReaction(msg.Value, msg.User)
```
RiccardoM: I think there's a missing check to make sure that the `msg.Value` is a registered reaction before saving it

x/posts/internal/keeper/handler.go
```diff
@@ -281,3 +283,31 @@ func handleMsgAnswerPollPost(ctx sdk.Context, keeper Keeper, msg types.MsgAnswer
...
 	return &result, nil
 }
+
+func handleMsgRegisterReaction(ctx sdk.Context, keeper Keeper, msg types.MsgRegisterReaction) (*sdk.Result, error) {
+	_, isAlreadyRegistered := keeper.DoesReactionForShortCodeExist(ctx, msg.ShortCode, msg.Subspace)
+
+	if isAlreadyRegistered {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf(
+			"reaction with shortcode %s and subspace %s has already been registered", msg.ShortCode, msg.Subspace))
+	}
```
RiccardoM: I would merge these lines by doing

```suggestion
	if _, isAlreadyRegistered := keeper.DoesReactionForShortCodeExist(ctx, msg.ShortCode, msg.Subspace); isAlreadyRegistered {
		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf(
			"reaction with shortcode %s and subspace %s has already been registered", msg.ShortCode, msg.Subspace))
	}
```

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("%s has already reacted with %s to the post with id %s",
+			reaction.Owner, reaction.Value, postID)
+	}
+
+	// Check if the reaction is a registered one
+	post, _ := k.GetPost(ctx, postID)
+	if _, exist := k.DoesReactionForShortCodeExist(ctx, reaction.Value, post.Subspace); !exist {
+		return fmt.Errorf("reaction with short code %s isn't registered yet and can't be used to react to the post with ID %s and sub %s, please register it before use",
+			reaction.Value, postID, post.Subspace)
+	}
```
RiccardoM: I think it's better to move this inside the `Handler` instead so that this method can be called from the `InitGenesis` without much problems

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,150 @@
...
+	key := types.ReactionsStoreKey(shortcode, subspace)
+
+	bz := store.Get(key)
+
+	if bz != nil {
+		k.Cdc.MustUnmarshalBinaryBare(bz, &reaction)
+		return reaction, true
+	}
+
+	return nil, false
```
RiccardoM: Probably these lines can be simplified by writing something like

```suggestion
	if !store.HasKey(key) {
		return nil, false
	}
	
	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &reaction)
	return reaction, true
```

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,150 @@
...
+
+// RegisterReaction allows to register a new reaction for later reference
+func (k Keeper) RegisterReaction(ctx sdk.Context, reaction types.Reaction) {
+	store := ctx.KVStore(k.StoreKey)
+	key := types.ReactionsStoreKey(reaction.ShortCode, reaction.Subspace)
+	store.Set(key, k.Cdc.MustMarshalBinaryBare(&reaction))
+}
+
+// DoesReactionForShortCodeExist checks whether a reaction already exists for the given shortCode, returning it if it does.
+func (k Keeper) DoesReactionForShortCodeExist(ctx sdk.Context, shortcode string, subspace string) (reaction *types.Reaction, exist bool) {
```
RiccardoM: If we return an additional `bool` value, we should return `types.Reaction` instead of `*types.Reaction` for as the first value. Otherwise, we can omit the `bool` and return a pointer (that will be `null` when the reaction does not exist). 

However, I prefer we return a default `types.Reaction{}` if it does not exist. 

Also, by the name we can simply return the `bool`. Otherwise we might call the method `GetReactionForShortCode` instead.

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,150 @@
...
+
+	reactions := make(types.Reactions, 0)
+
+	for ; iterator.Valid(); iterator.Next() {
+		var reaction types.Reaction
+		k.Cdc.MustUnmarshalBinaryBare(iterator.Value(), &reaction)
+		reactions = append(reactions, reaction)
+	}
+
+	return reactions
```
RiccardoM: We can remove the middle `reactions :=` initialization by declaring it as the return value. Also, make sure to close the `Iterator`.

```suggestion
func (k Keeper) ListReactions(ctx sdk.Context) (reactions types.Reactions) {
	store := ctx.KVStore(k.StoreKey)
	iterator := sdk.KVStorePrefixIterator(store, types.ReactionsStorePrefix)
	defer iterator.Close()
	
	for ; iterator.Valid(); iterator.Next() {
		var reaction types.Reaction
		k.Cdc.MustUnmarshalBinaryBare(iterator.Value(), &reaction)
		reactions = append(reactions, reaction)
	}

	return reactions
```

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,150 @@
...
+
+	var reactions types.PostReactions
+	k.Cdc.MustUnmarshalBinaryBare(store.Get(types.PostReactionsStoreKey(postID)), &reactions)
+	return reactions
+}
+
+// GetReactions allows to returns the list of reactions that have been stored inside the given context
+func (k Keeper) GetReactions(ctx sdk.Context) map[types.PostID]types.PostReactions {
+	store := ctx.KVStore(k.StoreKey)
+	iterator := sdk.KVStorePrefixIterator(store, types.PostReactionsStorePrefix)
```
RiccardoM: We need to close this

```suggestion
	iterator := sdk.KVStorePrefixIterator(store, types.PostReactionsStorePrefix)
	defer iterator.Close()
```

x/posts/internal/types/events.go
```diff
@@ -4,10 +4,11 @@ package types
 const (
 	EventTypePostCreated         = "post_created"
 	EventTypePostEdited          = "post_edited"
-	EventTypeReactionAdded       = "post_reaction_added"
+	EventTypePostReactionAdded   = "post_reaction_added"
 	EventTypePostReactionRemoved = "post_reaction_removed"
 	EventTypeAnsweredPoll        = "post_poll_answered"
 	EventTypeClosePoll           = "post_poll_closed"
+	EventTypeRegisterReaction    = "register_reaction"
```
RiccardoM:
```suggestion
	EventTypeRegisterReaction    = "reaction_registered"
```

x/posts/internal/types/msg_reactions_test.go
```diff
@@ -0,0 +1,98 @@
...
+				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e"),
+			error: sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "reaction value should be a URL or an emoji unicode"),
+		},
+		{
+			name: "Invalid subspace returns error",
+			msg: types.NewMsgRegisterReaction(testOwner, ":smile:", "https://smile.jpg",
+				"1234"),
+			error: sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("reaction subspace must be a valid sha-256 hash")),
+		},
+	}
```
RiccardoM: Can we add the following case as well? I would like to make sure registering an emoji reaction works properly

```golang
{
	name: "Valid emoji value returns no error",
	msg: types.NewMsgRegisterReaction(testOwner, ":smile:", "💙", "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e"),
}
```

x/posts/internal/types/msgs_post_reactions.go
```diff
@@ -15,7 +15,7 @@ import (
 // MsgAddPostReaction defines the message to be used to add a reaction to a post
 type MsgAddPostReaction struct {
 	PostID PostID         `json:"post_id"` // Id of the post to react to
-	Value  string         `json:"value"`   // Reaction of the reaction
+	Value  string         `json:"value"`   // PostReaction of the reaction
```
RiccardoM:
```suggestion
	Value  string         `json:"value"`   // Value of the reaction
```

x/posts/internal/types/msgs_post_reactions.go
```diff
@@ -70,7 +70,7 @@ func (msg MsgAddPostReaction) GetSigners() []sdk.AccAddress {
 type MsgRemovePostReaction struct {
 	PostID   PostID         `json:"post_id"`  // Id of the post to unlike
 	User     sdk.AccAddress `json:"user"`     // Address of the user that has previously liked the post
-	Reaction string         `json:"reaction"` // Reaction of the reaction to be removed
+	Reaction string         `json:"reaction"` // PostReaction of the reaction to be removed
```
RiccardoM:
```suggestion
	Reaction string         `json:"reaction"` // Value of the reaction to be removed
```

x/posts/internal/types/msgs_post_reactions.go
```diff
@@ -99,7 +99,7 @@ func (msg MsgRemovePostReaction) ValidateBasic() error {
 	}
 
 	if len(strings.TrimSpace(msg.Reaction)) == 0 {
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Reaction value cannot be empty nor blank")
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "PostReaction value cannot be empty nor blank")
```
RiccardoM:
```suggestion
		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "Reaction value cannot be empty nor blank")
```

x/posts/internal/types/reaction.go
```diff
@@ -6,23 +6,25 @@ import (
...
-	Value string         `json:"value"` // Reaction of the reaction
+	ShortCode string
+	Value     string
+	Subspace  string
+	Creator   sdk.AccAddress
 }
 
-// NewReaction returns a new Reaction
-func NewReaction(value string, owner sdk.AccAddress) Reaction {
+// NewPostReaction returns a new PostReaction
```
RiccardoM:
```suggestion
// NewReaction returns a new Reaction
```

x/posts/internal/types/reaction.go
```diff
@@ -37,28 +39,44 @@ func (reaction Reaction) String() string {
...
 	return reaction.Value == other.Value &&
-		reaction.Owner.Equals(other.Owner)
+		reaction.ShortCode == other.ShortCode &&
+		reaction.Subspace == other.Subspace &&
+		reaction.Creator.Equals(other.Creator)
 }
 
 // ------------
-// --- Reactions
+// --- PostReactions
```
RiccardoM:
```suggestion
// --- Reactions
```

x/posts/internal/types/reaction.go
```diff
@@ -37,28 +39,44 @@ func (reaction Reaction) String() string {
...
 // ------------
 
-// Reactions represents a slice of Reaction objects
+// PostReactions represents a slice of Reaction objects
```
RiccardoM:
```suggestion
// Reactions represents a slice of Reaction objects
```

x/posts/legacy/v0.1.0/types.go
```diff
@@ -27,8 +27,8 @@ type Post struct {
 	Owner             sdk.AccAddress `json:"owner"`              // Creator of the post
 }
 
-// Reaction is a struct of a user like
+// PostReaction is a struct of a user like
```
RiccardoM: Please revert this
```suggestion
// Reaction is a struct of a user like
```

x/posts/legacy/v0.1.0/types.go
```diff
@@ -27,8 +27,8 @@ type Post struct {
...
 type Like struct {
 	Created sdk.Int        `json:"created"` // Block height at which the like was created
-	Owner   sdk.AccAddress `json:"owner"`   // User that has inserted the like
+	Owner   sdk.AccAddress `json:"owner"`   // Creator that has inserted the like
```
RiccardoM: Please revert
```suggestion
	Owner   sdk.AccAddress `json:"owner"`   // User that has inserted the like
```

x/posts/legacy/v0.2.0/types.go
```diff
@@ -65,8 +65,8 @@ func (p Post) ContentsEquals(other Post) bool {
 		p.Creator.Equals(other.Creator)
 }
 
-// Reaction is a struct of a user reaction to a post
+// PostReaction is a struct of a user reaction to a post
```
RiccardoM: Please revert
```suggestion
// Reaction is a struct of a user reaction to a post
```

x/posts/legacy/v0.2.0/types.go
```diff
@@ -65,8 +65,8 @@ func (p Post) ContentsEquals(other Post) bool {
...
 type Reaction struct {
-	Owner sdk.AccAddress `json:"owner"` // User that has created the reaction
-	Value string         `json:"value"` // Reaction of the reaction
+	Owner sdk.AccAddress `json:"owner"` // Creator that has created the reaction
```
RiccardoM: Please revert
```suggestion
	Owner sdk.AccAddress `json:"owner"` // User that has created the reaction
```

x/posts/legacy/v0.2.0/types.go
```diff
@@ -65,8 +65,8 @@ func (p Post) ContentsEquals(other Post) bool {
...
+	Value string         `json:"value"` // PostReaction of the reaction
```
RiccardoM: Please revert
```suggestion
	Value string         `json:"value"` // Reaction of the reaction
```

x/posts/legacy/v0.3.0/types.go
```diff
@@ -39,8 +39,8 @@ type Post struct {
 	Creator        sdk.AccAddress `json:"creator"`                 // Creator of the Post
 }
 
-// Reaction is a struct of a user reaction to a post
+// PostReaction is a struct of a user reaction to a post
```
RiccardoM: Please revert
```suggestion
// Reaction is a struct of a user reaction to a post
```

x/posts/legacy/v0.3.0/types.go
```diff
@@ -39,8 +39,8 @@ type Post struct {
...
 type Reaction struct {
-	Owner sdk.AccAddress `json:"owner"` // User that has created the reaction
-	Value string         `json:"value"` // Reaction of the reaction
+	Owner sdk.AccAddress `json:"owner"` // Creator that has created the reaction
```
RiccardoM: Please revert
```suggestion
	Owner sdk.AccAddress `json:"owner"` // User that has created the reaction
```

x/posts/legacy/v0.3.0/types.go
```diff
@@ -39,8 +39,8 @@ type Post struct {
...
+	Value string         `json:"value"` // PostReaction of the reaction
```
RiccardoM: Please revert
```suggestion
	Value string         `json:"value"` // Reaction of the reaction
```

cli_test/cli_posts_test.go
```diff
@@ -508,3 +522,54 @@ func TestDesmosCLIPostsReactions(t *testing.T) {
 
 	f.Cleanup()
 }
+
+func TestDesmosCLIRegisterReaction(t *testing.T) {
```
RiccardoM: Can we have a test using an emoji as value such as :tada: ?

x/posts/internal/keeper/handler.go
bragaz: this check is performed inside the `SavePostReaction` keeper method to perform the check even on genesis data

x/posts/internal/types/msg_reactions_test.go
bragaz: In order to make sure this will work we need a converter from emoji to unicode. With the go-emoji utils I can make one in a couple of minutes!

RiccardoM: I'm actually not sure why we need that. Why is so? As far as I know, emojis can be converted into bytes and stored properly inside the chain store without much problems. This is also shown by the currently stored reactions that have the form of emojis. 

Why do we now need to convert them to unicode explicitly? Aren't they converted when stored in any case?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Added the support for posting empty-message posts with medias (#110)
 - Implemented the support for hashtags in posts (#96)
 - Updated the post create CLI command in posts (#117)
+- Implemented the support for registering new emojis (#94)
```
RiccardoM:
```suggestion
- Implemented the support for registering new reactions (#94)
```

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,145 @@
...
+	defer iterator.Close()
+
+	defer iterator.Close()
```
RiccardoM:
```suggestion
```

RiccardoM:
Currently opened discussions: 

- https://github.com/desmos-labs/desmos/pull/120#discussion_r398414284
- https://github.com/desmos-labs/desmos/pull/120#discussion_r398424334
- https://github.com/desmos-labs/desmos/pull/120#discussion_r398432190

x/posts/internal/keeper/keeper_reactions.go
```diff
@@ -0,0 +1,144 @@
...
+	return reactions
+}
+
+// GetReactions allows to returns the list of reactions that have been stored inside the given context
+func (k Keeper) GetReactions(ctx sdk.Context) map[types.PostID]types.PostReactions {
+	store := ctx.KVStore(k.StoreKey)
+	iterator := sdk.KVStorePrefixIterator(store, types.PostReactionsStorePrefix)
+	defer iterator.Close()
+
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

PR #121, opened 2020-03-24
Update the post creation CLI command

### LINKED ISSUE #117: Update the post creation CLI command
## Feature description 
With #113 we've allowed to create posts with an empty message when there are some medias attached to them. This has caused the CLI command in that case to be very unnatural to use.

Currently the command is
```
desmoscli tx posts create <subspace> <message> <allows-comments> --flags
```

Which results (when an empty message is specified) to be: 
```
desmoscli tx posts create "subspace" "" true/false --flags
```

This is very unnatural for the users and we should probably fix it. 

## Implementation proposal
What I suggest to do is improve the posting command by making the message an optional flag instead of a required argument: 

```
desmoscli tx posts create <subspace> <allows-comments> \
  --message="My message"
  --flags
```

Of course, with this implementation inside the CLI tx command handler we need to check that the message is not empty when the medias are, and vice-versa, to make sure that at least one between the message and the medias are specified.

bragaz:
## Description
Updating the CLI create post's command to better fit the latest edits. Now the message field is optional and the allow-comments one become a flag.
Closes #117 

## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/cli_test/cli_posts_test.go
+++ b/cli_test/cli_posts_test.go
@@ -32,7 +32,7 @@ func TestDesmosCLIPostsCreateNoMediasNoPollData(t *testing.T) {
 	require.Equal(t, startTokens, fooAcc.GetCoins().AmountOf(denom))
 
 	// Create a post
-	success, _, sterr := f.TxPostsCreate(subspace, message, true, fooAddr, "-y")
+	success, _, sterr := f.TxPostsCreate(subspace, message, fooAddr, "-y")
 	require.True(t, success)
 	require.Empty(t, sterr)
 	tests.WaitForNextNBlocksTM(1, f.Port)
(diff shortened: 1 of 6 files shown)


x/posts/client/cli/tx.go
```diff
@@ -206,6 +220,7 @@ E.g.
 		},
 	}
 
+	cmd.Flags().Bool(flagAllowsComments, true, "Possibility to comment post or not")
```
RiccardoM:
```suggestion
	cmd.Flags().Bool(flagAllowsComments, true, "Possibility to comment the post or not")
```

cli_test/cli_posts_test.go
```diff
@@ -32,7 +32,7 @@ func TestDesmosCLIPostsCreateNoMediasNoPollData(t *testing.T) {
 	require.Equal(t, startTokens, fooAcc.GetCoins().AmountOf(denom))
 
 	// Create a post
-	success, _, sterr := f.TxPostsCreate(subspace, message, true, fooAddr, "-y")
+	success, _, sterr := f.TxPostsCreate(subspace, message, fooAddr, "-y")
```
RiccardoM: Can we also add a new test set using `--allows-comments=false` and making sure that the message is generated/sent properly?

PR #132, opened 2020-04-14
Post ID improvement

### LINKED ISSUE #131: Changing the way PostIDs are created
## Context
Currently when creating a new `Post`, it's `ID` field is generated sequentially by taking the last id and adding `1` to it. 

While this surely works, it is just a very naive way of generating ids in a very determininistic way that might have some downsides as newer ids can be inferred. 

A part from this, a serie of checks must be put in place in order to maintain the ordering of such ids, as I've also outlined inside #90. 

## Solution
In order to properly change this, I propose a new way of deterministically generating post IDs that cannot be infered and also will allow for some performance improvements: 

```
sha-256(post_creation_date + post_creator + post_subspace) 
```

This way we could also remove the current way of checking if a post collides with another one which is currently causing a lot of performance probles as we iterate over all the posts. 

Using this ID system, instead, we could simply check if such key already exists inside the store and if it does, it means the post is duplicated. 

The implementation could be done directly inside the `NewPost` method, generating the ID there instead of outside of it. 

## Notes 
When implementing this, a migration should also be implemented to transform `0.3.0` post ids into the new type by performing the appropriate convertion.

bragaz:
## Description
This PR contains some core changes to the way the Post's IDs are created.
This has been done to optimize the creation and edit of posts.
Closes #131.

## Checklist
- [x] Wrote unit tests.


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Version 0.4.0
 ## Changes
+- Improved `postID` generation (#131)
```
RiccardoM, line 3:
```suggestion
- Improved the generation of post ids (#131)
```

docs/developers/queries/poll-answers.md
```diff
@@ -6,13 +6,13 @@ This query endpoint allows you to retrieve the details of answers made to a post
...
+# desmoscli query posts poll-answers a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc
 ``` 
 
 **REST**
 ```
 /posts/{postId}/poll-answers
 
 # Example
-# curl https://morpheus1000.desmos.network/posts/10/poll-answers
+# curl https://morpheus4000.desmos.network/posts/10/poll-answers
```
RiccardoM, line 17:
```suggestion
# curl http://lcd.morpheus.desmos.network:1317/posts/10/poll-answers
```

docs/developers/queries/post.md
```diff
@@ -6,13 +6,13 @@ This query endpoint allows you to retrieve the details of a single post having i
...
+# desmoscli query posts post a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc
 ``` 
 
 **REST**
 ```
 /posts/{postId}
 
 # Example
-# curl https://morpheus1000.desmos.network/posts/10
+# curl https://morpheus4000.desmos.network/posts/a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc
```
RiccardoM, line 17:
```suggestion
# curl http://lcd.morpheus.desmos.network:1317/posts/a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc
```

docs/developers/queries/posts.md
```diff
@@ -42,5 +42,5 @@ Available parameters:
 
 ```bash
 # Example
-# curl https://morpheus1000.desmos.network/posts?parent_id=1&allows_comments=true&subspace=desmos&sort_by=created&sort_order=descending
+# curl https://morpheus4000.desmos.network/posts?parent_id=a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc&allows_comments=true&subspace=desmos&sort_by=created&sort_order=descending
```
RiccardoM, line 45:
```suggestion
# curl http://lcd.morpheus.desmos.network:1317/posts?parent_id=a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc&allows_comments=true&subspace=desmos&sort_by=created&sort_order=descending
```

docs/developers/queries/session.md
```diff
@@ -14,5 +14,5 @@ desmoscli query sessions session [id]
 /sessions/{sessionId}
 
 # Example
-# curl https://morpheus1000.desmos.network/sessions/66
+# curl https://morpheus4000.desmos.network/sessions/66
```
RiccardoM, line 17:
```suggestion
# curl http://lcd.morpheus.desmos.network:1317/sessions/66
```

x/genutil/legacy/v0.4.0/migrate.go
```diff
@@ -0,0 +1,49 @@
...
+	v030posts "github.com/desmos-labs/desmos/x/posts/legacy/v0.3.0"
+	v040posts "github.com/desmos-labs/desmos/x/posts/legacy/v0.4.0"
+)
+
+// Migrate migrates exported state from v0.3.0 to a v0.4.0 genesis state.
+// It requires args to contain an integer value representing the block interval that should be considered when
+// converting block height-based timestamps into time.Time timestamps.
+func Migrate(appState genutil.AppMap, _ ...interface{}) genutil.AppMap {
+	// Perform the Cosmos SDK migration first
+	appState = cosmosv0380.Migrate(appState)
```
RiccardoM, line 18: No need for this as the `0.3.0` is already based on Cosmos `0.38`

```suggestion
```

x/genutil/legacy/v0.4.0/migrate.go
```diff
@@ -0,0 +1,49 @@
...
+	// Migrate magpie state
+	if appState[v030magpie.ModuleName] != nil {
+		var genDocs v030magpie.GenesisState
+		v030Codec.MustUnmarshalJSON(appState[v030magpie.ModuleName], &genDocs)
+
+		delete(appState, v030magpie.ModuleName) // delete old key in case the name changed
+		appState[v040magpie.ModuleName] = v040Codec.MustMarshalJSON(
+			v040magpie.Migrate(genDocs),
+		)
+	}
```
RiccardoM, line 35: I don't think we need to put this as there is nothing to migrate for the `magpie` module I guess

```suggestion
```

x/magpie/legacy/v0.4.0/migrate.go
```diff
@@ -0,0 +1,31 @@
+package v040
```
RiccardoM, line 1: Probably we can remove this entire file

x/magpie/legacy/v0.4.0/types.go
```diff
@@ -0,0 +1,34 @@
+package v040
```
RiccardoM, line 1: Probably we can remove this as well

x/posts/client/cli/query.go
```diff
@@ -93,12 +93,11 @@ $ %s query posts posts --page=2 --limit=100
 
 			// ParentID
 			if parentID := viper.GetString(flagParentID); len(parentID) > 0 {
-				parentID, err := types.ParsePostID(parentID)
-				if err != nil {
+				pID := types.PostID(parentID)
```
RiccardoM, line 96: Can we rename `pID` to be `parentID` or something more explicit?

x/posts/client/rest/query.go
```diff
@@ -77,12 +77,12 @@ func queryPostsWithParameterHandlerFn(cliCtx context.CLIContext) http.HandlerFun
 		}
 
 		if v := r.URL.Query().Get(RestParentID); len(v) != 0 {
-			parsedParentID, err := types.ParsePostID(v)
-			if err != nil {
+			id := types.PostID(v)
```
RiccardoM, line 80: Can we rename `id` to be `parentID` or something else more explicit?

x/posts/internal/types/post.go
```diff
@@ -1,77 +1,51 @@
...
+func ComputeID(creationDate time.Time, creator sdk.AccAddress, subspace string) PostID {
+	hash := sha256.Sum256([]byte(creationDate.String() + creator.String() + subspace))
+	return PostID(hex.EncodeToString(hash[:]))
 }
 
-// Next returns the subsequent id to this one
-func (id PostID) Next() PostID {
-	return id + 1
+// Valid tells if the id can be used safely
+func (id PostID) Valid() error {
```
RiccardoM, line 32: I'm really not a big fan of this method returning an `error`. Since the error can only be one (the post id is not valid), it can be much easier be represented by a `bool` return value instead.

Can we switch it back to it returning `bool` instead of `err`? This should also be much more easier to use.

x/posts/internal/types/post.go
```diff
@@ -334,8 +296,8 @@ type Posts []Post
 func (p Posts) String() string {
 	out := "ID - [Creator] Message\n"
 	for _, post := range p {
-		out += fmt.Sprintf("%d - [%s] %s\n",
-			post.PostID, post.Creator, post.Message)
+		out += fmt.Sprintf("%s - [%s] %s\n",
+			post.PostID.String(), post.Creator, post.Message)
```
RiccardoM, line 300: Since `PostID` implements the `String()` method, we can remove the call here as it will be implicit

```suggestion
			post.PostID, post.Creator, post.Message)
```

x/posts/internal/types/querier.go
```diff
@@ -14,7 +14,7 @@ type QueryPostsParams struct {
 	SortBy    string // Field that should determine the sorting
 	SortOrder string // Either ascending or descending
 
-	ParentID       *PostID
+	ParentID       PostID
```
RiccardoM, line 17: Why has this been changed to `PostID`? Couldn't we leave it `*PostID` for better understanding?

x/posts/internal/types/post_sorting.go
```diff
@@ -1,7 +1,6 @@
 package types
 
 const (
-	PostSortByID           = "id"
```
RiccardoM, line 4: I think we can leave it if a user wants to sort it by ID for any reason

x/posts/internal/keeper/keeper.go
```diff
@@ -161,18 +143,9 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
...
-			result = first.Created.Before(second.Created)
-			if params.SortOrder == types.PostSortOrderDescending {
-				result = first.Created.After(second.Created)
-			}
-
-		default:
-			result = first.PostID < second.PostID
-			if params.SortOrder == types.PostSortOrderDescending {
-				result = first.PostID > second.PostID
-			}
```
RiccardoM, line 175: I think we should revert this change as the ID sorting should be supported in any case

x/posts/internal/keeper/keeper.go
```diff
@@ -114,7 +96,7 @@ func (k Keeper) GetPostsFiltered(ctx sdk.Context, params types.QueryPostsParams)
 		matchParentID, matchCreationTime, matchAllowsComments, matchSubspace, matchCreator, matchHashtags := true, true, true, true, true, true
 
 		// match parent id if valid
-		if params.ParentID != nil {
+		if params.ParentID != "" {
```
RiccardoM, line 99: I think it's better to undo this change as a user might want to search for all the posts that do not have a parent (and so `params.ParentID == ""`) . 

There's always a difference between a `params.ParentID == nil` (if a user has not specified any filtering) and `params.ParentID = ""` (if a user wants to search all posts without a parent).

x/posts/internal/types/post.go
bragaz: I've switch to this cause there's a lot of repetitions of the same error string, so I think it would be better doing it like this..

PR #136, opened 2020-04-22
Generic improvements

### LINKED ISSUE #125: Improve storing and reading
## Context
While storing and reading posts, poll answers, reactions and other on-chain data, we are currently doing it in a very naive method: 

- when reading data, we iterate over all the set to find the data that we most like
- when writing data, we replace the entire existing slice with a new one to which we have added the new element 

While this methods surely work, in the latest transactions they are now showing some weak points. As reading and writing cost gas while performing a transaction, in fact, some recent transactions have been failing due to gas shortage (you can see it even [here](https://morpheus.desmos.network/transactions/952E999BA51BB32F20AF079F831F285D2609E006AB01DF367547F701FF5B3528) and [here](https://morpheus.desmos.network/transactions/FC85D7C0F8FAA06F1F4FD835A753069E80EB239649C9265777791058CC50F69A)). 

In order to make sure that our transactions do not cost too much, we should probably improve how we read and write data. 

## Possible solutions
Below, there are some possible solutions I've thought that might help us out with these problems. 

### Reading
When finding a specific post, reaction or similar inside a slice, instead of iterating over all the slice itself we should probably do it using a more efficient algorithm. An example could be the [binary search](https://en.wikipedia.org/wiki/Binary_search_algorithm) algorithm, which is already implemented inside Go using the [`sort.Search method`](https://flaviocopes.com/golang-algorithms-binary-search/).

Also, another idea to further improve the reading time could be the one to keep the data sorted when saving it. This way the search might be faster.

### Writing 
Currently I have no idea on how we could improve the writing speed. 

## Conclusion
I'd like @bragaz and @kwunyeung ideas on these approaches.

### LINKED ISSUE #137: Change how reactions can be added
## Context
Currently when a user wants to add a reaction, he is able to do so only by specifying its short code, which is in the format of `:code:` and needs to correspond to a previously registered reaction. 

While this method is very good when wanting to add custom reactions (such as GIFs and pictures), it is not that good when wanting to use emojis. Following an example of what problems such approach might cause. 

Suppose we have an application client (i.e. Mooncake) that allows the user to use whatever emoji, and at the same time the emoji shortcodes are generated from a third party library.

Now, consider this: 

1. The user wants to add the "🤩" reaction.
2. Such reaction is converted by the third party library to the `:grinning_face_with_star_eyes:` shortcode. 
3. The same reaction is registered inside Desmos using the `:start-struck:` shortcode instead. 

What the client app should do is: 

1. Fetch the registered reactions list. 
2. Inverse search through all the list of reactions to find the one having unicode "🤩"
3. Use the registered code instead of the one provided by the library. 

While this is generically a good practice, it requires a network call for something that could be done completely offline. 

## Solution
To solve this problem, what we should do is allow the usage of emojis while adding a reaction to a post. In this case, we should extend the handling of `MsgAddPostReaction` to also allow values like "🤩" to be put inside the reaction value field. 

Then, the handling of the message should be extended as follow: 

1. If the reaction value is an emoji, get the shortcode of the emoji and use that shortcode while storing the reaction inside the state. 

2. If the reaction value is a shortcode, proceed as it currently is.

### LINKED ISSUE #138: Auto registration for common emojis
## Context 
With the current system, users are able to register whatever reaction they want. Unluckily, I've realized that this system does not take into consideration the fact that some reactions like emojis are widely used and should be registered by default into any subspace. 

Allowing the users to individually register emoji could lead to some scenario in which the user registers a common emoji and later changes its value leading to a series of problems. 

## Solution 
To solve this problem, I suggest we change how the registration of reactions are done as follows. 

1. Avoid users registering single emojis. 
   What we should do is check the value of the registered reaction, and if it's a single emoji return an error telling the user he cannot register that. 

2. Auto register common emojis during a reaction addition. 
  Once #137 is implemented, we can auto registered new reactions when they are used during the handling of `MsgAddPostReaction` messages. This way we can make sure the reactions are always registered and that the registrar is the module itself, disallowing any future change of the reaction value of code from users.

### LINKED ISSUE #140: Improve how posts are returned to the clients
> Taken from a [conversation](https://github.com/desmos-labs/desmos/pull/136#issuecomment-618424093) which happened inside #136

While taking a look at the code, I thought that maybe we want to improve how the reactions list is returned to the user when he asks for a posts list of a post details. Currently the returned object is 

```golang
	Post
	PollAnswers []UserAnswer  `json:"poll_answers,omitempty"`
	Reactions   PostReactions `json:"reactions"`
	Children    PostIDs       `json:"children"`
}
```

What if we make it as follows?

```golang
	Post
	PollAnswers []UserAnswer           `json:"poll_answers,omitempty"`
	Reactions   PostReactionsResponses `json:"reactions"`
	Children    PostIDs                `json:"children"`
}
```

Each object of `PostReactionsResponses` should be defined as:

```golang
type PostReactionsResponse struct { 
	Value string         `json:"value"`
	Code string          `json:"code"`
 	User sdk.AccAddress  `json:"user"`
}
```

This way I think the client could handle the reaction better having both the code and the value (emoji or URI). What do you think @bragaz?

RiccardoM:
## Description
Contains some generic improvements that should make sure the code is ready to be tagged for `0.4.0`. 

- Make sure the gas usage does not exceed 200.000 units (closes #125) 
- Improved the adding of reactions (closes #137) 
- Improved how emoji reactions are handled (closes #138)
- Fixed some bugs
- Improved the simulation tests
- Improved the download of golint while using Go 1.14 or later
- Fixed the documentation
- Improved how posts are returned to the client when querying them (closes #140)

## Checklist
- [x] Wrote unit tests.


RiccardoM:
@bragaz I just thought that maybe we want to improve how the reactions list is returned to the user when he asks for a posts list of a post details. Currently the returned object is 

```golang
	Post
	PollAnswers []UserAnswer  `json:"poll_answers,omitempty"`
	Reactions   PostReactions `json:"reactions"`
	Children    PostIDs       `json:"children"`
}
```

What if we make it 

```golang
	Post
	PollAnswers []UserAnswer           `json:"poll_answers,omitempty"`
	Reactions   PostReactionsResponses `json:"reactions"`
	Children    PostIDs                `json:"children"`
}
```

Where each object of `PostReactionsResponses` is made as follows? 

```golang
type PostReactionsResponse { 
	Value string         `json:"value"`
	Code string          `json:"code"`
 	User sdk.AccAddress  `json:"user"`
}
```

This way I think the client could handle the reaction better having both the code and the value (emoji or URI). What do you think?

bragaz:
Currently the `PostReaction` type has the following signature:
```golang
// PostReaction is a struct of a user reaction to a post
type PostReaction struct {
	Owner sdk.AccAddress `json:"owner"` // Creator that has created the reaction
	Value string         `json:"value"` // PostReaction of the reaction
}
```
So in order to do that we should firstly get all the registered reaction (with the current `ListReaction(ctx sdk.Context)`) and then iterate over them to get the corresponded value associated with the shortcode of the `postReaction` which sounds crazy and would cost a lot of gas probably.

I think that a way to simplify the life of clients could be improve our actual `queryRegisteredReactions` by returning something like a `RegisteredReactionResponse`with which they can better search for the value associated to the shortcode instead of the simple `registeredReactions` slice.

PR #149, opened 2020-05-05
Allow empty messages with non-empty polls

### LINKED ISSUE #148: Allow for empty message posts if the poll is specified
## Context
Currently the user cannot create an empty message post which contains a message, because the message is marked optional only if the medias are not empty. 

We should definitely allow the creation of empty messages posts if they contain a poll. 

## Solution
In order to allow this, we should reduce the `tx posts create` command required arguments from 2 to 1, making the message optional. 

Then we need to check in both the `ValidateBasic` method of `MsgCreatePost` and `Validate` of `Post` if the post has either a message, a poll or at least a media to be valid.

bragaz:
## Description
This PR allows the creation of posts with empty message if a poll inside them is specified.
Closes #148.

## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/x/posts/client/cli/tx.go
+++ b/x/posts/client/cli/tx.go
@@ -91,6 +91,8 @@ If you want to add a poll to your post you need to specify it through two flags:
   2. --poll-answer, which accepts a slice of answers that will be provided to the users once they want to take part in the poll votations.	
      Each answer should be identified by the text of the answer itself.
 
+If polls are provided, the post could be created even without any message as following:
+
 E.g.
 %s tx posts create "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e" "Post with poll" true \
 	--poll-details "question=Which dog do you prefer?,multiple-answers=false,allows-answer-edits=true,end-date=2020-01-01T15:00:00.000Z" \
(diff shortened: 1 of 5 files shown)


x/posts/client/cli/tx.go
```diff
@@ -91,6 +91,8 @@ If you want to add a poll to your post you need to specify it through two flags:
   2. --poll-answer, which accepts a slice of answers that will be provided to the users once they want to take part in the poll votations.	
      Each answer should be identified by the text of the answer itself.
 
+If polls are provided, the post could be created even without any message as following:
```
RiccardoM, line 94:
```suggestion
If a poll is provided, the post can be created even without specifying any message as following:
```

x/posts/client/cli/tx.go
```diff
@@ -91,6 +91,8 @@ If you want to add a poll to your post you need to specify it through two flags:
   2. --poll-answer, which accepts a slice of answers that will be provided to the users once they want to take part in the poll votations.	
      Each answer should be identified by the text of the answer itself.
 
+If a poll is provided, the post can be created even without specifying any message as following:
```
RiccardoM, line 94:
```suggestion
If a poll is provided, the post can be created even without specifying any message as follows:
```

PR #150, opened 2020-05-05
Generic reactions improvements

### LINKED ISSUE #144: Improve reactions events
## Context
Currently when a user adds or removes a reaction to a post, the following event is emitted: 

```golang
sdk.NewEvent(
	types.EventTypePostReactionAdded,
	sdk.NewAttribute(types.AttributeKeyPostID, msg.PostID.String()),
	sdk.NewAttribute(types.AttributeKeyPostReactionOwner, msg.User.String()),
	sdk.NewAttribute(types.AttributeKeyReactionShortCode, reactionValue),
)
```

As we can see, the event contains only the associated short code of the reaction that is added to the post. While this might be ok, it presents two problems: 

1. If the user adds ":smile:" as a reaction, the event will instead contain `:smile:`, which might be confusing to watch from an external point of view cause it doesn't tell where that code comes from and where the value has gone. 

2. If the user needs to have both the added value as well as the added shortcode, he will need to parse the event and the message, which is pretty strange. 

## Solution
In order to improve the usability of such events, what I suggest we should do is adding another attribute to both the event emitted when processing `MsgAddPostReaction` and the one emitted when processing `MsgRemovePostReaction` so that both contain: 

- the `shortcode` of the added reaction
- the `value` (either emoji or URI) of the added reaction. This can be read from the store (if already registered) or from the value (if an emoji).

### LINKED ISSUE #145: Remove the automatic registration of emoji reactions
## Context
Currently, when an emoji is used a reaction to a post, a new reaction is registered inside the state to make sure that when it's stored, no errors are returned. While this methods works great and ensures that each reaction is registered before being added to a post, it also has one big downside: duplication. 

When we register reactions, each one of them is associated to a subspace which identifies a client application. This means that if we consider having 10 different client applications, and inside each one of them the ":smile:" reaction is used, we will have 10 different registered reaction with the same shortcode and the same value, but a different subspace each time. Also, if there's one emoji which has multiple shortcodes, this will be registered multiple times for the same client. 

While this might seem like a small thing, considering that there are currently more than 3,000 emojis this lead in the future to an unnecessary increased size in the store. 

## Solution 
A possible solution to this problem is changing how reactions are currently handled inside Desmos. This could be done as follows. 

Once a user adds a reaction to a post, the following operations will take place: 

1. If the reaction value is a shortcode but does not represent an emoji, check if the shortcode is registered. 
   If not throw an error. 
   Otherwise, store inside the state the shortcode associated to the reaction. 
2. If the reaction is an emoji, get the shortcode associated to it (if there are multiple, pick the first) and store inside the state the associated shortcode. 
3. In both cases, emit both the reaction value and shortcode. 

Once a user removes a reaction to a post, the following operations will take place: 

1. If the reaction value is a shortcode: 
   1. If the shortcode represents a registered reaction, search a reaction with the same code. 
   2. If the shortcode represents an emoji, get the emoji with that shortocde and search all the reactions with either one of the associated shortcodes. 
2. If the reaction value is an emoji, get the emoji with that shortcode and remove all the reactions having either one of the shortcodes as value. 

This way we will be able to have a store inside which the reactions are **all represented by their shortcodes** which will allow us in a later time to get both the `value` and `shortocode` when returning them to the clients that asks them. 

### Notes
Note that emoji-based reactions are not checked for registration, thus reducing the size of the used disk space as well as making the whole process a lot more faster. 

## Conclusion
I think this system will perform better in a long run, but I want to hear if you have suggestions about it @kwunyeung @bragaz

### LINKED ISSUE #147: Improve the registration error message
## Context
Currently when sending a `MsgRegisterReaction` if the reaction shortcode is not valid the following error is returned to the user: 

```
'invalid request: reaction short code must be an emoji short code'
```

This does not tell him what needs to be changed and how. 

## Solution 
To solve this problem, I think we should return a more descriptive message such as :+1: 

> The specified shortcode is not valid. To be valid it must only contains a-z, 0-9, - and _ and must start and end with a :

bragaz:
## Description
This PR fixes the problems discussed in issues #144 #145 #147.
Closes #144.
Closes #145.
Closes #147. 

## Checklist
- [x] Wrote unit tests.


x/posts/internal/keeper/handler.go
```diff
@@ -144,18 +143,33 @@ func handleMsgAddPostReaction(ctx sdk.Context, keeper Keeper, msg types.MsgAddPo
...
+				return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+					fmt.Sprintf("reaction with short code %s isn't registered yet and can't be used to react to the post with ID %s and subspace %s, please register it before use",
+						parsedReaction, post.PostID, post.Subspace))
+			}
+			reactionValue = regReaction.Value
+			reactionShortcode = regReaction.ShortCode
+		} else {
+			reactionValue = emo.Value
+			reactionShortcode = emo.Shortcodes[0]
+		}
```
RiccardoM, line 169: I'm not a big fan of this piece of code's structure. I think it can be improve by doing something like 

```golang
if emojiReact, err := emoji.LookupEmoji(parsedReaction); err == nil {
  // The reaction was an emoji Unicode character
} else if emojiReact, err := emoji.LookupEmojiByCode(parsedReaction); err == nil {
  // The reaction was an emoji shotcode
} else {
  // The reaction was a shortcode that should be registered
}
```

I think it would be more easy to read. Also, since this entire piece of code is used as well inside the `handleMsgRemovePostReaction`, I would avoid duplicating it. To do this, we should move it into an external function.

bragaz: yeah I was not so conviced as well but I couldn’t think a better way to do that in that moment so I was hoping for some advice from you!

PR #153, opened 2020-05-07
Post Media tags implementation

### LINKED ISSUE #118: Implement media tagging
## Feature description 
Currently, when creating a media the user can specified it's `URI` and `MimeType`. In order to better support media sharing (particularly images) on a wide range of social networks, we should implement media tagging. Thanks to this features, users will be able to tag other users on a particular image to make sure they are someone recognizes when that image is viewed. 

## Implementation proposal
In order to implement media tagging, the following changes should be made. 

### Update he PostMedia type
First of all, we need to update the `PostMedia` type to be: 

```golang
type PostMedia struct {
	URI      string           `json:"uri"`
	MimeType string           `json:"mime_type"`
        Tags     []sdk.AccAddress `json:"tags"`     // Contains the list of all the tagged users
}
```

### Update the CLI and REST commands
Then we need to update the CLI and REST commands to allow tagging. The most particular one will be the CLI one. I think the `--media` flag can be changed to something like:

```
--media "<URI>,<MimeType>,<Tags...>
```

So an example could be

```
--media "https://example.com,image/jpeg,desmos1p7c8h59nrc8e5hxvgvu2g7tpp0xwn4mzevzgg7,desmos1p7ad878nealg249qkkdl9ldxrllst23lklngcx
```

What are your thoughts on this @bragaz @kwunyeung?

bragaz:
## Description
This PR implements the post media tagging by adding a new field `tags` of type `[]sdk.AccAddress` into the `PostMedia` type.
Closes #118.
## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/cli_test/cli_posts_test.go
+++ b/cli_test/cli_posts_test.go
@@ -208,15 +208,10 @@ func TestDesmosCLIPostsCreateWithMediasAndNonEmptyMessage(t *testing.T) {
 	startTokens := sdk.TokensFromConsensusPower(140)
 	require.Equal(t, startTokens, fooAcc.GetCoins().AmountOf(denom))
 
-	tag, err := sdk.AccAddressFromBech32("desmos15ux5mc98jlhsg30dzwwv06ftjs82uy4g3t99ru")
-	require.NoError(t, err)
-	tag2, err2 := sdk.AccAddressFromBech32("desmos1ulmv2dyc8zjmhk9zlsq4ajpudwc8zjfm82aysr")
-	require.NoError(t, err2)
-
 	// Create a post
 	success, _, sterr := f.TxPostsCreate(subspace, message, fooAddr, "-y",
-		"--media https://example.com/media1,text/plain,desmos15ux5mc98jlhsg30dzwwv06ftjs82uy4g3t99ru",
-		"--media https://example.com/media2,application/json,desmos1ulmv2dyc8zjmhk9zlsq4ajpudwc8zjfm82aysr",
+		"--media https://example.com/media1,text/plain",
+		"--media https://example.com/media2,application/json",
 		"--media https://example.com/media3,text/plain",
 	)
 	require.True(t, success)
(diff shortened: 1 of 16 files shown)


bragaz:
There's also migration v0.5.0 files cause I've wrongly pulled from the `v050` branch instead of master when updating mine

cli_test/cli_posts_test.go
```diff
@@ -141,10 +141,15 @@ func TestDesmosCLIPostsCreateWithMediasAndEmptyMessage(t *testing.T) {
...
+	require.NoError(t, err)
+	tag2, err2 := sdk.AccAddressFromBech32("desmos1ulmv2dyc8zjmhk9zlsq4ajpudwc8zjfm82aysr")
+	require.NoError(t, err2)
+
 	// Create a post
 	success, _, sterr := f.TxPostsCreate(subspace, message, fooAddr, "-y",
-		"--media https://example.com/media1,text/plain",
-		"--media https://example.com/media2,application/json")
+		"--media https://example.com/media1,text/plain,desmos15ux5mc98jlhsg30dzwwv06ftjs82uy4g3t99ru",
+		"--media https://example.com/media2,application/json,desmos1ulmv2dyc8zjmhk9zlsq4ajpudwc8zjfm82aysr")
```
RiccardoM, line 152: Can we also keep some tests **without** the tags to make sure they keep working for backwards compatibility?

x/posts/internal/simulation/utils.go
```diff
@@ -125,14 +125,16 @@ func RandomHashtag(r *rand.Rand) string {
...
 	mediaNumber := r.Intn(20)
 
+	acc, _ := sim.RandomAcc(r, accs)
+
 	postMedias := make(types.PostMedias, mediaNumber)
 	for i := 0; i < mediaNumber; i++ {
 		host := RandomHosts[r.Intn(len(RandomHosts))]
 		mimeType := RandomMimeTypes[r.Intn(len(RandomMimeTypes))]
-		postMedias[i] = types.NewPostMedia(host+strconv.Itoa(i), mimeType)
+		postMedias[i] = types.NewPostMedia(host+strconv.Itoa(i), mimeType, []sdk.AccAddress{acc.Address})
```
RiccardoM, line 137: I think we should probably randomize even the number of tags to be used here. So: 

1. Generate a random int `x`
2. Get `x` accounts
3. Tag each account into the media

x/posts/internal/types/models/common/post_media_test.go
```diff
@@ -225,43 +294,61 @@ func TestPostMedias_Validate(t *testing.T) {
...
+		},
+		{
+			name: "No errors media",
 			postMedia: common.PostMedia{
 				URI:      "https://example.com",
 				MimeType: "text/plain",
+				Tags:     []sdk.AccAddress{tag},
 			},
 			expErr: "",
 		},
```
RiccardoM, line 346: Can we add a test case to make sure the validate works even with empty tags?

x/posts/internal/types/models/common/post_media_test.go
```diff
@@ -69,35 +86,87 @@ func TestPostMedias_Equals(t *testing.T) {
...
 					MimeType: "text/plain",
+					Tags:     []sdk.AccAddress{tag},
+				},
+				common.PostMedia{
+					URI:      "uri",
+					MimeType: "application/json",
+					Tags:     []sdk.AccAddress{tag2},
 				},
 			},
 			expEquals: false,
```
RiccardoM, line 172: Can we add a test case to make sure two medias with empty tags return `true` if they are equals?

x/posts/internal/types/models/post_response_test.go
```diff
@@ -78,7 +78,7 @@ func TestPostQueryResponse_MarshalJSON(t *testing.T) {
...
+			expResponse: `{"id":"dd065b70feb810a8c6f535cf670fe6e3534085221fa964ed2660ebca93f910d1","parent_id":"","message":"Post","created":"2020-02-02T15:00:00Z","last_edited":"0001-01-01T00:00:00Z","allows_comments":true,"subspace":"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e","creator":"cosmos1y54exmx84cqtasvjnskf9f63djuuj68p7hqf47","medias":[{"uri":"https://uri.com","mime_type":"text/plain","tags":["cosmos1y54exmx84cqtasvjnskf9f63djuuj68p7hqf47"]}],"poll_data":{"question":"poll?","provided_answers":[{"id":"1","text":"Yes"},{"id":"2","text":"No"}],"end_date":"2050-01-01T15:15:00Z","is_open":true,"allows_multiple_answers":false,"allows_answer_edits":true},"poll_answers":[{"answers":["1"],"user":"cosmos1s3nh6tafl4amaxkke9kdejhp09lk93g9ev39r4"}],"reactions":[{"value":"https://example.com/like","shortcode":":like:","owner":"cosmos1s3nh6tafl4amaxkke9kdejhp09lk93g9ev39r4"},{"value":"👍","shortcode":":+1:","owner":"cosmos15lt0mflt6j9a9auj7yl3p20xec4xvljge0zhae"}],"children":["dd065b70feb810a8c6f535cf670fe6e3534085221fa964ed2660ebca93f910d1","dd065b70feb810a8c6f535cf670fe6e3534085221fa964ed2660ebca93f910d1"]}`,
```
RiccardoM, line 81: Can we also add a test case with a media that has no tags?

cli_test/cli_posts_test.go
RiccardoM: This is still unsolved. I think what we can do is remove the tags from within `TestDesmosCLIPostsCreateWithMediasAndNonEmptyMessage` to have it implemented.

PR #154, opened 2020-05-08
RocksDB implementation

### LINKED ISSUE #111: Switching to RocksDB as the default db backend
Closes #97 

The latest version of `Tendermint` added an option for using `RocksDB` as the db backend.

https://github.com/tendermint/tendermint/pull/4239

which is available in `SDK v0.38.x`

`RocksDB` was created by `Facebook` as a high performance implementation of `LevelDB`.

https://github.com/facebook/rocksdb/wiki

Some benchmarks in earlier years have shown that although `RocksDB` would create a large DB in size but with better performance in read/white/delete operations.

https://www.influxdata.com/blog/benchmarking-leveldb-vs-rocksdb-vs-hyperleveldb-vs-lmdb-performance-for-influxdb/

With a similar use case as `Facebook`, I suggest we should try using `RocksDB` as the default db backend. 

Of course we have to study if there is any drawback after adapting it. How the performance gain cost resources usage, would it increase difficulties running fullnodes, etc.

bragaz:
## Description
This PR implements the rocksDB as desmos backendDB.
Closes #111.
## Checklist
- [ ] Wrote unit tests.


Makefile
```diff
@@ -26,9 +26,9 @@ ifneq ($(GOSUM),)
   ldflags += -X github.com/cosmos/cosmos-sdk/version.VendorDirHash=$(shell $(GOSUM) go.sum)
 endif
 
-ifeq ($(WITH_CLEVELDB),yes)
+ifeq ($(WITH_ROCKSDB),yes)
   build_tags += gcc
-  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=cleveldb
+  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
```
RiccardoM, line 31: I think that by doing this we are not using RocksDB as the default DB backend (as specified inside #111). Instead, we are allowing people to use it if they want when compiling the Desmos binaries. 

I think the correct way of implementing #111 is by reverting these changes and changing the original

```
ifeq ($(WITH_CLEVELDB),yes)
  build_tags += gcc	 
  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=cleveldb
endif
```

To be 

```
ifeq ($(DB_BACKEND),cleveldb)
  build_tags += gcc	 
  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=cleveldb
else ifeq($(DB_BACKEND),leveldb)
  // Do nothing as LevelDB is the default one inside Cosmos SDK
else
  build_tags += gcc	 
  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
endif
```

This would probably make RocksDB the default system in our case while still allowing users to go back to LevelDB or CLevelDB

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.6.0
+## Changes
+- Switched backendDB from golevelDB to rocksDB (#111)
```
RiccardoM, line 3:
```suggestion
- Changed default database backend from [LevelDB](https://github.com/google/leveldb) to [RocksDB](https://github.com/facebook/rocksdb) (#111)
```

Makefile
```diff
@@ -26,9 +26,14 @@ ifneq ($(GOSUM),)
...
 
-ifeq ($(WITH_CLEVELDB),yes)
+ifeq ($(DB_BACKEND),cleveldb)
   build_tags += gcc
   ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=cleveldb
+else ifeq($(DB_BACKEND),leveldb)
+ # Do nothing as LevelDB is the default backendDB of Cosmos SDK
+else
+build_tags += gcc
+ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
```
RiccardoM, line 36: I've tried the following three commands: 

1. `make install DB_BACKEND=cleveldb` 
2. `make install DB_BACKEND=leveldb`
3. `make install` 

Number 1 and 2 executed properly, but number 3 didn't. 

I think this is due to the spaces that are being put inside the branches conditions and this should fix it: 

```suggestion
ifeq ($(DB_BACKEND),cleveldb)
  build_tags += gcc
  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=cleveldb
else ifeq ($(DB_BACKEND),leveldb)
  # Do nothing as LevelDB is the default backendDB of Cosmos SDK
else ifeq ($(DB_BACKEND),)
  build_tags += gcc
  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
endif
```

PR #179, opened 2020-05-22
Post Report implementation

### LINKED ISSUE #50: Posts can be flagged and owners can appeal
A user can flag inappropriate of messages. Conditions of hiding messages should be implemented. For example, the message will be hidden if at least be flagged by 2 individuals.

The user of the flagged message can appeal. The appeal should be juried by the community, maybe via some kind of governance?

bragaz:
## Description
This PR added the possibility to report a post.
Closes #50.

## Checklist
- [x] Wrote unit tests.


x/reports/client/cli/tx.go
```diff
@@ -0,0 +1,65 @@
...
+		GetCmdReportPost(cdc),
+	)...)
+
+	return postsTxCmd
+}
+
+// GetCmdReportPost is the CLI command for reporting a post
+func GetCmdReportPost(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "report [post-id] [reports-type] [reports-message]",
```
RiccardoM, line 41: I think that maybe changing this to `create` might be better than `report`, this would avoid the `reports report` repetition by making it 

```
desmoscli tx reports create ...
```

What do you think?

x/reports/internal/keeper/common_test.go
```diff
@@ -0,0 +1,71 @@
...
+func testCodecPosts() *codec.Codec {
+	var cdc = codec.New()
+
+	// register the different types
+	cdc.RegisterInterface((*crypto.PubKey)(nil), nil)
+	posts.RegisterCodec(cdc)
+
+	cdc.Seal()
+	return cdc
+}
```
RiccardoM, line 64: I think this can be removed safely as it shouldn't be used

x/reports/internal/keeper/keeper.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/desmos-labs/desmos/x/posts"
+	"github.com/desmos-labs/desmos/x/reports/internal/types"
+	"github.com/desmos-labs/desmos/x/reports/internal/types/models"
+)
+
+// Keeper maintains the link to data storage and exposes getter/setter methods for the various parts of the state machine
+type Keeper struct {
+	PostKeeper posts.Keeper // post's keeper to perform checks on the postIDs
```
RiccardoM, line 15: I've looked at the code and it seems to me that this isn't used anywhere. Am I right?

x/reports/internal/keeper/keeper.go
```diff
@@ -0,0 +1,100 @@
...
+// If the same reports has already been inserted, nothing will be changed.
+func (k Keeper) SaveReport(ctx sdk.Context, postID posts.PostID, report types.Report) bool {
+	store := ctx.KVStore(k.StoreKey)
+	key := models.ReportStoreKey(postID)
+	// Get the list of reports related to the given postID
+	var reports models.Reports
+	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &reports)
+
+	// try to append the given reports
+	reports, appended := reports.AppendIfMissing(report)
```
RiccardoM, line 40: i'm thinking that maybe this way of checking if the report already exists could lead to a problem if there are a lot of reports for the same post (which can easily happen if it contains nudity for example). 

Also, I don't see why we would want to block multiple reports to the same post (even from the same user). Instead, what I think we should do is allow a user to report as many times he wants the same post. It will be the dApp responsibility to handle that.

x/reports/internal/keeper/keeper.go
```diff
@@ -0,0 +1,100 @@
...
+	if appended {
+		store.Set(key, k.Cdc.MustMarshalBinaryBare(&reports))
+	}
+
+	return appended
+}
+
+// RegisterReportsTypes allows to save the given reportType inside the chain state.
+func (k Keeper) RegisterReportsTypes(ctx sdk.Context, reportType types.ReportType) bool {
```
RiccardoM, line 49: I really don't know how to feel about the pre-registered report types and the fact that someone might not be able to create a report if the type isn't registered. 

I'm thinking about the possibility that some dApps might want to use custom report types that are more specific, for example Uber when you report your driver or something else. Having pre-registered and a fixed amount of report types could lead to this not being possible. 

For this reason, I think it's better to get rid of the pore-registered report type, de-facto allowing dApps and users to specify any report type they want when creating them. This should allow for a more easy usage of the protocol overall.

x/reports/internal/simulation/genesis.go
```diff
@@ -0,0 +1,36 @@
...
+	reportsMap = make(map[string]types.Reports, reportsMapLen)
+	for i := 0; i < reportsMapLen; i++ {
+		reportsLen := simState.Rand.Intn(20)
+		reports := make(types.Reports, reportsLen)
+		for j := 0; j < reportsLen; j++ {
+			privKey := ed25519.GenPrivKey().PubKey()
+			reports[j] = types.NewReport(
+				RandomReportTypes(simState.Rand),
+				RandomReportMessage(simState.Rand),
+				sdk.AccAddress(privKey.Address()))
```
RiccardoM, line 30: I think it's better to format this

```suggestion
				sdk.AccAddress(privKey.Address()),
			)
```

x/reports/client/cli/query.go
```diff
@@ -0,0 +1,51 @@
...
+	postQueryCmd.AddCommand(flags.GetCommands(
+		GetCmdQueryPostReports(cdc),
+	)...)
+	return postQueryCmd
+}
+
+// GetCmdQueryPostReports queries a post's reports
+func GetCmdQueryPostReports(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "all [post-id]",
```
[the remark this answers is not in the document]
bragaz: yeah you're right, `all` could be confusing

x/reports/internal/keeper/keeper.go
bragaz: it is used inside the `handler` to check if there is any post related to the ID given, otherwise anyone could report non-existent posts.

x/reports/internal/keeper/keeper.go
bragaz: I was thinking that from the "spam-side" but I keep forgetting that they will spend token each report they'll made so... 🤣

x/reports/internal/keeper/keeper.go
bragaz: While I'm implementing the pre-registered type lists I was referring to your statement in the linked issue:
>To be consistent across different applications, we should define a list of supported Types (probably inside the genesis) so that custom applications can have a list of them.

I'm fine with removing the pre-registered list cause it would lead to less checks during the report's saving but doing this we will lose the consistency that you are talking about in the  issue. If that's not important we could remove it w/o problems.

RiccardoM: @bragaz Yeah you're right, that was my initial idea. However, on a comment later on I asked to leave out this point while implementing the feature (see [here](https://github.com/desmos-labs/desmos/issues/50#issuecomment-628397322)). 

I think it's better to remove this all together to have more freedom and less checks so please proceed with the removal of pre-registered types.

x/reports/internal/keeper/keeper.go
RiccardoM: @bragaz Can we encapsulate this operation inside a keeper method such as `CheckPostExistance` and make the `PostKeeper` private (renaming it `postsKeeper`)? This would hide this dependency that I think will result in a more easy to comprehend code

bragaz: we could have some problems with sim tests but I can try

docs/developers/msgs/report-post.md
```diff
@@ -0,0 +1,66 @@
...
+"unauthorized sale",  
+"hatred incitement",  
+"promotion of drug use",  
+"non-consensual intimate images",  
+"pornography",  
+"children abuse",  
+"animals abuse",  
+"bullying",  
+"scam" 
+```
```
RiccardoM, line 44: I think this can be deleted 

```suggestion
```

x/reports/genesis.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+
+// InitGenesis initializes the chain state based on the given GenesisState
+func InitGenesis(ctx sdk.Context, keeper Keeper, data GenesisState) []abci.ValidatorUpdate {
+	for postID, reports := range data.Reports {
+		postID := posts.PostID(postID)
+		if !postID.Valid() {
+			panic(fmt.Errorf("invalid postID: %s", postID))
+		}
```
RiccardoM, line 24: I think that it might be better here to use the `ParsePostID` function to make to compatible with future changes if there ever going to be: 

```suggestion
		postID, err := posts.ParsePostID(postID)
		if err != nil {
			panic(err)
		}
```

x/reports/internal/keeper/keeper.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// SaveReport allows to save the given reports inside the current context.
+// It assumes that the given reports has already been validated.
+// If the same reports has already been inserted, nothing will be changed.
+func (k Keeper) SaveReport(ctx sdk.Context, postID posts.PostID, report types.Report) {
+	store := ctx.KVStore(k.StoreKey)
+	key := models.ReportStoreKey(postID)
+	// Get the list of reports related to the given postID
+	var reports models.Reports
```
RiccardoM, line 43: Better adding a space here IMO

```suggestion
	key := models.ReportStoreKey(postID)
	
	// Get the list of reports related to the given postID
	var reports models.Reports
```

x/reports/internal/types/models/report.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return nil
+}
+
+// Equals checks if the two reports are the same or not
+func (r Report) Equals(other Report) bool {
+	return r.Type == other.Type &&
+		r.Message == other.Message &&
+		r.User.Equals(other.User)
+}
```
RiccardoM, line 59: I don't think we need this anymore so this can be deleted safely

```suggestion
```

x/reports/internal/types/models/report.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for index, rep := range reports {
+		if !rep.Equals(other[index]) {
+			return false
+		}
+	}
+
+	return true
+}
```
RiccardoM, line 96: If we don't need this anywhere we better delete it entirely

PR #180, opened 2020-05-25
Updated profiles' create and edit

### LINKED ISSUE #170: Change how the edit of the account is performed
## Context
I was working on implementing the accounts support inside Mooncake and I realized a bug inside the account module: we currently cannot delete a field. 

If we use [`MsgEditProfile`](https://github.com/desmos-labs/desmos/blob/master/x/profile/internal/types/msgs.go#L106) setting the fields we want delete to either `""` or `nil`, what would happen is that when the message gets handled by the handler ([here](https://github.com/desmos-labs/desmos/blob/master/x/profile/internal/keeper/handler.go#L68)) the `nil` values would be replaced with the existing ones. 

Originally the solution was to ask the user to explicitly tell which fields he wanted to delete by assigning them the `default` value (which would then be reserved), but I think that this makes the working for the clients a lot harder. 

I then realized two things: 

1. Clients 99.99% of the time will show what are the current values to the user before allowing him to edit them.

2. The _Create a profile_ and _Create an account_ terms are usually interchangeable in other social network platforms, but not on ours. We refer as an _account_ when talking about a Cosmos account, and to a _profile_ when we talk about a social network profile. This can be confusing to users (I already got someone asking me the difference). 

## Solution
I've thought that the best solution to both of the problems above is pretty simple: merge the current `MsgCreateProfile` and `MsgEditProfile` into a single `MsgSaveProfile`. 

What this new `MsgSaveProfile` would do is simply __overwrite all the data for the current profile__. That being said: 

- if the developer wants to preserve the current value of a field, he should pass that value inside the message
- if a developer wants to delete the value of a field, he should pass either `""` or `nil` to it. 

This should make things a lot easier for developers and also for users as there is not going to be a _create profile_ term anymore. 

### CLI changes
With this change in mind, the current `tx profile edit` CLI command should change accordingly. What I thought it might be a good implementation is the following: 

```
desmoscli tx profiles edit \
  --name <name> \
  --surname <surname> \ 
  --moniker <moniker> \
 -- ...
 --from jack
``` 

In short, what I envision for this new command is: 

- make all the editable fields __optional__ by putting them as flags
- put the flags default value to `""` 

Once the command is executed, the following things should happen: 

1. Get the currently existing account from the chain. 
2. Replace the value of each field which associated flag has given `""` with the current profile value 
3. Create and send the message

This will be exactly the same method that all clients will use in the future as well.

bragaz:
## Description
This PR merge the current `MsgCreateProfile` and `MsgEditProfile` into the new
`MsgSaveProfile` that handles both the situations.
Closes #170 .
## Checklist
- [x] Wrote unit tests.


x/profile/client/cli/tx.go
```diff
@@ -29,118 +29,56 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
-			newMoniker := GetFlagValueOrNilOnDefault(flagNewMoniker)
-			name := GetFlagValueOrNilOnDefault(flagName)
-			surname := GetFlagValueOrNilOnDefault(flagSurname)
-			bio := GetFlagValueOrNilOnDefault(flagBio)
+			picture := viper.GetString(flagProfilePic)
+			cover := viper.GetString(flagProfileCover)
+			moniker := viper.GetString(flagMoniker)
+			name := viper.GetString(flagName)
+			surname := viper.GetString(flagSurname)
+			bio := viper.GetString(flagBio)
```
RiccardoM, line 68: I'm not sure if using get `GetString` having an empty string works as we would want to. What I was thinking is the following. 

Suppose you execute `desmoscli tx save --name="Riccardo"` then what will happen is that all the remaining fields are rest to `""` instead of `nil` (I've checked and they are two different values). So I think we should change it appropriately. 

This could be done by creating the following functions: 

```golang
// getOrNil returns a pointer to the given value if it is not empty. Otherwise it returns nil
func getOrNil(value string) *string {
  if len(strings.TrimSpace(value)) == 0 {
    return nil
  }
  return &value
}
```

And use such function as 

```golang
moniker := getOrNil(viper.GetString(flagMoniker))
name := getOrNil(viper.GetString(flagMoniker))
```

x/profile/client/cli/tx.go
```diff
@@ -29,118 +29,56 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
...
 
-			msg := types.NewMsgEditProfile(newMoniker, name, surname, bio, picture, cover, cliCtx.FromAddress)
+			msg := types.NewMsgSaveProfile(moniker, &name, &surname, &bio, &picture, &cover, cliCtx.FromAddress)
```
RiccardoM, line 70: Instead of simply creating a `MsgSaveProfile` with the provided values, I think it might be better to: 

1. Query the existing profile. 
2. If the profile has some values, use them if the new ones are set to `nil`. 
3. Save the edited profile. 

Point 1 can be achieved using 

```golang
route := fmt.Sprintf("custom/%s/%s/%s", types.QuerierRoute, types.QueryProfile, cliCtx.FromAddress.String())
res, _, err := cliCtx.QueryWithData(route, nil)
if err != nil {
  fmt.Printf("Could not find a profile with moniker %s \n", args[0])
  return nil
}

var out types.Profile
cdc.MustUnmarshalJSON(res, &out)
```

For point 2, I suggest creating a set of methods called `EditXXX` that replaces the exiting values of they are not `nil`. For example: 

```golang
profile := profile.EditMoniker(newMoniker).EditName(newName).EditSurname(newSurname)
```

where

```golang
func (profile Profile) EditName(newName *string) Profile {
  if newName != nil && len(newName) > 0 {
    profile.Name = newName
  }
  return profile
}
```

bragaz: I think we could do this inside the `handler`, I'll edit it and commit asap

bragaz: While I'm editing the handler I realized that, with a method like `editName` a user will never be able do "delete" a field, if he want to do so.

RiccardoM: @bragaz You're right. Good point!

PR #184, opened 2020-05-27
Gov module restore

bragaz:
## Description
This PR restore `gov` module and its tests + fix the `DefaultGasValue` of sim tests
that was wrong
## Checklist
- [x] Targeted PR against correct branch.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Wrote unit tests.
- [x] Wrote integration tests (simulation & CLI). 
- [ ] Updated the documentation. 
- [x] Added an entry to the `CHANGELOG.md` file.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #185, opened 2020-06-01
Gov module restore and tests fixes

### LINKED ISSUE #173: Bring back the gov module
Taken from #142 

> Currently inside Desmos we were not making usage of the governance and upgrade modules. The first one was not used at all, while the second one was plugged in a couple of versions ago but never properly tested. 

> I've decided to remove them temporary inside `v0.4.0` to avoid having any problem inside our testnets, but we should definitely bring them back in the future.

With this issue I want to track the bringing back of the `gov` module.

bragaz:
## Description
This PR restore gov module and its tests + fix the tests that didn't work on the PR #184.
Closes #173. 
## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Version 0.7.0
 - Implemented benchmarks tests (#126)
 - Implemented posts' reports (#50)
+- Re-introduced the on-chain government module (#173)
 
 # Version 0.6.2
 ## Changes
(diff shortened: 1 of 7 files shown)


bragaz:
I notice that we're testing with 2mln (for report and profile modules) and 5mln (with posts) of default gas values. Strangely these values cause the failure of some sim tests and I cannot reproduce them locally because those tests passed without error on my machine. So I’ve opted to try with the regular value that we’re using on chain so (200K for report a profile) and (800k for posts which has some scenario where the gas is around 650k.

PR #188, opened 2020-06-03
Reaction registration fix

### LINKED ISSUE #187: Emoji reactions can be replaced by registered ones
## Bug description
Accepting a PR inside the Desmos Primer repository, I've stumbled across the following transaction: 

[`514AC64E99C6E709E001CC2250F1299FAEB29F7B38E5F62CB447476D6BBD3DA0`](https://morpheus.desmos.network/transactions/514AC64E99C6E709E001CC2250F1299FAEB29F7B38E5F62CB447476D6BBD3DA0)

As you can see from it, the user has registered a new reaction having shortcode `:smile:`, which is also the shortcode of the [:smile: emoji](https://emojipedia.org/grinning-face-with-smiling-eyes).

Since the implementation of #145, we shouldn't allow for this to happen. Instead, an error should be returned when trying to register the reaction with codes that also represent emojis.

## Steps to reproduce
1. Try to register a reaction using an emoji shortcode. 
2. No error is returned. 

## Expected behavior
The user should see an error telling him that the reaction cannot be registered due to the fact that already represents an emoji.

RiccardoM:
## Description
This PR fixed the reactions registration when they have the same shortcode of an existing emoji (closes #187). 

### Note
All the files not strictly related to the PR have been changed to make sure `make lint` doesn't return any error. 

## Checklist
- [x] Wrote unit tests.


RiccardoM:
@bragaz yes please take a loot at them

bragaz:
I think we shouldn't accept unicode value such as **U+1F600** anymore.   
They should be reserved only to emoji. We can achieve this by removing the 2nd check here inside the `reaction` `Validate()` function 

```golang
if !common.URIRegEx.MatchString(reaction.Value) && !IsEmoji(reaction.Value) {
		return fmt.Errorf("reaction value should be a URL or an emoji")
	}
```

PR #204, opened 2020-06-29
Move posts' const to chain params

### LINKED ISSUE #172: Move the current posts constants to chain parameters
## Context
Currently when checking the validity of the posts when the user creates or edited one, there are the current checks that are performed: 

- a post cannot be over `500` character long
- the maximum number of optional fields is `10`
- each optional field value can be at most `200` characters long

While all these values have currently worked out pretty good, having them hard coded creates two main problems: 

1. in order to know their value you need to check at the chain code; 
2. they are arbitrarily chosen by us and cannot be changed without a source code edit. 

## Solution 
In order to solve both above mentioned problems, I think we should move them from being hard-coded inside the `keys.go` file to be on-chain parameters. This would allow: 

1. everyone to query those specifications (particularly useful for clients); 
2. changeable through governance proposals. 

In order to do this we can do the same way that other modules do it. A reference can for example be the [`x/gov` module of the Cosmos SDK](https://github.com/cosmos/cosmos-sdk/tree/master/x/gov). The references files should be (might not be a complete list): 

- [`keeper/params.go`](https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/keeper/params.go)
- [`types/params.go`](https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/types/params.go)

Also the params registration should be done inside the `app.go` file as well ([`app.go` registration for the `gov` params](https://github.com/cosmos/cosmos-sdk/blob/master/simapp/app.go#L180)).

### LINKED ISSUE #205: Update CHANGELOG
With fd64bf9289c55f54cd2987ac3dc0a79580979443 you've renamed the `x/profile` module to `x/profiles`. This should be logged in the CHANGELOG accordingly

bragaz:
## Description
This PR move posts' const into chain params in order to let them be changed through proposal and queried easily from clients.
Close #172 .
Close #205 .

## Checklist
- [x] Wrote unit tests.


x/posts/internal/keeper/querier_test.go
```diff
@@ -488,3 +488,35 @@ func Test_queryRegisteredReactions(t *testing.T) {
...
 }
+
+func Test_queryParams(t *testing.T) {
+	tests := []struct {
+		name      string
+		path      []string
+		expResult types.Params
+	}{
+		{
+			name:      "Returning profile parameters correctly",
```
RiccardoM, line 499:
```suggestion
			name:      "Returning posts parameters correctly",
```

x/posts/internal/types/params.go
```diff
@@ -0,0 +1,138 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	paramsModule "github.com/cosmos/cosmos-sdk/x/params/subspace"
+)
+
+const (
+	// default paramspace for paramsModule keeper
+	DefaultParamspace = ModuleName
+)
+
+// Default profile paramsModule
```
RiccardoM, line 16:
```suggestion
// Default posts params
```

x/posts/internal/types/params.go
```diff
@@ -0,0 +1,138 @@
...
+// NewParams creates a new Params obj
+func NewParams(maxPostMLen, maxOpDataFieldNum, maxOpDataFieldValLen sdk.Int) Params {
+	return Params{
+		MaxPostMessageLength:            maxPostMLen,
+		MaxOptionalDataFieldsNumber:     maxOpDataFieldNum,
+		MaxOptionalDataFieldValueLength: maxOpDataFieldValLen,
+	}
+}
+
+// DefaultParams return default paramsModule
```
RiccardoM, line 50:
```suggestion
// DefaultParams return default params object
```

x/posts/legacy/v0.8.0/types.go
```diff
@@ -0,0 +1,22 @@
+package v060
```
RiccardoM, line 1: I think this should either be `v080` or this file should be moved

x/posts/legacy/v0.8.0/migrate_test.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	require.Len(t, migrated.RegisteredReactions, len(expected.RegisteredReactions))
+	for index, reaction := range migrated.RegisteredReactions {
+		require.Equal(t, expected.RegisteredReactions[index], reaction)
+	}
+
+}
+
+func TestMigrate060(t *testing.T) {
```
RiccardoM, line 166: I think this reference to `060` as well as all the ones below should be `080` instead.

x/posts/legacy/v0.8.0/migrate_test.go
```diff
@@ -0,0 +1,206 @@
...
+	for _, reaction := range v080state.RegisteredReactions {
+		// Make sure each reaction shortcode does not represent an emoji
+		_, err := emoji.LookupEmojiByCode(reaction.ShortCode)
+		require.Error(t, err)
+
+		// Make sure no reaction value is an emoji
+		_, err = emoji.LookupEmoji(reaction.Value)
+		require.Error(t, err)
+	}
+
```
RiccardoM, line 188: I think we should check if the params for the `posts` and `profiles` modules have been set correctly as well

x/posts/legacy/v0.8.0/migrate_test.go
```diff
@@ -0,0 +1,206 @@
...
+	require.Equal(t, len(v080state.PostReactions), len(v060state.PostReactions))
+	for index, postReactions := range v080state.PostReactions {
+		require.Contains(t, postReactions, v060state.PostReactions[index])
+	}
+
+	// Make sure the poll answers are all the same
+	require.Equal(t, len(v080state.UsersPollAnswers), len(v060state.UsersPollAnswers))
+	for index, answer := range v080state.UsersPollAnswers {
+		require.Equal(t, answer, v060state.UsersPollAnswers[index])
+	}
```
RiccardoM, line 205: I'm not sure these checks are correctly as the objects should not be equals to the one find inside the state, but they should be migrated. I would personally remove these checks.

x/posts/internal/keeper/keeper.go
```diff
@@ -3,24 +3,32 @@ package keeper
...
 // Keeper maintains the link to data storage and exposes getter/setter methods for the various parts of the state machine
 type Keeper struct {
+	// The reference to the ParamsStore to get and set posts specific params
+	paramSubspace params.Subspace
+
 	StoreKey sdk.StoreKey // Unexposed key to access store from sdk.Context
 	Cdc      *codec.Codec // The wire codec for binary encoding/decoding.
 }
 
 // NewKeeper creates new instances of the magpie Keeper
```
RiccardoM, line 22:
```suggestion
// NewKeeper creates new instances of the posts Keeper
```

x/posts/internal/keeper/handler.go
```diff
@@ -35,6 +35,35 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
+		if len(value) > int(maxOpFieldValLen) {
+			return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+				fmt.Sprintf("post optional data values cannot exceed %d characters. %s of post with id %s is longer than this",
+					maxOpFieldValLen, key, post.PostID))
+		}
+	}
+
+	return nil
+}
+
```
RiccardoM, line 66: Since this is also used inside the `invariants.go` file, I would move this to a standalone file to make it stand out more

x/posts/internal/keeper/handler.go
```diff
@@ -35,6 +35,35 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
 }
 
+// ValidatePost checks if the given post is valid according to the current posts' module params
+func ValidatePost(ctx sdk.Context, k Keeper, post types.Post) error {
+	params := k.GetParams(ctx)
+	maxMsgLen := params.MaxPostMessageLength.Int64()
+	maxOpFieldNum := params.MaxOptionalDataFieldsNumber.Int64()
+	maxOpFieldValLen := params.MaxOptionalDataFieldValueLength.Int64()
+
+	if len(post.Message) > int(maxMsgLen) {
```
RiccardoM, line 45: I think that for this line as well all the other lengths checks, it's better to convert the left side of the condition to `int64` instead of forcing the right one to `int`. This would avoid any exception if `maxMsgLen` is greater than the max `int` value allowed:


```golang
	if int64(len(post.Message)) > maxMsgLen {
```

x/posts/legacy/v0.8.0/migrate_test.go
bragaz: I was thinking that it's fine because (as we did with the previous versions) this is the test that migrate `v0.6.0` to the `v0.8.0`

RiccardoM: @bragaz Yeah the method is fine, it's just the names that are bugging me :joy: We are inside the `v0.8.0` folder and the method is called `TestMigrate060` instead as `TestMigrate080` :joy:

PR #216, opened 2020-07-02
Remove all the JSON-style String method

### LINKED ISSUE #199: Remove all the JSON-style String method
Currently in many places (particularly custom types such as `profile`, `post`, etc) when implementing the `String` method what we return is the JSON representation of such object.

I personally consider this an error, since the string representation should depend on the value of the `--output=` flag when executing CLI commands. 

For this reason, I suggest we entirely remove the custom `String` method implementations, leaving the serialization to the core implementation of the CLI output serializer.

bragaz:
## Description
This PR removed all the JSON-style `String()` methods around modules where it's possibile.
Some of them were used inside `decoder.go` and couldn't be deleted so this PR change them in order to not rely on JSON
for the string representation.
Closes #199 .
## Checklist
- [x] Wrote unit tests.


bragaz:
> @bragaz I've seen that `post`, `profile` and `post_reaction` kept their custom-made String method. I think we should remove it. Have you tried doing that?

Yes but they were used inside `decoder.go` inside `fmt.Sprintf()` functions. Deleting them we have an error there. I take a look at the cosmos SDK and they leave a custom method for this reason as well.

PR #217, opened 2020-07-02
Replace SetupTestInput with testify test suite

### LINKED ISSUE #198: Replace SetupTestInput methods with test suits
Looking at the [Cosmos SDK codebase](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/keeper_test.go#L60) it looks like they've swapped all the `SetupTestInput` methods for easier-to-understand [`testify.Suite`](https://github.com/stretchr/testify/blob/master/suite/suite.go#L22) implementations. 

I think we should adopt the same strategy, making the tests easier to setup and maintain.

bragaz:
## Description
This PR replace the `SetupTestInput` function with the `testify` `suite` in order to have a better handling
of all modules'  tests.
Closes #198.

## Checklist
- [x] Wrote unit tests.


x/posts/internal/keeper/common_test.go
```diff
@@ -55,35 +64,38 @@ func testCodec() *codec.Codec {
...
 }
 
-var testPostOwner, _ = sdk.AccAddressFromBech32("cosmos1y54exmx84cqtasvjnskf9f63djuuj68p7hqf47")
-var timeZone, _ = time.LoadLocation("UTC")
-var testPostCreationDate = time.Date(2020, 1, 1, 15, 15, 00, 000, timeZone)
-var testPostEndPollDate = time.Date(2050, 1, 1, 15, 15, 00, 000, timeZone)
-var testPostEndPollDateExpired = time.Date(2019, 1, 1, 1, 15, 00, 000, timeZone)
-var answer = types.PollAnswer{ID: types.AnswerID(1), Text: "Yes"}
+// common test
+var (
```
RiccardoM, line 68: Can we maybe define all these variables inside a custom made class such as `TestData` which would contain all the variables and later include that inside the suite itself? 

I was something something like: 

```golang
type TestData struct {
  user sdk.AccAddress
  time time.Time
  poll posts.PostPoll
  ...
}


type KeeperTestSuite struct {
  suite.Suite

  cdc          *codec.Codec
  ctx          sdk.Context
  keeper       keeper.Keeper
  paramsKeeper params.Keeper
  testData TestData
}
```

bragaz: yeah sure!

PR #243, opened 2020-07-21
Replacing Creation and Edit Dates with BlockTime

### LINKED ISSUE #215: Remove user specified post creation time and last edited in favor of block time usage
## Context
Currently when the user creates a post, it needs to specify by-hand the `CreationDate` and `LastEdited` values. This approach has some pros and cons to it. 

**PROs**
- It allows developers to specify a past creation date, so that they can post all their posts to Desmos easily
- It allows developers to create a post with a specific date, and then sync them to the chain easily in a second time, making sure the creation time is not changed

**CONs**
- It requires us to validate the creation and last edit times each time a new post is created or an existing one is edited, to make sure that the creation date is not equals or after the last edit one
- It requires us to validate the creation and last edit time to make sure they are not in the future based on the current block time

I was actually thinking about replacing those times with the value of `ctx.BlokTime()`: 

- if the post is being created, `post.CreationDate = ctx.BlockTime()`
- if the post is being edited, `post.LastEdit = ctx.BlockTime()`

This would solve the two above mentioned **cons**, but remote the above mentioned **pros**. 

What do you @kwunyeung @bragaz think about this? What option do you think is the best? Leaving it as it is, or replacing the values with the current block time?

bragaz:
## Description
This PR remove the `CreationDate` and `LastEdit` fields provided by users in order to use the `BlockTime` of Desmos chain.
Closes #215 .
## Checklist
- [x] Linked to Github issue with discussion and accepted design PR link to spec that describes this work.
- [x] Wrote unit tests.


bragaz:
I don't understand why sims are failing cause locally they all pass the tests.
Can you please check locally @RiccardoM ?

x/posts/keeper/handler.go
```diff
@@ -38,14 +38,15 @@ func NewHandler(keeper Keeper) sdk.Handler {
 
 // handleMsgCreatePost handles the creation of a new post
 func handleMsgCreatePost(ctx sdk.Context, keeper Keeper, msg types.MsgCreatePost) (*sdk.Result, error) {
+	postTime := ctx.BlockTime()
 	post := types.NewPost(
-		types.ComputeID(msg.CreationDate, msg.Creator, msg.Subspace),
+		types.ComputeID(postTime, msg.Creator, msg.Subspace),
```
RiccardoM, line 43: Before removing the `CreationDate` from the post, `ComputeID` made sure that the user could not create two different posts with the same creation date for the same subspace. 

Now that the creation time is the block time, however, this check should change. Suppose that a client app like Mooncake gathers all the posts to be sent to the chain every 30 seconds, and then uses a single transaction to broadcast them. If we would have to keep the `subspace + time + creator` unique key, each transaction would fail. Since all the posts are sent together, and the block is going to be only one, all of them would have the same creation date, the same subspace and the same creator. This will result in many very different posts having the same id. 

To solve this problem, I think that what we can do is change the `ComputeID` to be a method of the handler itself (since it's used only here) that accepts a `MsgCreatePost` and returns the hash of its JSON representation. This way we would make sure that for each block the message is unique. 

Initially I've thought about doing `sha256(time + creator + subspace + message)`. Then I thought that the message is optional, and also medias and the poll should be considered. For this reason, I think that `sha256(json(message))` is the best alternative to the current implementation to assure truly unique post ids.

x/posts/keeper/keeper.go
```diff
@@ -42,9 +42,26 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey, paramSpace params.Subspa
...
 	store := ctx.KVStore(k.StoreKey)
 
+	// Retrieve the total number of posts, if null it will be equal to 0
+	var numberOfPosts sdk.Int
+	if bz := store.Get(types.PostTotalNumberPrefix); bz != nil {
+		k.Cdc.MustUnmarshalBinaryBare(bz, &numberOfPosts)
+	} else {
+		numberOfPosts = sdk.NewInt(0)
+	}
+
```
RiccardoM, line 52: I think that this can be simplified as 

```suggestion
	// Retrieve the total number of posts, if null it will be equal to 0
	var numberOfPosts = sdk.ZeroInt
	if store.Has(types.PostTotalNumberPrefix) {
		k.Cdc.MustUnmarshalBinaryBare(store.Get(types.PostTotalNumberPrefix), &numberOfPosts)
	}
```

x/posts/keeper/keeper.go
```diff
@@ -42,9 +42,26 @@ func NewKeeper(cdc *codec.Codec, storeKey sdk.StoreKey, paramSpace params.Subspa
...
 	// Save the post
 	store.Set(types.PostStoreKey(post.PostID), k.Cdc.MustMarshalBinaryBare(&post))
 
+	// Check if the postID got an associated index, if so, don't increment the number of posts
+	if store.Get(types.PostIndexedIDStoreKey(post.PostID)) == nil {
+		numberOfPosts = numberOfPosts.Add(sdk.NewInt(1))
```
RiccardoM, line 58: Since `numberOfPosts` will be used only inside this `if`, I think it's better to initialize it here to avoid any unnecessary initialization otherwise.

x/posts/keeper/keeper.go
```diff
@@ -82,15 +99,24 @@ func (k Keeper) GetPostChildrenIDs(ctx sdk.Context, postID types.PostID) types.P
...
 	})
 
-	return posts
+	store := ctx.KVStore(k.StoreKey)
+	postsSorted := make(types.Posts, len(posts))
+	for _, post := range posts {
+		var index sdk.Int
+		k.Cdc.MustUnmarshalBinaryBare(store.Get(types.PostIndexedIDStoreKey(post.PostID)), &index)
+		postsSorted[index.Int64()-1] = post
+	}
```
RiccardoM, line 117: While this implementation surely works, I don't think that it's the best. One problem that I see with it is that other places that will use `IteratePosts` will have to copy/paste this piece of code in order to have them ordered as well. 

For this reason, I think that the best way to implement the iteration of posts based on their creation date is to edit the `IteratePosts` method directly. 

Assuming we have two keystore set of key/value pairs defined as: 


| Key | Value |
| :---| :--- |
| `1` | `PostID_a` |
| `2` | `PostID_b` |
| ... | ... |
| `n` | `PostID_z` |

| Key | Value |
| :---| :--- |
| `PostID_a` | Post `a` content |
| `PostID_b` | Post `b` content |
| ... | ... |
| `PostID_z` | Post `z` content |

It should behave as follows: 

1. Iterate over the ordering indexes (`1`, `2`, ... `n`) in *ascending order*. 
2. For each index, get the associated post id. 
3. With the post ID, get the post contents.

x/posts/keeper/handler.go
```diff
@@ -36,16 +38,27 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
+// ComputeID returns a sha256 hash of the msg's json representation
+// nolint: interfacer
+func ComputeID(msg types.MsgCreatePost) types.PostID {
+	bz, err := msg.MarshalJSON()
+	if err != nil {
+		panic(err)
+	}
+	hash := sha256.Sum256(bz)
+	return types.PostID(hex.EncodeToString(hash[:]))
+}
```
RiccardoM, line 50: While this implementation surely works, it has a major flaw in my opinion: clients will not be able to use it properly.

The advantage of having a deterministic id is that even client application will be able to generate them before event sending them to the chain. If we require them to know the concept and structure of a `MsgCreatePost`, this might make such pre-generation process harder for them. 

For this reason, I think that we should define a better way of generating the id. What I was thinking we might want to do is the following: 

1. Remove the `ID` field from the `NewPost` method. 
2. Inside the `NewPost` method body, create the proper ID hashing all the other data together (`ParentId`, `Message`, `AllowsComments`, `Subspace`, `OptionalData`, `CreationTime` and `Creator`). 

This way, the ID will be generated when creating the post object in a way that it's easily reproducible by dApps.

x/posts/keeper/keeper.go
bragaz: I think we could move the whole `numberOfPosts` check inside the `if` statement so that we check on the store only if a post has never been added before.

RiccardoM: @bragaz Yes, that would be great!

x/posts/keeper/handler.go
bragaz: I agree. With so many possible clients the last thing to do is surely complicate their lives.

bragaz: Actually, since the `NewPost` method has 85 reference around the code, I've preferred to *not* edit it but edit only the `ComputeID` method instead.

x/posts/keeper/keeper.go
bragaz: I tried to switch to this implementation but I'm having problems with the sim tests as the KV store (postIndex, postID) comes in two different orders  inside `ImportExport` test for the way that string sorting is implemented in GO.
Since it's not our interest to keep that store ordered, do you think there's something we could do to exclude it from sim tests? Or it should be ordered as well?  Also some comments are not ordered..I will recheck tomorrow, but I'm not sure if this is the right way to do this

RiccardoM: @bragaz Your current implementation is based on the following relationships (left side is the key, right side is the value): 

- `PostID -> Index` 
- `PostID -> Content` 

Then what you do is the following: 

1. Iterate over all the `PostID -> Content` entries and read thei contents. 
2. Iterate over all the `PostID -> Index` entries and re-order the posts. 

What I was suggesting is changing one of the two relationships. The ones that I thought were: 

- `Index -> PostID` 
- `PostID -> Content` 

This way, since the `Index` will be an `sdk.Int`, if we iterate over all the `Index` keys they will be iterated over in an ascending way (from `1` to `n`). 

So the implementation should be something like

```golang
func (k Keeper) IteratePosts(ctx sdk.Context, fn func(index int64, post types.Post) (stop bool)) {
  store := ctx.KVStore(k.StoreKey)
  
  iterator := sdk.KVStorePrefixIterator(store, types.PostIndexStorePrefix)
  defer iterator.Close()
  
  i := int64(0)
  for ; iterator.Valid(); iterator.Next() {
    // iterator.Value() will be the post ID
    // We need to: 
    // 1. Read the post id
    // 2. Based on the post ID, get the content
  }
}
```

x/posts/keeper/handler.go
RiccardoM: @bragaz Ok, that's a nice choice. I would still make so that `ComputeID` does not need a `MsgCreatePost` tho, but instead the whole set of data that should be used. This is to avoid the problem described above. 

Currently if we use the implementation you have written, a client library would hardly use it. Right now it should behave as follows: 

1. Create a JSON `MsgCreatePost`
2. Create a `Post` object

So, the client would have to know the format of both the `MsgCreatePost` and `Post` objects, which is not that good. Instead, what we should do is the following: 

1. Create a `Post` object. 
2. Hash all of its content in the JSON format, excluding the ID. 

So, for example: 

```javascript
const post = {
  "parent_id": "",
  "message": "Desmos is great!",
  "allows_comments": true,
  "subspace": "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
  "creator": "desmos1w3fe8zq5jrxd4nz49hllg75sw7m24qyc7tnaax",
  "creation_date": "2020-01-01T12:00:00Z"
}

const postID = sha256(jsonEncode(post))
```

This can probably be achieved using the following code: 

```golang
func NewPost(id, parentID PostID, message string, allowsComments bool, subspace string,
  optionalData map[string]string, created time.Time, creator sdk.AccAddress) Post {
  post := Post{ ... }
  
  postBz, err := json.Marshal(&post)
  if err != nil {
    panic(err)
  }
  
  postID := sha256.Sum256(postBz)
  post.PostID = PostID(hex.EncodeToString(hash[:]))
  return post
}
```

This would make sure that even client libraries can reproduce these steps without knowing the format of a `MsgCreatePost`. Also, thanks to it, we would be able to get a post id based on its content instead of knowing the original `MsgCreatePost` which is not available inside some places like the genesis.

x/posts/keeper/keeper.go
bragaz: sorry I forgot to push the last commit yesterday, you can now take a look at the implementation you've suggested

x/posts/keeper/handler.go
bragaz: We should move this to another PR @RiccardoM

RiccardoM:
@bragaz Tests are failing

app/sim_test.go
```diff
@@ -151,6 +151,9 @@ func TestAppImportExport(t *testing.T) {
 	appState, _, err := app.ExportAppStateAndValidators(false, []string{})
 	require.NoError(t, err)
 
+	file, _ := os.Create("genesis.json")
+	_, _ = file.WriteString(string(appState))
+
```
RiccardoM, line 156: I think we can remove these as they were put there just to debug

```suggestion
```

x/posts/keeper/common_functions.go
```diff
@@ -9,20 +9,42 @@ import (
...
 		stop := fn(i, post)
 		if stop {
 			break
 		}
 		i++
 	}
+
+	//freeing up memory
+	//nolint
+	postsSorted = nil
```
RiccardoM, line 47: Since `postsSorted` is created inside its body, when the function returns it will be cleared up anyway. I suggest we remove this then

```suggestion
```

PR #255, opened 2020-08-04
Replacing computeID with inner computation

### LINKED ISSUE #253: Replace ComputeID with inner ID computation
## Context
Currently when creating a new post, the ID is computed by hashing all of its property together, in order to come up with a deterministic id. As of today, this is done using the external `ComputeID` method. 

While this solution works, it actually is kind of bizarre. Since the ID is deterministic based on the post content, it can be created automatically without us even calling any method. 

## Solution
In order to fix this, I think that we can include the ID generation inside the `NewPost` method and delete the currently existing `ComputePostID` method. This would result in a much cleaner code that does not need to rely on external methods when creating a new post object, but only on the `NewPost` method itself.

bragaz:
## Description
This PR replace the `computeID` function in favour of a computation inside the `NewPost` function.
Closes #253.

## Checklist
- [x] Linked to Github issue with discussion and accepted design PR link to spec that describes this work.
- [x] Wrote unit tests.


x/posts/types/models/post.go
```diff
@@ -122,6 +124,15 @@ func NewPost(id, parentID PostID, message string, allowsComments bool, subspace
 		OptionalData:   optionalData,
 		Creator:        creator,
 	}
+
+	// PostID calculation
+	bz := []byte(parentID.String() + message + subspace + strconv.FormatBool(allowsComments) + created.String() +
+		creator.String())
+	hash := sha256.Sum256(bz)
+
+	post.PostID = PostID(hex.EncodeToString(hash[:]))
```
RiccardoM, line 133: I really like how cleanly this has been implemented :100: However, there is only one concern that comes to my mind: incompatibility with polls and attachments.

Let me explain this with an example. 

Let's suppose that Alice created three different posts and syncs them together, resulting in them being inserted in the same block. The posts are formed as follows 

1. First post  
   ```yml
   parent_id: "1"
   message: "Hello!"
   subspace: "mooncake"
   allows_comments: true
   ```

2. Second post
   ```yml
   parent_id: "1"
   message: "Hello!"
   subspace: "mooncake"
   allows_comments: true
   poll: 
     question: "Do you like cats?"
     options: 
        - "Yes"
        - "No"
   ```

3. Third post
   ```yml
   parent_id: "1"
   message: "Hello!"
   subspace: "mooncake"
   allows_comments: true
   attachments: 
     - url: "https://test.example.com/file" 
   ```

Now, all of them would have the same ID since this is computed **not** considering the poll nor the attachments info. 

In order to solve this problem, what I would suggest to do is the following. First, create a new private method: 

```go
// Computes a post ID based on the content of the given post.
func computePostId(post Post) string {
  // Hash together all the post data: 
  // ParentID, Message, Subspace, AllowsComments, Created, Creator
  // and also PollData and Attachments if not empty
}
```

Then, call this method in three different places: 

1. When the post is created the first time, inside the `NewPost` method. 
2. When some attachments are added to the post, inside `WithAttachments` 
3. When some poll data is added to the post, inside the `WithPollData` method

This way we would have a deterministic ID that is computed based on the whole post data, instead just part of it. Also, the post id would always be refreshed when some data is added/removed or changes.

bragaz: Perfect I didn't think about this before! I will provide to implment it asap

x/posts/types/models/post.go
```diff
@@ -106,13 +108,33 @@ type Post struct {
...
 	Attachments    Attachments    `json:"attachments,omitempty" yaml:"attachments,omitempty"`     // Contains all the attachments that are shared with the post
-	PollData       *PollData      `json:"poll_data,omitempty" yaml:"poll_data"`                   // Contains the poll details, if existing
+	PollData       *PollData      `json:"poll_data,omitempty" yaml:"poll_data,omitempty"`         // Contains the poll details, if existing
 }
 
-func NewPost(id, parentID PostID, message string, allowsComments bool, subspace string,
+// computeID computes a post ID based on the content of the given post.
+// nolint
+func computeID(parentID PostID, message, subspace string, allowsComments bool, creationTime time.Time, creator sdk.AccAddress,
+	attachments Attachments, pollData *PollData) PostID {
```
RiccardoM, line 117: What about passing a `Post` to `computeID` instead of all the separated data? I think this might be easier to maintain in the future if other data must be added. What do you think?

x/posts/types/models/post.go
```diff
@@ -106,13 +108,33 @@ type Post struct {
...
+	id := parentID.String() + message + subspace + strconv.FormatBool(allowsComments) + creationTime.String() +
```
RiccardoM, line 118: Instead of creating a string concatenating the different parts of a post, I think that it might be better to use the JSON serialization. This would also make it easier for extenal libraries to implement this method on their own and compute the IDs locally as well. 

Ideally, the same approach as when signing a `Message` should be followed: 

1. Serialize the entire `Post` as a JSON. 
2. Order the inner keys alphabetically in ascending order. 
3. Take the result and hash it using SHA-256. 

This would be ideal as it would always produce the same ID not matter the order of the serialized keys. What do you think about this?

x/posts/types/models/post.go
bragaz: I was thinking about that too, I'll try

x/posts/types/models/post.go
bragaz: Yup. This should be good enough for third parties.
Also the two first points are done at the same time inside `json.Marshal` so no big deal in using such approach. 👍

bragaz:
Coverage diff is only for the if statement inside `computeID` that checks for any problems during marshal

RiccardoM (approved):
Looks great! :100:

PR #260, opened 2020-08-17
Leonardo/users relationship impl

### LINKED ISSUE #168: Implement users relationship
## Feature description 
In all the social networks currently existing there is a concept that bonds two users between them in a tighter relationship than the rest of the users. For Facebook, this is called _"friendship"_, for Twitter it is called _"followage"_, for YouTube it is called _"subscription"_. Currently inside Desmos we do not have the way of expressing this relationship and we should definitely implement it. 

## Implementation proposal
One thing to not about all the above mentioned relationship is that they can be either mono- or bi-directional. Inside Twitter when you follow a user, you are the one observing his actions, and the same happens on YouTube. On Facebook, on the other hand, _"friendship"_ expresses a mutual connection, where two friends can observe the action of one another. 

### Types implementation
In order to properly implement both of these concepts, I think we could define a new object as follows: 

```golang
// Relationship represents a single relationship between two users. Sender is the one that first
// sent the relationship request, and Receiver is the one that received it and (optionally) accepted it.
type Relationship interface { 
  Sender()   sdk.AccAddress
  Receiver() sdk.AccAddress
}

// MonodirectionalRelationship implements Relationship and represents a monodirectional 
// relationships that does not require the receiver to accept it before being effective. 
type MonodirectionalRelationship struct {
  Sender   sdk.AccAddress   `json:"sender" yaml:"sender"`
  Receiver sdk.AccAddress   `json:"receiver" yaml:"receiver"`
}

// BidirectionalRelationship implements Relationship and represents a bidirectional relationship
// that can have different statuses and requires the receiver to accept it before becoming effective.
type BidirectionalRelationship struct {
  Sender   sdk.AccAddress     `json:"sender" yaml:"sender"`
  Receiver sdk.AccAddress     `json:"receiver" yaml:"receiver"`
  Status   RelationshipStatus `json:"status" yaml:"status"`
}

// RelationshipStatus represents the status of a bidirectional relationship
type RelationshipStatus int

const (
  RelationshipStatus Sent     = 0 // Indicates that the relationship has been sent but not yet accepted or denied
  RelationshipStatus Accepted = 1 // Tells that the relationships has been accepted by the receiver
  RelationshipStatus Denied   = 2 // Tells that the relationship has been denied from the receiver
)
```

### Messages implementation
The operations that two users should be able to perform are distinct based on the relationship type they want to create. 

#### Monodirectional relationships
The only operation that should be sufficient for monodirectional relationships are the creation of the relationship itself. For this the following message should be sufficient: 

```golang
// Creates a monodirectional relationship (type MonoDirectional = 0) between the sender and 
// the receiver. Monodirectional relationship do not need the receiver to accept in order to be valid.
// An example of monodirectional relationship is the follow on Twitter or the subscribe on YouTube.
type MsgCreateMonoDirectionalRelationship struct {
  Sender   sdk.AccAddress   `json:"sender" yaml:"sender"`
  Receiver sdk.AccAddress   `json:"receiver" yaml:"receiver"`
}
```

#### Bidirectional relationship
For bidirectional relationships, due to the fact that the receiver needs to accept, the following set of operations should be defined: 

1. Send the relationship request. 
2. Accept a relationship request or
3. Deny a relationship request. 

The following messages should allow for all of them: 

```golang
// MsgRequestBidirectionalRelationship allows the specified Sender to ask for a bidirectional
// relationship to the specified Receiver, attaching the given (optional) message to the request.
// In order to accept, the Receiver must be user either MsgAcceptBidirectionalRelationship or
// MsgDenyBidirectionalRelationship.
type MsgRequestBidirectionalRelationship struct {
  Sender   sdk.AccAddress   `json:"sender" yaml:"sender"`
  Receiver sdk.AccAddress   `json:"receiver" yaml:"receiver"`
  Message  string           `json:"message,omitempty" yaml: "message,omitempty"`
}

// MsgAcceptBidirectionalRelationship allows the receiver of a bidirectional relationship request
// to accept that request leading to the effective creation of the relationship itself. 
type MsgAcceptBidirectionalRelationship struct {
  Receiver sdk.AccAddress   `json:"receiver" yaml:"receiver"`
}

// MsgDenyBidirectionalRelationship allows the receiver of a bidirectional relationship request
// to deny that request. 
type MsgDenyBidirectionalRelationship struct {
    Receiver sdk.AccAddress   `json:"receiver" yaml:"receiver"`
}
```

Two things should be considered about the working of bidirectional relationships: 

1. Once accepted, a request **cannot** be denied anymore. However, it can still be cut (see below). 
2. Once denied, a request **can** still be accepted in the future. This will allow platforms similar to Facebook to implement the "Ignore" feature. 

#### Common operations
For both mono- and bi- directional relationship, once the relationship has been created the users taking part to it should be able to cut it off. The difference are: 

* In monodirectional relationships, the cut off can be done only by the `Sender` of the request. 
* In bidirectional relationships, the cut off can be done by either one of the two participants. 

In order to cut off a relationship the following message should be sufficient: 

```golang
// MsgDeleteRelationships allows the specified User to cut off the relationship he had prevously
// created with the specified Counterparty. 
// If the relationship was a monodirectional relationship, the user must be the original Sender of 
// the relationship, otherwise, if it was a bidirectional one, it can be either one of the two users 
// taking part to it.
type MsgDeleteRelationships struct {
  User         sdk.AccAddress `json:"user" yaml:"user"`
  Counterparty sdk.AccAddress `json:"counterparty" yaml:"counterparty"`
}
```

### Keeper implementation
In order to support all the above mentioned, I think that the following keeper method should be enough: 

```golang
// SaveRelationship allows to store the given relationship returning an error if something goes wrong.
func (k Keeper) SaveRelationship(ctx sdk.Context, relationship types.Relationship) error {}

// DeleteRelationship allows to delete the relationship between the given user and his counterparty
func (k Keeper) DeleteRelationship(ctx sdk.Context, user, counterparty sdk.AccAddress) error {}

// GetUserRelationships allows to list all the relationships that involve the given user.
func (k Keeper) GetUserRelationships(ctx sdk.Context, user sdk.AccAddress) []types.Relationship {}
```

### Conclusion
@kwunyeung @bragaz let me know what you guys think about this proposal.

bragaz:
## Description
This PR implements the users' relationships.
Close #168 . 

## Checklist
- [x] Wrote unit tests.


bragaz:
@RiccardoM I've made all the changes we discussed during the call.
I've not updated the docs yet, I'll wait for the approval of the code than I'll do it.

RiccardoM (commented):
I think you made some refactoring errors. Namely: 

- All `ID` references inside strings have been changed to `RelationshipID`
- All the `profilesTypes "github.com/desmos-labs/desmos/x/profiles/types"` imports have been changed to `"github.com/desmos-labs/desmos/x/profiles/types/models"`

Can you please revert that so that I have to view less files? Otherwise it would be complicated for me to see actual changes in the code

x/profiles/client/rest/query.go
```diff
@@ -15,6 +15,7 @@ func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/profiles/parameters", queryProfilesParamsHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/profiles/{address_or_dtag}", queryProfileHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/profiles", queryProfilesHandlerFn(cliCtx)).Methods("GET")
+	r.HandleFunc("/relationships/{address}", queryUserRelationships(cliCtx)).Methods("GET")
```
RiccardoM, line 18: What about adding a new endpoint to query all the relationship for all the profiles?

x/profiles/client/cli/tx.go
```diff
@@ -30,6 +31,8 @@ func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
 	profileTxCmd.AddCommand(flags.PostCommands(
 		GetCmdSaveProfile(cdc),
 		GetCmdDeleteProfile(cdc),
+		GetCmdCreateMonoDirectionalRelationship(cdc),
+		GetCmdDeleteUserRelationship(cdc),
```
RiccardoM, line 35: What about adding a new command to query all the relationship for all the profiles?

x/profiles/keeper/keeper_relationships.go
```diff
@@ -0,0 +1,76 @@
...
+	key := types.RelationshipsStoreKey(user)
+
+	var relationships []sdk.AccAddress
+	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &relationships)
+
+	return relationships
+}
+
+// GetUsersRelationshipsMap allows to returns the map of all the users and their associated storedRelationships
+func (k Keeper) GetUsersRelationshipsMap(ctx sdk.Context) map[string][]sdk.AccAddress {
```
RiccardoM, line 42: I think we can rename this to be simply `GetUsersRelationships`

x/profiles/keeper/keeper_relationships.go
```diff
@@ -0,0 +1,76 @@
...
+	store := ctx.KVStore(k.StoreKey)
+	key := types.RelationshipsStoreKey(user)
+	var relationships []sdk.AccAddress
+	k.Cdc.MustUnmarshalBinaryBare(store.Get(key), &relationships)
+
+	for _, addr := range relationships {
+		if addr.Equals(receiver) {
+			return fmt.Errorf("relationship already exists with %s", receiver)
+		}
+	}
```
RiccardoM, line 22: Instead of performing this check here, I think we can: 

1. Extract these lines in a new method, maybe something like `DoesRelationshipExist`
2. Call the new method from the handler
3. Inside this method, if the relationship already exists then simply not add it to the array instead of throwing an error

RiccardoM (commented):
The implementation looks good. I left some comments. Also, I would like to discuss the possibility of moving the whole relationship implementation inside their own module instead of leaving it inside the `x/profile` one. What do you think?

x/profiles/client/rest/query.go
bragaz: Yeah it could be good, Also ok for moving relationships on a new module

x/profiles/keeper/keeper_relationships.go
bragaz: I don't think extracting this is the better choice since we use this only here and, with a method like `DoesRelationshipExist`, we'll need to access two times to the store instead of one inside the `msg` `Handler`.

RiccardoM: @bragaz Ok, I see. Then let's keep it that way

x/relationships/client/cli/query.go
```diff
@@ -0,0 +1,74 @@
...
+			cdc.MustUnmarshalJSON(res, &out)
+			return cliCtx.PrintOutput(out)
+		},
+	}
+}
+
+// GetCmdQueryUserRelationships queries all the profiles' users' relationships
+func GetCmdQueryUserRelationships(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "user_relationships [address]",
```
RiccardoM, line 56: I think we can rename this to be simply `user [address]`

RiccardoM (commented):
Looks good to me. Can we rename all the `MonoDirectionalRelationship` referenced to be simply `Relationship` instead?

docs/developers/msgs/create-relationship.md
```diff
@@ -0,0 +1,38 @@
+# `MsgCreateRelationship`
+This message allows you to create a mono directional relationship with a specified user.
```
RiccardoM, line 2:
```suggestion
This message allows you to create a relationship between the signer and a specified user.
```

docs/developers/msgs/create-relationship.md
```diff
@@ -0,0 +1,38 @@
...
+Mono directional relationships are like the follow of today's social networks.
```
RiccardoM, line 3:
```suggestion
```

docs/developers/msgs/delete-relationship.md
```diff
@@ -0,0 +1,37 @@
+# `MsgDeleteRelationship`
+This message allows you to delete a mono directional relationship with a specified counterparty.
```
RiccardoM, line 2:
```suggestion
This message allows you to delete a relationship existing with a specified counterparty.
```

docs/developers/perform-transactions.md
```diff
@@ -16,10 +16,14 @@ Here is the list of currently available [messages](developer-faq.md#what-is-a-me
...
 * [`MsgRegisterReaction`](msgs/register-reaction.md): allows you to register a reaction.
 
-### Profile
+### Profiles
 * [`MsgSaveProfile`](msgs/save-profile.md): allows you to create or edit an existing profile.
 * [`MsgDeleteProfile`](msgs/delete-profile.md): allows you to delete an existing profile.
 * [`EditParamsProposal`](msgs/edit_param_proposal.md): allows you to open a proposal to change profile's params.
 
+## Relationships
+* [`MsgCreateMonoDirectionalRelationship`](msgs/create-relationship.md): allows you to create a mono directional relationship.
```
RiccardoM, line 25:
```suggestion
* [`MsgCreateRelationship`](msgs/create-relationship.md): allows you to create a relationship.
```

docs/developers/queries/user_relationships.md
```diff
@@ -0,0 +1,18 @@
+## Query user relationships
+This query endpoint allows you to retrieve the details of a relationship where the creator is the given `address`.
```
RiccardoM, line 2:
```suggestion
This query endpoint allows you to retrieve the details of a relationship where the creator has the given `address`.
```

x/relationships/client/cli/query.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/desmos-labs/desmos/x/relationships/types"
+	"github.com/spf13/cobra"
+)
+
+// GetQueryCmd adds the query commands
+func GetQueryCmd(cdc *codec.Codec) *cobra.Command {
+	profileQueryCmd := &cobra.Command{
```
RiccardoM, line 17: We should rename this `profileQueryCmd` with a more generic `cmd` I think, so that we don't have any problem moving it around if needed

x/relationships/client/cli/tx.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+	"github.com/spf13/cobra"
+)
+
+// GetTxCmd set the tx commands
+func GetTxCmd(_ string, cdc *codec.Codec) *cobra.Command {
+	profileTxCmd := &cobra.Command{
```
RiccardoM, line 22: We can rename `profileTxCmd` with just `cmd` for the sake of simplicity

x/relationships/client/cli/tx.go
```diff
@@ -0,0 +1,85 @@
...
+
+			return utils.GenerateOrBroadcastMsgs(cliCtx, txBldr, []sdk.Msg{msg})
+		},
+	}
+
+	return cmd
+}
+
+// GetCmdDeleteUserRelationship is the CLI command for deleting a relationship
+func GetCmdDeleteUserRelationship(cdc *codec.Codec) *cobra.Command {
```
RiccardoM, line 64: I think we can rename this `GetCmdDeleteRelationship` for simplicity

x/relationships/client/rest/tx.go
```diff
@@ -0,0 +1,91 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+	"github.com/desmos-labs/desmos/x/relationships/types"
+	"github.com/gorilla/mux"
+)
+
+func registerTxRoutes(cliCtx context.CLIContext, r *mux.Router) {
+	r.HandleFunc("/relationships/create/{address}", createRelationshipHandler(cliCtx)).Methods("POST")
+	r.HandleFunc("/relationships/delete/{address}", deleteRelationshipHandler(cliCtx)).Methods("DELETE")
```
RiccardoM, line 16: I think we can delete the `/create/` and `/delete/` parts of the query, and maybe the `/{address}` as well, requiring the user to put it into the body of the request instead. What do you think?

```suggestion
	r.HandleFunc("/relationships", createRelationshipHandler(cliCtx)).Methods("POST")
	r.HandleFunc("/relationships", deleteRelationshipHandler(cliCtx)).Methods("DELETE")
```

x/relationships/keeper/keeper_test.go
```diff
@@ -0,0 +1,120 @@
...
+				store := suite.ctx.KVStore(suite.keeper.StoreKey)
+				store.Set(types.RelationshipsStoreKey(test.user), suite.keeper.Cdc.MustMarshalBinaryBare(&test.storedRelationships))
+			}
+			err := suite.keeper.StoreRelationship(suite.ctx, test.user, test.receiver)
+			suite.Equal(test.expErr, err)
+		})
+	}
+}
+
+func (suite *KeeperTestSuite) TestKeeper_GetUserRelationshipsMap() {
```
RiccardoM, line 46: Can we add a test case where we try to read the relationships with an empty set?

x/relationships/keeper/keeper_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	_ = suite.keeper.StoreRelationship(suite.ctx, suite.testData.user, suite.testData.otherUser)
+	_ = suite.keeper.StoreRelationship(suite.ctx, suite.testData.otherUser, suite.testData.user)
+
+	actualIDsMap := suite.keeper.GetUsersRelationships(suite.ctx)
+	suite.Equal(relationshipsMap, actualIDsMap)
+}
+
+func (suite *KeeperTestSuite) TestKeeper_GetUserRelationships() {
```
RiccardoM, line 59: Can we add a test case where we try to read the relationships with an empty slice?

x/relationships/keeper/keeper_test.go
```diff
@@ -0,0 +1,120 @@
...
+			storedRelationships: []sdk.AccAddress{addr1, addr2, addr3},
+			expRelationships:    []sdk.AccAddress{addr1, addr3},
+			userToDelete:        addr2,
+		},
+		{
+			name:                "Delete a relationship with len(relationships) == 1",
+			storedRelationships: []sdk.AccAddress{addr1},
+			expRelationships:    nil,
+			userToDelete:        addr1,
+		},
```
RiccardoM, line 107: Can we add a test case where we try deleting a relationship with `len(relationships) == 0`?

RiccardoM (commented):
LGTM. Some small fixed and tests additions and it should be ready to be merged :100:

x/relationships/client/rest/tx.go
bragaz: I think there will not be much difference. The only concern I have is if it's possibile to have two endpoints with the exact same name

RiccardoM: @bragaz Yes, it is cause the REST method will be different. In the first one it will be `POST`, and the other one it will be `DELETE`. Removing the `/create` and `/delete` parts will make the endpoint follow the REST guidelines (no verbs inside endpoint names, only nouns). And removing the `{address}` variable will make it coherent with other requests as well

RiccardoM (approved):

PR #269, opened 2020-09-03
Changed relationships implementation

### LINKED ISSUE #266: Change how relationships are represented
## Context
With #168 we've introduced user relationships as a simple `sdk.AccAddress -> []sdk.AccAddress` map. This was done in order to allow applications to handle their relationship however they wanted on their level. 

With the current system the following is possible: 

1. A relationship between Alice and Bob is created on Mooncake. 
2. Another application comes in place and uses the same relationship. 
3. Mooncake decides to delete the relationship. 
4. The relationship is deleted from the other application as well. 

The blockchain should represent the data storage. Since Desmos is a protocol, applications should not see the data they use changed if they don't want to. 

## Solution
To fix this error, I think that we should revisit the relationship definition in order to include the subspace of the application for which the relationship is being created as well: 

```go
type Relationship struct { 
  Recipient sdk.AccAddress `json:"recipient" yaml:"recipient"`
  Subspace  string         `json:"subspace" yaml:"subspace"`
} 
```

Then, the storage can be a simple `sdk.AccAddress -> []Relationship`.

bragaz:
## Description
This PR change how the relationship were implemented. Now a relationship is represented by a `relationship` object containing an additional `subspace` field besides the `recipient`.
Closes #266.
## Checklist
- [x] Wrote unit tests.


docs/developers/msgs/create-relationship.md
```diff
@@ -7,7 +7,8 @@ This message allows you to create a relationship between the signer and a specif
   "type": "desmos/MsgCreateRelationship",
   "value": {
     "sender": "<Desmos address that's creating the relationship>",
-    "receiver": "<Desmos address that's receiving the relationship>"
+    "receiver": "<Desmos address that's receiving the relationship>",
+    "subspace": "<Subspace of a relationship>"
```
RiccardoM, line 11:
```suggestion
    "subspace": "<Subspace of the relationship>"
```

docs/developers/msgs/create-relationship.md
```diff
@@ -17,14 +18,16 @@ This message allows you to create a relationship between the signer and a specif
 | :-------: | :----: | :-------- |
 | `sender`  | String | Desmos address of the user that is creating the relationship |
 | `receiver`| String | Desmos address of the relationship's recipient |
+| `subspace`| String | Required string that identifies the app where the user make the relationship |
```
RiccardoM, line 21:
```suggestion
| `subspace`| String | Identifies the app for which the the relationship should be valid |
```

docs/developers/msgs/delete-relationship.md
```diff
@@ -7,7 +7,8 @@ This message allows you to delete an existing relationship with a specified coun
   "type": "desmos/MsgDeleteRelationship",
   "value": {
     "sender": "<Desmos address that's deleting the relationship>",
-    "counterparty": "<Desmos address that's with which sender want to cut-off the relationship>"
+    "counterparty": "<Desmos address that's with which sender want to cut-off the relationship>",
+    "subspace": "<Subspace of a relationship>"
```
RiccardoM, line 11:
```suggestion
    "subspace": "<Subspace of the relationship>"
```

docs/developers/msgs/delete-relationship.md
```diff
@@ -7,7 +7,8 @@ This message allows you to delete an existing relationship with a specified coun
   "type": "desmos/MsgDeleteRelationship",
   "value": {
     "sender": "<Desmos address that's deleting the relationship>",
-    "counterparty": "<Desmos address that's with which sender want to cut-off the relationship>"
+    "counterparty": "<Desmos address that's with which sender want to cut-off the relationship>",
```
RiccardoM, line 10:
```suggestion
    "counterparty": "<Desmos address with which the sender want to cut the relationship>",
```

docs/developers/msgs/delete-relationship.md
```diff
@@ -17,14 +18,16 @@ This message allows you to delete an existing relationship with a specified coun
 | :-------: | :----: | :-------- |
 | `sender`  | String | Desmos address of the user that is deleting the relationship |
 | `counterparty`| String | Desmos address of the relationship's counterparty |
+| `subspace`| String | Required string that identifies the app where the user make the relationship |
```
RiccardoM, line 21:
```suggestion
| `subspace`| String | String that identifies the app where for which the relationship was valid |
```

docs/types/relationships/relationship.md
```diff
@@ -0,0 +1,12 @@
+# Relationship
+Inside Desmos, users can create relationships with other users.
+A relationship inside desmos is the equivalent of the follow of traditional social networks.
```
RiccardoM, line 3:
```suggestion
A relationship inside desmos is the equivalent of the "follow" concept of traditional social networks.
```

docs/types/relationships/relationship.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Contained data
+Relationship contains data that allows identifying the dApp where users create the relationship.
```
RiccardoM, line 6:
```suggestion
Each relationship contains the data that allows identifying the dApp where the users created such relationship.
```

x/relationships/types/models/Relationship.go
```diff
@@ -0,0 +1,58 @@
+package models
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	posts "github.com/desmos-labs/desmos/x/posts/types"
+)
+
+type Relationship struct {
```
RiccardoM, line 10: Can we add some documentation to this type?

x/relationships/types/models/Relationship.go
```diff
@@ -0,0 +1,58 @@
+package models
```
RiccardoM, line 1: I think it's better to rename this file to `relationship.go` (all lower-case)

x/relationships/types/models/Relationship_test.go
```diff
@@ -0,0 +1,86 @@
+package models_test
```
RiccardoM, line 1: I think it's better to rename this file as `relationship_test.go` (all lower case)

x/relationships/client/cli/tx.go
```diff
@@ -51,7 +51,7 @@ func GetCmdCreateRelationship(cdc *codec.Codec) *cobra.Command {
 				return err
 			}
 
-			msg := types.NewMsgCreateRelationship(cliCtx.FromAddress, receiver)
+			msg := types.NewMsgCreateRelationship(cliCtx.FromAddress, receiver, args[1])
```
RiccardoM, line 54: Should we check the validity of the subspace directly here since we can? I think this might make the command fail earlier which is what we should always do. What do you think? :thinking:

x/relationships/client/cli/tx.go
```diff
@@ -76,7 +76,7 @@ func GetCmdDeleteRelationship(cdc *codec.Codec) *cobra.Command {
 				return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, fmt.Sprintf("invalid receiver address: %s", receiver))
 			}
 
-			msg := types.NewMsgDeleteRelationship(cliCtx.FromAddress, receiver)
+			msg := types.NewMsgDeleteRelationship(cliCtx.FromAddress, receiver, args[1])
```
RiccardoM, line 79: As above, should we add a subspace check here? :thinking:

x/relationships/keeper/keeper_test.go
```diff
@@ -9,24 +9,29 @@ import (
...
 			name:                "relationship added correctly",
 			storedRelationships: nil,
 			user:                suite.testData.user,
-			receiver:            suite.testData.otherUser,
-			expErr:              nil,
+			relationship: types.NewRelationship(suite.testData.otherUser,
+				"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e"),
+			expErr: nil,
 		},
 	}
```
RiccardoM, line 36: Can we add two tests to make sure that adding a new relationship with:

- a different subspace
- a different receiver

work properly if another relationship already exists?

x/relationships/client/cli/tx.go
bragaz: Yeah sure, we can leave the check both on message (since they are reached if someone use REST cmds) and inside the CLI

RiccardoM (approved):
LGTM :rocket:

PR #270, opened 2020-09-07
Allow users to change their DTag

### LINKED ISSUE #226: Allow users to change their DTag
## Feature description 
Currently after a user has registered a profile, the DTag value will be de-facto immutable, forcing him to build a whole new profile if he wants to change. 

Since the only constriction about DTags is that they should be unique, and they should be used when tagging a user as a human-readable references to the tagged person, I think we can safely allow any user to change his DTag at any time. 

## Implementation proposal
To implement this feature, I think it's sufficient to change the checks inside the `HandleMsgSaveProfile`, by **removing*** the condition that checks whether the provided `DTag` inside the message is equals to the exiting one.

Thanks to this change, and the fact that we already have in place a check to make sure that the provided DTag is not already associated to another user, the trick should be done. 

### Clients tagging consideration 
About tagging, I think that the clients should not rely on the `DTag` values when tagging users. Instead, they should always replace the `DTag` with the user `Address` value to make sure it never changes in the future. 

So, suppose the user writes the following post inside the client: 

> Hello @desmos! 

This should be translated from the client into: 

> Hello @[desmos1...]

Then, when rendered, the `[desmos1...]` part should be replaced with the user tag, allowing to click on it properly. 

What do you guys think @kwunyeung @bragaz?

bragaz:
## Description
This PR make a small edit inside `profiles` workflow to allow users to change their DTag.
Closes #226.

## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -192,7 +192,6 @@ jobs:
 
   integration-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v2
       - uses: technote-space/get-diff-action@v1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 ## Changes
 - Changed `relationships`' implementation adding a `subspace` field to identify in which app
   users make relationships (#266)
-- Allow users to change their DTag (#226)
 - Implemented the possibility to (un)blocking a specific user (#169)
+- Allow users to change their DTag (#226)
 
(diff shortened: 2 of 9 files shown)


RiccardoM (approved):
LGTM :rocket:

RiccardoM:
Looks like there is a test failing @bragaz. Otherwise looks good to be merged.

bragaz:
@RiccardoM Everything works fine. CLI tests keep stopping for an unknown error, but they all pass locally.
I think we can merge now

PR #294, opened 2020-10-10
Change OptionalData representation

### LINKED ISSUE #272: Change how OptionalData are represented inside posts
## Context 
Currently when working with Mooncake, we found ourselves struggling when using a post's optional data field. 

Particularly, trying to perform a transaction that create a post using that field results in an error being returned: 

```json
{
  "jsonrpc": "2.0",
  "id": -1,
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "unsupported type map"
  }
}
```

Talking with the Tendermint team, it seems like this is a problem with Go-Amino, the library used when (de)serializing JSON objects including messages. Particularly, it seems it **does not support maps**. 

## Solution 
In order to solve this problem, we should change how optional data are represented inside posts. A better and supported approach would be to use a list instead:

```golang
OptionalData   []OptionalData  `json:"optional_data,omitempty" yaml:"optional_data,omitempty"`
```

Where `OptionalData` is a new type defined as: 

```golang
type OptionalData struct {
  Key   string `json:"key`  
  Value string `json:"value"`
}
```

## Notes
The only thing to consider when adopting this approach is **how things should be handled during an automatic upgrade**. 
Right now I've asked the Cosmos SDK team [the following question](https://discordapp.com/channels/669268347736686612/680435043570941973/755355373002555454) and I'm waiting for their response. I'll update this issue once I get the answer.

bragaz:
## Description
This PR change how the `OptionalData` are represented inside a post moving from the previous `map[string]string{}` to a struct ```type OptionalData struct {
Key   string `json:"key" yaml:"key"`
	Value string `json:"value" yaml:"value"`
}``` .
Closes #272 .
## Checklist
- [x] Wrote unit tests.


docs/developers/msgs/posts/create-post.md
```diff
@@ -10,7 +10,7 @@ This message allows you to create a new public post. If you want to know more ab
     "message": "<Post message>",
     "allows_comments": false,
     "subspace": "<Subspace of a post>",
-    "optional_data": {},
+    "optional_data": "<Optional Data's array that contains all the optional data>",
```
RiccardoM, line 13:
```suggestion
    "optional_data": [],
```

docs/developers/msgs/posts/create-post.md
```diff
@@ -41,7 +41,8 @@ This message allows you to create a new public post. If you want to know more ab
     "allows_comments": true,
     "subspace": "4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
     "optional_data": {
-      "custom_field": "My custom value"
+      "key": "My custom key",
+      "value": "My custom value"
     },
```
RiccardoM, line 46: This should be an array, so it should become

```json
"optional_data": [
  {  
    "key": "My custom key",
    "value": "My custom value"
  }
]
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Version 0.13.0
 ## Changes
 - Removed the user_block event attribute key prefix. (#291)
+- Changed `OptionalData` representation from `map[string]string` to an `OptionalData` object (#272)
```
RiccardoM, line 4:
```suggestion
- Changed how posts' optional data are represented (#272) 
```

docs/types/posts/post.md
```diff
@@ -34,7 +34,7 @@ A common value that is used when you don't want to crete your own is `4e188d9c17
 If you instead prefer having a custom subspace, you can create your own by hashing any plain-text using any online SHA256 calculator such as [this one](https://emn178.github.io/online-tools/sha256.html).
 
 ### `OptionalData`
-In order to allow developers to specify any arbitrary data they want inside a post, we've introduced the `OptionalData` field. This field is a string-to-string map that allows to insert up to 10 fields containing any value you prefer. 
+In order to allow developers to specify any arbitrary data they want inside a post, we've introduced the `OptionalData` field. This field allows inserting up to 10 fields containing any value you prefer.
```
RiccardoM, line 37:
```suggestion
In order to allow developers to specify any arbitrary data they want inside a post, we've introduced the `OptionalData` field. This field is an array of objects containing a key and a value, and it allows inserting up to 10 fields containing any value you prefer. 
```

x/genutil/legacy/v0.13.0/migrate_test.go
```diff
@@ -0,0 +1,53 @@
...
+	v0130state := v0130.Migrate(v012state)
+
+	var v0130postsState v0130posts.GenesisState
+	err = cdc.UnmarshalJSON(v0130state[v060posts.ModuleName], &v0130postsState)
+	require.NoError(t, err)
+	require.Equal(t, len(v012postsState.Posts), len(v0130postsState.Posts))
+
+	// Make sure that all the posts' polls are migrated correctly
+	for _, post := range v0130postsState.Posts {
+		require.Equal(t, "[]v0130.OptionalData", reflect.TypeOf(post.OptionalData).String())
```
RiccardoM, line 51: I think the test here can be better. For example, we can test that: 

- if the original post's optional data are empty, the new post's optional data is empty
- if the original post's optional data are not empty, then the new post's optional data contain the proper objects having the key and value equals to the original post's optional data key and values

x/posts/legacy/v0.13.0/migrate.go
```diff
@@ -0,0 +1,56 @@
...
+			Attachments:    post.Attachments,
+			PollData:       post.PollData,
+		}
+	}
+	return posts
+}
+
+func ConvertOptionalData(oldOptionalData v040posts.OptionalData) []OptionalData {
+	optionalDataSize := len(oldOptionalData)
+	allOptionalData := make([]OptionalData, optionalDataSize)
```
RiccardoM, line 43: I think the first line can be merged into one. I also would suggest renaming `allOptionalData` to something like `optionalData` instead.

```suggestion
	optionalData := make([]OptionalData, len(oldOptionalData))
```

x/posts/legacy/v0.13.0/migrate.go
```diff
@@ -0,0 +1,56 @@
...
+	index := 0
+	for key, value := range oldOptionalData {
+		if index < optionalDataSize {
```
RiccardoM, line 46: Since we are iterating over the original map, this check will always be true. For this reason, I think we can remove this

x/posts/legacy/v0.13.0/types.go
```diff
@@ -0,0 +1,40 @@
...
+// GenesisState contains the data of a v0.12.0 genesis state for the posts module
+type GenesisState struct {
+	Posts               []Post                            `json:"posts"`
+	UsersPollAnswers    map[string][]v040posts.UserAnswer `json:"users_poll_answers"`
+	PostReactions       map[string][]v060.PostReaction    `json:"post_reactions"`
+	RegisteredReactions []v040posts.Reaction              `json:"registered_reactions"`
+	Params              v080posts.Params                  `json:"params"`
+}
+
+type OptionalData struct {
```
RiccardoM, line 23: I think it would be better to rename this an `OptionalDataEntry` instead, since it represents a single entry inside the optional data array

x/posts/types/models/common/post_optional_data.go
```diff
@@ -1,65 +1,30 @@
...
 
 // OptionalData represents a Posts' optional data and allows for custom
 // Amino and JSON serialization and deserialization.
-type OptionalData map[string]string
-
-// KeyValue is a simple key/value representation of one field of a OptionalData.
-type KeyValue struct {
-	Key   string
-	Value string
+type OptionalData struct {
```
RiccardoM, line 7: I think it would be better to rename this an `OptionalDataEntry` instead, since it represents a single entry inside the optional data array. 

Then, we can define the type `OptionalData` as 

```
type OptionalData []OptionalDataEntry
```

RiccardoM (commented):
The implementation overall looks good, I would suggest to change some small things to make it cleaner

x/posts/legacy/v0.13.0/migrate.go
bragaz: I've splitted the two lines because I need the size after inside the for loop

RiccardoM (approved):

PR #304, opened 2020-10-15
Multiple DTag bugs fixes

### LINKED ISSUE #295: Multiple DTag transfers between the same accounts are not allowed
## Bug description
Looking at transaction [E3CA26225BD97E2886FCF5C405A3C5091C43166BDE6AA9ED66C7A2F1775EE075](http://lcd.morpheus.desmos.network:1317/txs/E3CA26225BD97E2886FCF5C405A3C5091C43166BDE6AA9ED66C7A2F1775EE075), we can see that this failed with the following error:

```
the transfer request from desmos1rpurg200edtv6rgvfjuf7j6s8mghnsljkzvurg to desmos1k7q92w7ku72ycutz6xqrx67v92yycff3emz88s has already been made
```

As this says, it looks like that once a DTag transferring from A to B is concluded, no more transfers can be done. 

To be honest, I think this should not be the case. There might be a lot of cases in which A wants to transfer more than one DTag to B. 

For this reason, I think that once a DTag trade has been concluded, no trace of such trace should be kept and the same users should be allowed to trade again.

## Steps to reproduce
1. From Alice, request Bob's DTag. 
2. From Bob, accept Alice's request.
3. From Alice, request Bob's DTag again. 

## Expected behavior
The second request from Alice to Bob should not error and the trading process should start again.

### LINKED ISSUE #296: Bugs inside the DTag transferring process
## Bug description
During the completion of the Desmos Primer Phase 6 challenge that required them to perform a DTag transferring, the Discord user MikeSnowman75 told me they ran into a strange behavior that did not allow him to exchange the DTag properly. 

## Steps to reproduce
Reporting his words, here is what he did with an additional analyses from me. 

He tried to transfer the dtag first in these 2 transactions:
1. [First](http://lcd.morpheus.desmos.network:1317/txs/77D58E64007C98A8B6B0A1AF5D0B9E7E03CD140B7094BA96057C497E5E15EC6F)
2. [Second](http://lcd.morpheus.desmos.network:1317/txs/DD2A9293CC66807BFC190553C936E944C37BE910107D7AEC6792B3B81715E8D4)

As we can already see, there are two problems here: 

1. The first transaction has an empty `dtag_to_trade`.  
   This might say that the sending profile did not exist, or it didn't have the DTag set. This is a bug I've already reported inside #292. 

2. The second message fails with error `profile of desmos1k7q92w7ku72ycutz6xqrx67v92yycff3emz88s doesn't exist`.  
   This should not have happened. If a user wants a DTag of another user without having created a profile first, then a new profile for the receiving user should be created and the DTag transferring should continue properly.

After this, he tried with other two transactions: 

1. [First](http://lcd.morpheus.desmos.network:1317/txs/3B68A9CC4F72FE51C68F48A6E4F487EDAE1BA8C6D6AE150B071E1B7E9FE34584)

2. [Second](http://lcd.morpheus.desmos.network:1317/txs/6D819AB4F4B30E5D4C2F266065D0C1770AB5354315DC6F0A62E788B98372ED59)

From them, we can see that: 

1. The first transaction was successful. 
2. The second transaction failed with error `the owner's DTag is different from the one to be exchanged`.  
   Apparently, something here messed up because this one should not have failed here. 

Also, they said that they tried performing [another transaction](http://lcd.morpheus.desmos.network:1317/txs/E3CA26225BD97E2886FCF5C405A3C5091C43166BDE6AA9ED66C7A2F1775EE075). As you can see, this failed with error: 

```
the transfer request from desmos1rpurg200edtv6rgvfjuf7j6s8mghnsljkzvurg to desmos1k7q92w7ku72ycutz6xqrx67v92yycff3emz88s has already been made
```

To be honest, I don't think this error should ever verify. If an Alice -> Bob transfer was made successfully, then once it concluded it should be removed allowing Alice to send Bob another DTag if she wants. I've reported this bug inside #295.

## Expected behavior
The first transferring should have completed properly, as well as the second one.

bragaz:
## Description
This PR add two CLI test that checks the whole process of DTag transfer in the case that:
-  A user hasn't a profile
- A user has a profile

The two tests also make sure that after the first successful transfer:
- a new requests between the same profiles can be made
- previous requests of both sender and receiver were deleted

Closes #295 . 
Closes #296 .
## Checklist
- [x] Wrote unit tests.


x/posts/legacy/v0.13.0/migrate_test.go
```diff
@@ -228,22 +228,19 @@ func TestConvertOptionalData(t *testing.T) {
...
+
+	for key, value := range oldOptionalData {
+		for index, entry := range actualOptionalData {
+			if result := entry.Key == key && entry.Value == value; result {
+				found[index] = true
+			}
+		}
+	}
+
+	require.Equal(t, []bool{true, true, true}, found)
```
RiccardoM, line 245: I think there is an easier way of doing this, without having to use an additional `found` array: 

```suggestion
	for key, value := range oldOptionalData {
	        found := false
		for index, entry := range actualOptionalData {
			found = entry.Key == key && entry.Value == value
			if found {
			    break
			}
		}
		require.True(t, found)
	}
```

RiccardoM (approved):

PR #309, opened 2020-10-21
User block integration

### LINKED ISSUE #298: Implement proper blocking usage
## Feature description 
Currently, when a user blocks another user that blockage is never considered anywhere. Instead, we should consider it when: 

1. a blocked user asks for the DTag of the blocking user;
2. a blocked user adds a relationship towards the blocking user.

## Implementation proposal
What I think we should do is the following: 

1. when a DTag transferring request is made from Alice to Bob, check if Bob has blocked Alice. If so, return an error and halt the process;

2. when a relationships is created from Alice towards Bob, check if Bob has blocked Alice. If so, return an error and halt the process.

bragaz:
## Description
This PR integrate the user blocks into the DTag transfering process and relationship creation process.
Closes #298 . 
## Checklist
- [x] Wrote unit tests.


RiccardoM (commented):
I have few suggestions for the current implementation. 

First of all, there's no need to duplicate the following code in both handlers: 

```go
// CheckForBlockedUser checks if the given user address is present inside the blocked users array
func CheckForBlockedUser(blockedUsers []relationships.UserBlock, addr sdk.Address) bool {
  for _, user := range blockedUsers {
    if user.Blocked.Equals(addr) {
      return true
    }
  }
  return false
}
```

Instead, what can be done is:

**1.** Create a new method inside the relationships keeper (preferably in a different file such as `keeper_utils.go`) similar to:
```golang
// IsUserBlocked tells if the given blocker has blocked the given blocked user
func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked sdk.Address) bool {
  blockedUsers := k.GetUserBlocks(ctx, blocker)
  for _, user := range blockedUsers {
    if user.Blocked.Equals(addr) {
      return true
    }
  }
  return false
}
``` 

**2.** Keep the `relationships` keeper **private** inside the `profiles` keeper:
```go
type Keeper struct {
  ...
  relKeeper relationships.Keeper // Relationships keeper to keep track of blocked users
  ...
}
```

**3.** Create a utility method inside the profile keeper that simply calls the relationship keeper:
```go
// IsUserBlocked tells if the given blocker has blocked the given blocked user
func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked sdk.Address) bool {
  return k.relKeeper.IsUserBlocked(ctx, blocker, blocked)
}
```


This would make it a lot easier inside the handlers using the checks:

```go
if keeper.IsUserBlocked(ctx, receiver.Address, sender.Address) {
  return sim.Account{}, types.DTagTransferRequest{}, true
}
```

**Note**. There's no need to declare an `if`-variable if the only value returned if a boolean.

app/app.go
```diff
@@ -176,8 +176,8 @@ func NewDesmosApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest
 		gov.StoreKey, upgrade.StoreKey, params.StoreKey, evidence.StoreKey,
 
 		// Custom modules
-		magpieTypes.StoreKey, postsTypes.StoreKey, profilesTypes.StoreKey, reportsTypes.StoreKey,
-		relationshipsTypes.StoreKey,
+		magpieTypes.StoreKey, postsTypes.StoreKey, relationshipsTypes.StoreKey,
+		profilesTypes.StoreKey, reportsTypes.StoreKey,
```
RiccardoM, line 180: There's no need here to change the order of store keys. They are actually initialized based on the order that the `AppModuleBasic` are placed in

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Version 0.14.0
+- Implemented users blocks checks inside DTag transferring process and relationships creation process (#298)
```
RiccardoM, line 2: I would add this to the Changes subsection

```suggestion
### Changes
- Implemented blocked users checks (#298)
```

RiccardoM (commented):
I would also add the blocks checks inside the post creation. Users can tag another people inside the medias, so we should check that tagging someone who has blocked you does not work

bragaz:
I've also fixed a bug in `queryRelationships` tests due to the non-determinsm of maps

x/posts/simulation/operations.go
```diff
@@ -22,7 +22,7 @@ const (
 	OpWeightMsgAnswerPoll       = "op_weight_msg_answer_poll"
 	OpWeightMsgRegisterReaction = "op_weight_msg_register_reaction"
 
-	DefaultGasValue = 800000
+	DefaultGasValue = 1800000
```
RiccardoM, line 25: Why was this incremented?

x/posts/keeper/handler.go
```diff
@@ -115,6 +126,16 @@ func handleMsgEditPost(ctx sdk.Context, keeper Keeper, msg types.MsgEditPost) (*
 	existing.Message = msg.Message
 
 	if msg.Attachments != nil {
+		// Check if any of the tags have blocked the post creator
+		for _, attachment := range msg.Attachments {
+			for _, tag := range attachment.Tags {
+				// check if the request's receiver has blocked the sender before
+				if keeper.IsUserBlocked(ctx, tag, existing.Creator) {
+					return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+						fmt.Sprintf("The user with address %s has been blocked from %s", existing.Creator, tag))
```
RiccardoM, line 135: I think that it might be better return a more specific error here. Since the blocked user is the one creating the post, we can simply use

```suggestion
						fmt.Sprintf("The user with address %s has blocked you", tag))
```

x/posts/keeper/handler.go
```diff
@@ -115,6 +126,16 @@ func handleMsgEditPost(ctx sdk.Context, keeper Keeper, msg types.MsgEditPost) (*
...
+				}
+			}
+		}
```
RiccardoM, line 138: I would place this piece of code inside an external method since it's also used inside the `MsgCreatePost` handler

x/profiles/keeper/handler.go
```diff
@@ -156,6 +156,12 @@ func handleMsgDeleteProfile(ctx sdk.Context, keeper Keeper, msg types.MsgDeleteP
 
 // handleMsgRequestDTagTransfer handles the request of a dTag transfer
 func handleMsgRequestDTagTransfer(ctx sdk.Context, keeper Keeper, msg types.MsgRequestDTagTransfer) (*sdk.Result, error) {
+	// check if the request's receiver has blocked the sender before
+	if keeper.IsUserBlocked(ctx, msg.Receiver, msg.Sender) {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+			fmt.Sprintf("The user with address %s has been blocked from %s", msg.Sender, msg.Receiver))
```
RiccardoM, line 162: I think that since it's the same use that's creating the reaction and that has been blocked, we can return a more specific error:

```suggestion
			fmt.Sprintf("The user with address %s has blocked you", msg.Sender))
```

x/relationships/keeper/handler.go
```diff
@@ -31,6 +31,12 @@ func NewHandler(keeper Keeper) sdk.Handler {
 
 // handleMsgCreateRelationship handles the creation of a relationship
 func handleMsgCreateRelationship(ctx sdk.Context, keeper Keeper, msg types.MsgCreateRelationship) (*sdk.Result, error) {
+	// Check if the receiver has blocked the sender before
+	if keeper.IsUserBlocked(ctx, msg.Receiver, msg.Sender) {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+			fmt.Sprintf("The user with address %s has been blocked from %s", msg.Sender, msg.Receiver))
```
RiccardoM, line 37: Since the user that's creating the relationship is the same that is blocked, we can use a more specific error message here:

```suggestion
			fmt.Sprintf("The user with address %s blocked you", msg.Receiver))
```

x/posts/simulation/operations.go
bragaz: new checks require more gas in some cases

x/posts/keeper/handler.go
```diff
@@ -37,6 +37,20 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
+	for _, attachment := range attachments {
+		for _, tag := range attachment.Tags {
+			// check if the request's receiver has blocked the sender before
+			if keeper.IsUserBlocked(ctx, tag, creator) {
+				return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest,
+					fmt.Sprintf("The user with address %s has blocked you", tag))
+			}
+		}
+	}
+	return nil
```
RiccardoM, line 51: By taking a closer look at this code, I think that it could be further improved. 

Right now, what happens when this call is done is the following:

1. For each attachment, all tags are extracted. 
2. For each tag, we check if that tag is blocked. 

The thing is that to do (2) we: 

1. Extract all the users that have been blocked by the post creator
2. We iterate over that set 

This actually causes this method to access the store `m` times where `m` is the number of tags, causing the used gas to increment. 

Instead, what we could do is define a method such as 

```go
func (k Keeper) IsAnyUserBlocked(ctx sdk.Context, sdk.AccAddress blocker, ...sdk.AccAddress users) (bool, sdk.AccAddress) {
  blockedUsers := k.GetBlockedUsers(ctx, blocker)
  for _, user := range users {
    if blockedUsers.Contains(user) {
      return true, user
    }
  }
  return false, nil
}
```

This way we would have only one access to the store.  
(Note: The `Contains` method might be implemented inside the `common.go` file where other utils method lie).

This could then be used as follows:

```go
var tags []sdk.AccAddress
for _, attachment := range attachments {
  tags = append(tags, attachment.Tags...)
}

if blocked, blocker := keeper.IsAnyBlocked(ctx, creator, tags); blocked {
  return sdkerrors(sdkerrors.ErrInvalidRequest, 
      fmt.Sprintf("The user with address %s has blocked you", blocker))
}

return nil
```

bragaz: Unfortunately here we don't need to check which of the tags the post's creator blocked but from which of them he's blocked. So those checks are inevitable for the moment. 
We could open another issue with the aim to reduce the gas costs of create posts txs.

RiccardoM: You're right. I thought we had to check if the user blocked any of the tags, but actually it's the other way around.

x/posts/simulation/operations.go
RiccardoM: Can we try making this 1.000.000 instead of 1.800.000? This would result in an increment of `200.000` instead of `1.000.000` which I hope it's sufficient

RiccardoM (approved):

PR #314, opened 2020-10-26
Standard minimum fee implementation

### LINKED ISSUE #230: Standard or minimum transaction fee
## Feature description 

Be default, `Cosmos-SDK` let full nodes to define their `min-gas-prices`. This would lead to `Desmos` have a chance of facing zero-fee spam attack. I think we have to setup a minimum or standard fee structure for different `Posts` and `Reports` message types. We should not let users be able to post, make reactions or report without any fee. We can keep the way how other `StdTx` is working but only keeps transactions in `Posts` and `Reports` module have such constraints.

**IRISnet** has a standard minimum fee and **e-Money** has a PoA configurable minimum fee. I think we can learn how they are doing. While we are still studying the tokenomics and usage of a secondary fee token, we can first test this mechanism with the testing `udaric`. There is a module based `AnteDecorator` in the SDK for customizing the `AnteHandler`. Can this be achieved by making use of it?

bragaz:
## Description
This PR introduces a custom `AnteHandler` that checks minimum standard fee at `0.01 daric`.
Closes #230.  
## Checklist
- [x] Wrote unit tests.


RiccardoM (commented):
The implementation seems good, but I think there is one key point that could be improved. 

Right now the `FeeDenom` and `fixedRequiredFee` values are all hard coded. This makes it nearly impossible to change them without creating a new version of the application and then performing a hard fork of the chain. 

Instead, what I think could be done is create a new `x/fee` module inside which we can create some `Params` holding the data we want; eg:

```golang
type Params struct { 
  MinFee sdk.Coin `json:"min_fee",yaml:"min_fee"`
}
```

Then, inside the newly created `AnteDecorator` what we could do is the following: 

```golang
minFee := feekeper.GetMinFee(ctx)
if !stableRequiredQty.IsZero() && stableRequiredQty.GT(minFee) {
  ...
}
```

By doing so we could easily change the minimum fee amount or denom by simply using governance proposals, which is very nice to have.

RiccardoM:
While taking a look at the current implementation, I came up with some improvements that could be applied in order to make this module more generic. 

### File organization
I would move the `ante.go` and `ante_test.go` files from `x/ante` to `x/fees`. This would make it easier to comprehend that it's tied to that module instead of being part of another module.

### Improving the fee parameters
Currently, the fee parameter is formed as follows: 
```go
type Params struct { 
  MinFee sdk.Coin `json:"min_fee",yaml:"min_fee"`
}
```

While this might work in some cases, it surely does not for everyone. What if we want to have a minimum fee only for **some specific messages**? How can we achieve this? 

In order to make the module more generic and usable by other projects as well, I think what we can do is change the parameters to be as follows: 

```go
// MinFee contains the minimum amount of coins that should be paid as a fee for each message of the specified type sent
type MinFee struct {
  MessageType string   `json:"message", yaml:"message"`
  Amount      sdk.Coin `json:"amount", yaml:"amount"`
}

type Params struct { 
  MinFees []MinFee `json:"min_fees", yaml:"min_fees"`
}
```

By doing so, we could then have the following parameters:
```json
{
  "min_fees": [
    {
      "message": "desmos/MsgCreatePost", 
      "amount": {
        "amount": "10000",
        "denom": "udaric"
      }
    },
    {
      "message": "cosmos-sdk/MsgCreateValidator",
      "amount": {
        "amount": "2000",
        "denom": "stake"
      }
    }
  ]
}
```

### Improving the minimum fee check
currently, the minimum fee check is all performed inside the `checkMinimumFees` method of the ante handler. Instead, what I think should be done, is move this check inside the `x/fee` `Keeper`: 

```go
// CheckFees checks whether the given fees are sufficient to pay for all the given messages. 
// The check is performed considering the minimum fee amounts specified inside the module parameters.
func (k Keeper) CheckFees(ctx sdk.Context, fees sdk.Coins, msgs []sdk.Msg) error
```

This method should: 

1. Fetch the current parameters. 
2. For each message type included inside the parameters, compute how many fees should be paid
3. Check whether the given fees are enough to pay for all the ones that should be paid

This would simplify a lot the working of the `AnteHandle` that could then result in
```go
// ...

// Check the minimum fees
if err := mfd.feesKeeper.CheckFees(ctx, stdTx.Fee.Amount, stdTx.Msgs); err != nil {
	return ctx, err
}

// ...
```

bragaz:
@RiccardoM Finally fixed everything! 🎉
You can take a look now

.github/workflows/sims.yml
```diff
@@ -62,7 +62,7 @@ jobs:
 
   Test-Sim-Import-Export:
     runs-on: ubuntu-latest
-    needs: [build, install-runsim]
+    needs: [ build, install-runsim ]
```
RiccardoM, line 65: Revert
```suggestion
    needs: [build, install-runsim]
```

.github/workflows/sims.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
   Test-Sim-After-Import:
     runs-on: ubuntu-latest
-    needs: [build, install-runsim]
+    needs: [ build, install-runsim ]
```
RiccardoM, line 88: Revert
```suggestion
    needs: [build, install-runsim]
```

.github/workflows/sims.yml
```diff
@@ -108,7 +108,7 @@ jobs:
 
   Test-Sim-Multi-Seed-Short:
     runs-on: ubuntu-latest
-    needs: [build, install-runsim]
+    needs: [ build, install-runsim ]
```
RiccardoM, line 111: Revert
```suggestion
    needs: [build, install-runsim]
```

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Version 0.14.0
+## Changes
+- Implemented blocked users checks (#298)
+- Implemented a minimum transactions' fee (#230)
```
RiccardoM, line 4:
```suggestion
- Implemented the ability to set minimum messages fee (#230) 
```

docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     networks:
       localnet:
         ipv4_address: 192.168.10.2
-    command: ["desmosd", "start"]
+    command: [ "desmosd", "start" ]
```
RiccardoM, line 18: Revert
```suggestion
    command: ["desmosd", "start"]
```

docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
     networks:
       localnet:
         ipv4_address: 192.168.10.3
-    command: ["desmosd", "start"]
+    command: [ "desmosd", "start" ]
```
RiccardoM, line 34: Revert
```suggestion
    command: ["desmosd", "start"]
```

docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
     networks:
       localnet:
         ipv4_address: 192.168.10.4
-    command: ["desmosd", "start"]
+    command: [ "desmosd", "start" ]
```
RiccardoM, line 50: Revert
```suggestion
    command: ["desmosd", "start"]
```

docker-compose.yml
```diff
@@ -63,7 +63,7 @@ services:
     networks:
       localnet:
         ipv4_address: 192.168.10.5
-    command: ["desmosd", "start"]
+    command: [ "desmosd", "start" ]
```
RiccardoM, line 66: Revert
```suggestion
    command: ["desmosd", "start"]
```

x/fees/ante/ante.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (mfd MinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	// all transactions must be of type auth.StdTx
+	stdTx, ok := tx.(types.StdTx)
+	if !ok {
+		// Set a gas meter with limit 0 as to prevent an infinite gas meter attack
+		// during runTx.
+		newCtx = setGasMeter(simulate, ctx, 0)
+		return newCtx, errors.New("tx must be StdTx")
```
RiccardoM, line 55: Is this something we need to do here, or is this a check that might have already happened previously inside the ante chain?

x/fees/client/cli/query.go
```diff
@@ -0,0 +1,50 @@
...
+	postQueryCmd.AddCommand(flags.GetCommands(
+		GetCmdQueryFeesParams(cdc),
+	)...)
+	return postQueryCmd
+}
+
+// GetCmdQueryFeesParams queries all the fees' module params
+func GetCmdQueryFeesParams(cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "parameters",
```
RiccardoM, line 32: Probably it's better to name it as all the other modules
```suggestion
		Use:   "params",
```

x/fees/types/codec.go
```diff
@@ -0,0 +1,12 @@
+package types
```
RiccardoM, line 1: Nothing is done here. I think we can delete this file

x/fees/types/min_fee.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// MinFee contains the minimum amount of coins that should be paid as a fee for each message of the specified type sent
+type MinFee struct {
+	MessageType string  `json:"message_type" yaml:"message_type"`
+	Amount      sdk.Dec `json:"amount" yaml:"amount"`
```
RiccardoM, line 12: Instead of using `sdk.Dec` here, I think the best thing to do is to use `sdk.Coins` instead. This way, we can even set different fee coins based on a message type: 

```json
[
  {
    "message_type": "create_validator",
    "amount": {
      "amount": "200",
      "denom": "udaric"
    }
  },
  {
    "message_type": "edit_validator",
    "amount": {
      "amount": "100",
      "denom": "upotic"
    }
  }
]
```

x/fees/keeper/keeper.go
```diff
@@ -0,0 +1,69 @@
...
+	return p
+}
+
+// CheckFees checks whether the given fees are sufficient to pay for all the given messages.
+// The check is performed considering the minimum fee amounts specified inside the module parameters.
+func (k Keeper) CheckFees(ctx sdk.Context, fees authtypes.StdFee, msgs []sdk.Msg) error {
+	feesParams := k.GetParams(ctx)
+
+	// calculate required fees for the given messages
+	requiredFees := sdk.NewDec(0)
```
RiccardoM, line 48: If the `minFee.Amount` is changed to use `sdk.Coins`, then this can be changed to be `sdk.Coins` as well

x/fees/keeper/keeper.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, msg := range msgs {
+		for _, minFee := range feesParams.MinFees {
+			if msg.Type() == minFee.MessageType {
+				requiredFees = requiredFees.Add(minFee.Amount)
+			}
+		}
+	}
+
+	stableRequiredQty := requiredFees.Mul(sdk.NewDec(1000000))
```
RiccardoM, line 57: Why are we multiplying everything for 1.000.000 here?

x/fees/keeper/keeper.go
```diff
@@ -0,0 +1,69 @@
...
+	givenFeeAmount := sdk.NewDecFromInt(fees.Amount.AmountOf(feesParams.FeeDenom))
+
+	if !stableRequiredQty.IsZero() && stableRequiredQty.GT(givenFeeAmount) {
+		if stableRequiredQty.GT(givenFeeAmount) {
+			return sdkerrors.Wrap(sdkerrors.ErrInsufficientFee,
+				fmt.Sprintf("Expected %s %s amount, got %s", stableRequiredQty,
```
RiccardoM, line 63: I think this can be changed to produce messages like  `"expected at least <required_fees> but got <actual_fees> instead"

x/fees/types/min_fee.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func NewMinFee(messageType string, amount sdk.Dec) MinFee {
+	return MinFee{
+		MessageType: messageType,
+		Amount:      amount,
+	}
+}
+
+func (mf MinFee) String() string {
```
RiccardoM, line 22: I think this method is not necessary, and we can safely delete it

x/fees/types/params.go
```diff
@@ -0,0 +1,108 @@
...
+	MinFeesStoreKey  = []byte("MinFees")
+)
+
+// ParamKeyTable Key declaration for parameters
+func ParamKeyTable() paramsModule.KeyTable {
+	return paramsModule.NewKeyTable().RegisterParamSet(&Params{})
+}
+
+type Params struct {
+	FeeDenom string   `json:"fee_denom" yaml:"fee_denom"`
```
RiccardoM, line 33: I think that if the `MinFees.Amount` value is changed to use `sdk.Coins`, this can be safely deleted

x/posts/module.go
```diff
@@ -150,29 +151,43 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 //____________________________________________________________________________
 
 // AppModuleSimulation defines the module simulation functions used by the posts module.
-type AppModuleSimulation struct{}
+type AppModuleSimulation struct {
```
RiccardoM, line 154: I think that the changes in this file should all be reverted

x/profiles/module.go
```diff
@@ -148,29 +149,43 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
 //____________________________________________________________________________
 
 // AppModuleSimulation defines the module simulation functions used by the profile module.
-type AppModuleSimulation struct{}
+type AppModuleSimulation struct {
```
RiccardoM, line 152: All the changes here should be reverted. No need to declare a new `AppModuleSimulation` type when the `AppModule` already implements this

x/relationships/module.go
```diff
@@ -146,29 +147,43 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
 //____________________________________________________________________________
 
 // AppModuleSimulation defines the module simulation functions used by the profile module.
-type AppModuleSimulation struct{}
+type AppModuleSimulation struct {
```
RiccardoM, line 150: All the changes here should be reverted. No need to declare a new `AppModuleSimulation` type when the `AppModule` already implements this

x/reports/module.go
```diff
@@ -151,29 +152,45 @@ func (am AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.Valid
 //____________________________________________________________________________
 
 // AppModuleSimulation defines the module simulation functions used by the posts module.
-type AppModuleSimulation struct{}
+type AppModuleSimulation struct {
```
RiccardoM, line 155: All the changes here should be reverted. No need to declare a new `AppModuleSimulation` type when the `AppModule` already implements this

x/fees/keeper/keeper.go
```diff
@@ -0,0 +1,69 @@
...
+// GetParams returns the params from the store
+func (k Keeper) GetParams(ctx sdk.Context) (p types.Params) {
+	k.paramSubspace.GetParamSet(ctx, &p)
+	return p
+}
+
+// CheckFees checks whether the given fees are sufficient to pay for all the given messages.
+// The check is performed considering the minimum fee amounts specified inside the module parameters.
+func (k Keeper) CheckFees(ctx sdk.Context, fees authtypes.StdFee, msgs []sdk.Msg) error {
+	feesParams := k.GetParams(ctx)
```
RiccardoM, line 45: I think that we should check if the params are `nil` or not. Not having them set previously, this would return `nil` and then the access to its field would panic the whole application. 

I think that if no params are set, we should return here without an error (since there are indeed no min fees to be paid), instead of continuing the execution.

Also, a test case should be added to make sure that this particular case does not interrupt the code execution

RiccardoM (changes requested):
The implementation looks mostly OK. There are some things I recommend changing

x/fees/ante/ante.go
bragaz: Yup, others ante checks for particular txs types, so I think we needed that to be sure we're having only stdTx

x/fees/client/cli/query.go
bragaz: we use parameters in both profiles and posts modules, I think that in order to maintain the same name on every module we change them all or leave them like this

x/fees/keeper/keeper.go
bragaz: In order to bring the 0.01 to 10000

x/relationships/module.go
bragaz: I used this `AppModuleSimulartion` to be able to pass the `fees` keeper to the sims tests without touching the `AppModuleBasic` type.

x/fees/keeper/keeper.go
bragaz: I think that we could do it better by editing the `InitGenesis()` function and set the default params (so basically zero fees) whenever the passed `params` from genesis are empty. What do you think?

x/fees/types/min_fee.go
RiccardoM: Can you solve this please @bragaz

x/fees/types/min_fee.go
```diff
@@ -0,0 +1,33 @@
...
+	return fmt.Sprintf("Message Type: %s\nAmount: %s\n", mf.MessageType, mf.Amount)
+}
+
+// Validate check if the min fee parameters are valid
+func (mf MinFee) Validate() error {
+	if mf.MessageType == "" {
+		return fmt.Errorf("invalid minimum fee message type")
+	}
+
+	return nil
```
RiccardoM, line 32: We should also validate the `Amount` field by using the `IsValid` method of `sdk.Coins` here probably

x/fees/keeper/keeper.go
RiccardoM: I guess that wouldn't solve the problem. Let me explain it better. 

If the `fees` object is not set inside the genesis, the module will not even be initialized. This means that `InitGenesis` and all the other methods inside `module.go` will not be called. 

The problem is that, even thought the module is not going to be initialized, the `AnteHandler` is still going to be called if it is set. This will cause the `k.GetParams` to return `nil` (since they have never been initialized) and thus error. 

There are two alternatives in this case: 

1. We handle that error and return as soon as we identify that the params are `nil`:  
   ```go
   feesParams := k.GetParams(ctx)
   if feesParams == nil {
      return nil
   }
   ```

2. We return an error telling what's wrong:  
   ```go
   feesParams := k.GetParams(ctx)
   if feesParams == nil {
      return fmt.ErrorF("fees params cannot be null. Please make sure you have initialized the module correctly")
   } 
   ```

I guess that the best approach is number 2. This would allow the developer to know what's wrong and fix it accordingly.

x/fees/types/min_fee.go
bragaz: By default when creating a `Coin` or `Coins` type that check is performed automatically

bragaz:
@RiccardoM I had a look at the `nil` parameters problem and I found out two things:

1.  The code take care of checking if they're `nil` or not before returning them, with a check that, if fails, will cause a panic error.
2. When creating the different keepers inside the `app.go` file, the ones which have some parameters,  have a check inside the `NewKeeper` constructor which do the following:
```go
if !paramSpace.HasKeyTable() {
         paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
}
```

where the function `types.ParamKeyTable()` is declared as follow:
```go
// ParamKeyTable Key declaration for parameters
func ParamKeyTable() paramsModule.KeyTable {
	return paramsModule.NewKeyTable().RegisterParamSet(&Params{})
}
```

So basically we will never be in the case of having `nil` parameters.

x/fees/client/cli/query.go
RiccardoM: Let's leave them params for now then

x/fees/types/min_fee.go
RiccardoM: I think that check is performed only when creating the `Coin` or `Coins` using the `NewCoin` and `NewCoins` methods respectively. I don't think it's performed when the `Coins` are parsed from a JSON file. I think it might be better to add it for this reason anyway

RiccardoM:
@bragaz You're right. I've also checked if there would be some cases in which the params could be set as empty (`types.Params{}`) but it appear to be impossible. If such value is passed to the `SetParams` method it would panic since the validity is always checked first.

RiccardoM (approved):

PR #321, opened 2020-12-02
Leonardo/migrating fees

bragaz:
## Description
This PR migrates the `fees` module to the `v0.40` Stargate and solves all the conflicts of `riccardo/cosmos-v0.40`
branch.
## Checklist
- [x] Wrote unit tests.


x/fees/types/genesis_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}{
+		{
+			name:        "DefaultGenesis does not error",
+			genesis:     types.DefaultGenesisState(),
+			shouldError: false,
+		},
+		{
+			name: "Genesis with invalid params errors",
+			genesis: types.NewGenesisState(types.NewParams([]types.MinFee{
+				types.NewMinFee("", sdk.NewCoins(sdk.NewCoin("stake", sdk.NewInt(1))))})),
```
RiccardoM, line 24: Indentation
```suggestion
			genesis: types.NewGenesisState(
				types.NewParams([]types.MinFee{
					types.NewMinFee("", sdk.NewCoins(sdk.NewCoin("stake", sdk.NewInt(1)))),
				}),
			),
```

x/fees/keeper/querier_test.go
```diff
@@ -0,0 +1,38 @@
...
+			expResult: types.DefaultParams(),
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		suite.Run(test.name, func() {
+			suite.SetupTest()
+
+			suite.keeper.SetParams(suite.ctx, types.DefaultParams())
```
RiccardoM, line 28: Instead of always using `types.DefaultParams()`, I would suggest specifying the value to be stored instead each use case, to make sure that even non-default values are returned properly

x/fees/ante/ante_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/desmos-labs/desmos/x/posts/types"
+	"github.com/tendermint/tendermint/crypto"
+)
+
+func (suite *AnteTestSuite) TestAnteHandlerFees_MsgCreatePost() {
+	suite.SetupTest(true) // reset
```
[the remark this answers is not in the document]
bragaz: my error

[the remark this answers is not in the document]
bragaz: because it's used for the account creation in line 13

x/fees/types/min_fee.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	if !mf.Amount.IsValid() {
+		return fmt.Errorf("invalid minimum fee amount")
+	}
+
+	return nil
+}
+
+type MinFees []MinFee
```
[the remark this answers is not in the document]
bragaz: No we can't..I had to add this since without it I was having problem of mismatching types due to the protobuf generations

x/fees/ante/ante_test.go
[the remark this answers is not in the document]
RiccardoM: Can we move also the account creation just before running the test? I think that's the best place where to put it to make sure the sequence, balance etc are all the same for each test

x/fees/keeper/keeper_test.go
```diff
@@ -0,0 +1,138 @@
...
+		"poll?",
+		time.Date(2050, 1, 1, 15, 15, 00, 000, timeZone),
+		posts.NewPollAnswers(
+			posts.NewPollAnswer("1", "Yes"),
+			posts.NewPollAnswer("2", "No"),
+		),
+		false,
+		true,
+	)
+	attachments := posts.NewAttachments(posts.NewAttachment("https://uri.com", "text/plain", nil))
```
RiccardoM, line 73: I don't think it's necessary to specify a poll and some attachments since the fee checker does not consider them when computing the fees to apply. For this reason, I suggest removing them

x/fees/keeper/keeper_test.go
```diff
@@ -0,0 +1,138 @@
...
+	id := "dd065b70feb810a8c6f535cf670fe6e3534085221fa964ed2660ebca93f910d1"
+
+	var testOwner = "cosmos1cjf97gpzwmaf30pzvaargfgr884mpp5ak8f7ns"
```
RiccardoM, line 76: I would suggest manually specifying these values inside the different test cases, instead of declaring them here. I don't see any particular advantage in allocating them here since we are inside a test

x/fees/keeper/keeper_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	tests := []struct {
+		name      string
+		params    types.Params
+		givenFees sdk.Coins
+		msgs      []sdk.Msg
+		expError  error
```
RiccardoM, line 83: Instead of specifying an error, I suggest just checking if an error is raised or not. It has no particular advantage for us to check the specific error message and type. Instead, it may make it harder to maintain tests in the future. For example, if we change the error message, we need to also update the test. This is a huge overhead to the development process, without any particular advantage for us IMO.

This means making `expError bool` instead of `error`, and then inside the test simply do: 


```golang
if test.expError {
  suite.Require().Error(err)
} else {
  suite.Require.NoError(err)
}
```

PR #327, opened 2020-12-09
Create post improvements

RiccardoM:
## Description
This PR changes how a post is created. Now, all data should be passed to the `NewPost` method. This is to make sure that projects using Desmos as a dependency have a better way of creating `Post` instances. Also, it should remove non-determinisim from the tests, which was introduced with the `ComputeID` method being used everywhere and doing side-effects.

## Checklist
- [x] Wrote unit tests.


bragaz (approved):
💯  3rd party love this PR

PR #328, opened 2020-12-09
Stargate migration

bragaz:
## Description
This PR contains the migration of all our modules from the `v0.14`(even `v0.13`) to the version `v0.15`.
I written but commented the integration test for migration cause we need first to migrate from `v0.12.2`to `v0.13.0`.
## Checklist
- [x] Wrote unit tests.

### DIFF
--- a/x/genutil/client/cli/migrate.go
+++ b/x/genutil/client/cli/migrate.go
@@ -22,8 +22,8 @@ import (
 // a version of the chain to the next one. It contains an array as we need to support Cosmos SDK migrations
 // too if needed.
 var migrationMap = map[string]types.MigrationCallback{
-	"v0.13.0": v0130.Migrate,
-	"v0.15.0": v0150.Migrate,
+	"v0.13.0":  v0130.Migrate,
+	"stargate": v0150.Migrate,
 }
 
 const (
--- a/x/genutil/client/cli/migrate.go
+++ b/x/genutil/client/cli/migrate.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"encoding/json"
 	"fmt"
+	v0150 "github.com/desmos-labs/desmos/x/genutil/legacy/v0.15.0"
 	"sort"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 // too if needed.
 var migrationMap = map[string]types.MigrationCallback{
 	"v0.13.0": v0130.Migrate,
+	"v0.15.0": v0150.Migrate,
 }
 
 const (
--- a/x/genutil/client/cli/migrate.go
+++ b/x/genutil/client/cli/migrate.go
@@ -3,21 +3,19 @@ package cli
 import (
 	"encoding/json"
 	"fmt"
-	v0150 "github.com/desmos-labs/desmos/x/genutil/legacy/v0.15.0"
 	"sort"
 	"time"
 
-	"github.com/cosmos/cosmos-sdk/types/errors"
-
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/version"
 	extypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
+	v0130 "github.com/desmos-labs/desmos/x/genutil/legacy/v0.13.0"
+	v0150 "github.com/desmos-labs/desmos/x/genutil/legacy/v0.15.0"
+	"github.com/desmos-labs/desmos/x/genutil/types"
 	"github.com/spf13/cobra"
 	tm "github.com/tendermint/tendermint/types"
-
-	v0130 "github.com/desmos-labs/desmos/x/genutil/legacy/v0.13.0"
-	"github.com/desmos-labs/desmos/x/genutil/types"
 )
 
 // migrationMap contains the list of migrations that should be performed when migrating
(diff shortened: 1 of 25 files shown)


x/genutil/client/cli/migrate.go
```diff
@@ -6,24 +6,24 @@ import (
...
+	"github.com/spf13/cobra"
+	tm "github.com/tendermint/tendermint/types"
 )
 
 // migrationMap contains the list of migrations that should be performed when migrating
 // a version of the chain to the next one. It contains an array as we need to support Cosmos SDK migrations
 // too if needed.
 var migrationMap = map[string]types.MigrationCallback{
 	"v0.13.0": v0130.Migrate,
+	"v0.15.0": v0150.Migrate,
```
RiccardoM, line 26: I'd prefer we call this `stargate` as  I would like to call version 0.15.0 this way. No need to change the packages names though

x/posts/legacy/v0.15.0/migrate.go
```diff
@@ -0,0 +1,71 @@
...
+	return posts
+}
+
+func ConvertUsersPollAnswers(oldUsersPollAnswers map[string][]v040posts.UserAnswer) []UserAnswersEntry {
+	var userAnswersEntries []UserAnswersEntry
+	for key, value := range oldUsersPollAnswers {
+		userAnswersEntries = append(userAnswersEntries, newUserAnswerEntry(key, value))
+	}
+
+	return userAnswersEntries
```
RiccardoM, line 47: I think this can be made more efficient by pre-allocating the slice and then using the index to create the values:

```suggestion
	userAnswersEntries := make([]UserAnswersEntry, len(oldUsersPollAnswers))
	index := 0
	for key, value := range oldUsersPollAnswers {
		userAnswersEntries[index] = newUserAnswerEntry(key, value)
		index++
	}

	return userAnswersEntries
```

x/posts/legacy/v0.15.0/migrate.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func ConvertPostReactions(oldPostReactions map[string][]v060.PostReaction) []PostReactionsEntry {
+	var postReactionsEntries []PostReactionsEntry
+	for key, value := range oldPostReactions {
+		postReactionsEntries = append(postReactionsEntries, newPostReactionEntry(key, value))
+	}
+
+	return postReactionsEntries
```
RiccardoM, line 56: This can be made more efficient as well

```suggestion
	postReactionsEntries := make([]PostReactionsEntry, len(oldPostReactions))
	index := 0
	for key, value := range oldPostReactions {
		postReactionsEntries[index] newPostReactionEntry(key, value)
		index++
	}

	return postReactionsEntries
```

x/posts/legacy/v0.15.0/migrate_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	migrated := v0150posts.Migrate(v0130genesisState)
+
+	// Check for posts
+	require.Len(t, expectedGenState.Posts, len(migrated.Posts))
+	for index, post := range migrated.Posts {
+		require.Equal(t, expectedGenState.Posts[index].PostID, post.PostID)
+		require.Equal(t, expectedGenState.Posts[index].ParentID, post.ParentID)
+		require.Equal(t, expectedGenState.Posts[index].OptionalData, post.OptionalData)
```
RiccardoM, line 168: In order to make sure we are migrating all the data, we should check all the different fields for equality. Also, it might be better to use a variable to simplify this check: 

```golang
		expected := expectedGenState.Posts[index]
		require.Equal(t, expected.PostID, post.PostID)
		require.Equal(t, expected.ParentID, post.ParentID)
		require.Equal(t, expected.OptionalData, post.OptionalData)
		require.Equal(t, expected.AllowsComments, post.AllowsComments)
		// ...
```

x/posts/legacy/v0.15.0/migrate_test.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	}
+
+	// Check for registered reactions
+	require.Len(t, expectedGenState.RegisteredReactions, len(migrated.RegisteredReactions))
+	for index, regReaction := range migrated.RegisteredReactions {
+		require.Equal(t, expectedGenState.RegisteredReactions[index].Creator, regReaction.Creator)
+		require.Equal(t, expectedGenState.RegisteredReactions[index].ShortCode, regReaction.ShortCode)
+		require.Equal(t, expectedGenState.RegisteredReactions[index].Subspace, regReaction.Subspace)
+		require.Equal(t, expectedGenState.RegisteredReactions[index].Value, regReaction.Value)
```
RiccardoM, line 195: I would suggest allocating a variable to simplify the checks here

```suggestion
		expected := expectedGenState.RegisteredReactions[index]
		require.Equal(t, expected.Creator, regReaction.Creator)
		require.Equal(t, expected.ShortCode, regReaction.ShortCode)
		require.Equal(t, expected.Subspace, regReaction.Subspace)
		require.Equal(t, expected.Value, regReaction.Value)
```

RiccardoM (commented):
The implementation looks overall good. I would only suggest one thing that I've seen might cause problems. 

Currently, the types used as the destination of the migration (this means `v0.15.0` `GenesisState`) are defined inside the `legacy/v0.15.0` packages. What I would do is, instead of re-defining them there by hand, using Protobuf instead. This could be achieved by either: 

1. Using the structs that are now defined inside the `poststypes`, `relationshipstypes` etc packages.  

2. Copy/paste the Protobuf definitions inside the `legacy` folders and generate new implementations there. 

To be honest, I think number 1 is preferred in this case cause it would reduce any code duplication and make everything easier.

RiccardoM (approved):

PR #347, opened 2021-01-19
Version v0.15.1

### LINKED ISSUE #345: Update Cosmos SDK to v0.40.1
Tendermint is going to release a security patch in Tendermint Core v0.34.3 and Cosmos SDK v0.40.1 at 16:00 UTC on Tuesday, January 19, 2021. 

https://forum.cosmos.network/t/tendermint-core-security-advisory-mulberry/4309

Desmos should be updated asap once the new SDK version is released.

### LINKED ISSUE #348: Remove the usage of json inside genesis-related commands
## Context
Currently we're using a lot of `json.Marshal` and `json.Unmarshal` inside genesis-related methods (eg. `DefaultGenesis`, `VerifyGenesis` and `InitGenesis`). This is done due to cosmos/cosmos-sdk#8333.

Although it works, it makes it incredibly hard to manage things, as we need to always remember what to use (`json` vs `cdc`).

## Implementation proposal
I think we should remove all the `gogoproto.jsontag` usage from our Proto file definitions, and revert to using only `cdc` instead. This would make it easier and simplify a lot of things in the future.

RiccardoM:
## Description
- Fixed `VerifyGenesis` and `InitGenesis`
- Updated Cosmos to v0.40.1 (closes #345) 
- Removed `json` usage from genesis-related methods (closes #348)

## Checklist
- [ ] Wrote unit tests.


x/profiles/types/genesis.go
```diff
@@ -31,12 +37,41 @@ func ValidateGenesis(data *GenesisState) error {
...
 
 	return nil
 }
+
+// containDuplicates tells whether or not the profiles slice contain duplicates of the given profile
+func containDuplicates(profiles []Profile, profile Profile) bool {
+	var count = 0
+	for _, p := range profiles {
+		if p.Equal(profile) {
+			count++
```
bragaz, line 63: Wasn't it better to return `true` here so it stops when found a duplicate and not go through all the profiles array?
Am I missing some purpose with these implementation?

x/relationships/types/genesis.go
```diff
@@ -31,3 +39,14 @@ func ValidateGenesis(data *GenesisState) error {
 
 	return nil
 }
+
+// containDuplicates tells whether the given relationships slice contain duplicates of the provided relationship
+func containDuplicates(relationships []Relationship, relationship Relationship) bool {
+	var count = 0
+	for _, r := range relationships {
+		if r.Equal(relationship) {
+			count++
```
bragaz, line 48: I don't understand why you count the duplicates if you're not using that number later

x/reports/types/genesis.go
```diff
@@ -14,10 +16,26 @@ func DefaultGenesisState() *GenesisState {
...
 	}
 	return nil
 }
+
+// containDuplicates tells whether the given reports slice contain duplicates of the provided report
+func containDuplicates(reports []Report, report Report) bool {
+	var count = 0
+	for _, r := range reports {
+		if r.Equal(report) {
+			count++
```
bragaz, line 37: like above

bragaz (changes requested):
some small doubts about `containDuplicates` functions

x/profiles/types/genesis.go
RiccardoM: @bragaz This method is used to make sure that the given `profile` does not appear 2 or more times inside the profiled slice. This is used to check that, for any profile, there are no duplicates of that profile. This means that inside the given slice, there will always be 1 instance of that profile, but we want to check if there are 2 or more. For this reason, we cannot stop at the first instance that is equal to the given profile, because that will be the profile itself. Instead, we need to count all the duplicated occurrences and return whether or not there are 2 or more.

bragaz: Oh right. Gotcha

bragaz (approved):
👍

PR #385, opened 2021-03-15
Change how profiles are stored

### LINKED ISSUE #261: Change how profile data are stored
## Context
Currently the profiles are stored using a custom keeper and they are saved using the following data structure: 

```go
type Profile struct {
  DTag         string         `json:"dtag" yaml:"dtag"`
  Moniker      *string        `json:"moniker,omitempty" yaml:"moniker,omitempty"`
  Bio          *string        `json:"bio,omitempty" yaml:"bio,omitempty"`
  Pictures     *Pictures      `json:"pictures,omitempty" yaml:"pictures,omitempty"`
  Creator      sdk.AccAddress `json:"creator" yaml:"creator"`
  CreationDate time.Time      `json:"creation_date" yaml:"creation_date"`
}
```

However, since a _profile_ is just an extension of a Cosmos _account_, I think we should merge the two concepts into a single one. 

## Solution
In order to achieve a better organization of the code, what we could do is extend the Cosmos `Account` interface with the data that are currently defined inside the `Profile` structure. To do so, we could simply define a `Profile` as being the following: 

```go
type Profile struct { 
  *BaseAccount 
  DTag         *string        `json:"dtag" yaml:"dtag"`
  Moniker      *string        `json:"moniker,omitempty" yaml:"moniker,omitempty"`
  Bio          *string        `json:"bio,omitempty" yaml:"bio,omitempty"`
  Pictures     *Pictures      `json:"pictures,omitempty" yaml:"pictures,omitempty"`
}
```

This would have the following pros.

- **No need to have a custom saving mechanism**  
   By extending the `Account` interface, we could allow the `auth` module to take care of storing the profiles data.

- **No duplicated data**  
   Since the `Account` interface already contains an `Address` field, we would not have to include that inside the `Profile` structure (currently the `Creator` field). 

- **Easier future extension**  
   If other fields are going to be added in the future to the `Account` interface (eg. different types of keys), then we would have them directly inside the profiles as well. An example could be the address defined inside #258

-  **Easier data retrieval**  
   Since a profile would be a simple account, one single API call (`/auth/<address>`) would be sufficient to get all the latest info about a profile 

What do you guys @kwunyeung @bragaz think about this?

RiccardoM:
## Description
This PR changes how the profiles are implemented, making them implement the `authtypes.AccountI` interface and allowing them to be stored inside the `x/auth` keeper. 

Closes #261

### Notes
Currently, the only unit test failing should be `TestCmdQueryProfile/existing_profile_is_returned_properly`. This is due to the fact that right now the Cosmos SDK does not allow to specify custom accounts inside the in-memory network that is used when testing the CLI commands. I've opened cosmos/cosmos-sdk#8936 to fix this, so we will have to wait until that is merged and put inside a release to make this test work properly.

Also, currently the simulation tests fail because it appear that the `RandomGenesisAccount` method that is present inside the sim package does not initialize accounts with a public key. This leads to a bug that makes it impossible to sign message with any of those accounts. I've written this also inside the [#v0.41-stargate Discord channel](https://discord.com/channels/669268347736686612/723170066937413693/822434506006593566). I hope to get a PR accepted and merged the next week to fix this as well. 

**Update 16/03/2021**
I've opened a PR for the second error as well: cosmos/cosmos-sdk#8957

**Update 14/04/2021**  
I've forked the Cosmos SDK `v0.42.4` code and added some commits that allow to fix the above problems. The code is available [here](https://github.com/RiccardoM/cosmos-sdk/tree/v0.42.x-fixes) and it includes the changes from the following PRs: 

- cosmos/cosmos-sdk#8936 
- cosmos/cosmos-sdk#8957

## Checklist
- [x] Wrote unit tests.


x/profiles/types/account_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+	tests := []struct {
+		name       string
+		original   *types.Profile
+		update     *types.ProfileUpdate
+		expError   bool
+		expProfile *types.Profile
+	}{
+		{
+			name: "DoNotModify and empty fields do not update original values",
```
bragaz, line 34: why empty fields don't update the original values here? A test below say a different thing. Can we remove it?
```suggestion
			name: "DoNotModify do not update original values",
```

bragaz (changes requested):

bragaz:
Also, it looks like there's an error inside profiles CLI tests

x/profiles/types/account_test.go
RiccardoM: Thanks, I've cleared up the test name and also the test making sure that empty values effectively update the original ones

bragaz (approved):
Ready to ship! 🚀

PR #389, opened 2021-03-31
Mainnet-beta codebase

bragaz:
## Description
This PR contains the code that will be used for mainnet-beta testnet.
It's made only for review purposes and not intended to be merged.

I've basically disabled the txs from the modules that we don't want to include in the mainnet-beta.
I was thinking to edit the main `app.go` file but then I reasoned that this was a good solution that would as long it requires the minimum edits and make basically impossible to anyone to make transactions for those modules.
Do you think this is reasonable? 

## Checklist
- [ ] Wrote unit tests.


RiccardoM:
@bragaz I think there might be a better way of not including some modules inside the running code than to comment out the various return statements from withing the `module.go` files.

Currently, the Cosmos SDK initializes the different things for each module by calling the various modules' methods you've commented. However, those methods are called for a module, only if the module is included inside the `ModuleBasics` slice. If a module's `AppBasic` is not included there, then it won't be initialized.

That being said, what I suggest we could do is: 

1. Revert all those comments, leaving the proper calls where they were previously. 
2. Remove all the modules we do not want included in our mainnet from the `ModuleBasics` slice. 

Additionally, what we could do to make things even more clear to people using Desmos as a dependency package, is we could move all the non-mainnet modules into a `staging` folder inside the `x` directory: 

```
x
|- profiles
|- staging
|     |- posts
|     |- fees
|     |- reports
|     |- ...
```

This way, we can distinguish between those modules that are ready to be used  (they will appear directly inside the `x/` folder), and those which need more testing (in this case, they will be inside the `staging/` folder). What do you think?

RiccardoM (commented):
@bragaz I think there might be a better way of not including some modules inside the running code than to comment out the various return statements from withing the `module.go` files.

Currently, the Cosmos SDK initializes the different things for each module by calling the various modules' methods you've commented. However, those methods are called for a module, only if the module is included inside the `ModuleBasics` slice. If a module's `AppBasic` is not included there, then it won't be initialized.

That being said, what I suggest we could do is: 

1. Revert all those comments, leaving the proper calls where they were previously. 
2. Remove all the modules we do not want included in our mainnet from the `ModuleBasics` slice. 

Additionally, what we could do to make things even more clear to people using Desmos as a dependency package, is we could move all the non-mainnet modules into a `staging` folder inside the `x` directory: 

```
x
|- profiles
|- staging
|     |- posts
|     |- fees
|     |- reports
|     |- ...
```

This way, we can distinguish between those modules that are ready to be used  (they will appear directly inside the `x/` folder), and those which need more testing (in this case, they will be inside the `staging/` folder). What do you think?

bragaz:
@RiccardoM All this makes sense to me 👍. I'll proceed like that then.

RiccardoM (commented):
Looks good to me, @bragaz. What do you think about removing entirely the `magpie` module? I don't think it;'s necessary anymore. We will replace it with IBC anyway and we can remove its code and docs safely I think.

bragaz:
@RiccardoM Yes I think we can remove it safely.

RiccardoM (commented):
I think we should change the different `option go_package` values of the `.proto` file to match the new packages. 

Example: 
```proto
option go_package = "github.com/desmos-labs/desmos/x/posts/types";
```

Should be 


```proto
option go_package = "github.com/desmos-labs/desmos/x/staging/posts/types";
```

app/sim_test.go
```diff
@@ -16,11 +16,10 @@ import (
...
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
-	magpieTypes "github.com/desmos-labs/desmos/x/magpie/types"
-	postsTypes "github.com/desmos-labs/desmos/x/posts/types"
 	profilesTypes "github.com/desmos-labs/desmos/x/profiles/types"
-	relationshipsTypes "github.com/desmos-labs/desmos/x/relationships/types"
-	reportsTypes "github.com/desmos-labs/desmos/x/reports/types"
+	postsTypes "github.com/desmos-labs/desmos/x/staging/posts/types"
+	relationshipsTypes "github.com/desmos-labs/desmos/x/staging/relationships/types"
+	reportsTypes "github.com/desmos-labs/desmos/x/staging/reports/types"
```
RiccardoM, line 22: Can we use the same naming conventions as other imports (all lowercase)? `profilesTypes` -> `profilestypes`

x/staging/fees/simulation/genesis.go
```diff
@@ -7,25 +7,23 @@ import (
 	"fmt"
 	"math/rand"
 
-	magpieTypes "github.com/desmos-labs/desmos/x/magpie/types"
-	postsTypes "github.com/desmos-labs/desmos/x/posts/types"
 	profilesTypes "github.com/desmos-labs/desmos/x/profiles/types"
-	relationshipsTypes "github.com/desmos-labs/desmos/x/relationships/types"
-	reportsTypes "github.com/desmos-labs/desmos/x/reports/types"
+	postsTypes "github.com/desmos-labs/desmos/x/staging/posts/types"
```
RiccardoM, line 11: Same here for the imports: `postsTypes` -> `poststypes`

x/staging/posts/client/cli/cli_test.go
```diff
@@ -26,7 +26,8 @@ type IntegrationTestSuite struct {
 }
 
 func TestIntegrationTestSuite(t *testing.T) {
-	suite.Run(t, new(IntegrationTestSuite))
+	//TODO restore this when out of staging
+	//suite.Run(t, new(IntegrationTestSuite))
```
RiccardoM, line 30: Why has this been commented out? Can't we leave it as it was?

x/staging/relationships/client/cli/cli_test.go
```diff
@@ -28,7 +28,8 @@ type IntegrationTestSuite struct {
 }
 
 func TestIntegrationTestSuite(t *testing.T) {
-	suite.Run(t, new(IntegrationTestSuite))
+	//TODO restore this when out of staging
+	//suite.Run(t, new(IntegrationTestSuite))
```
RiccardoM, line 32: Can we restore this as well?

x/staging/reports/client/cli/cli_test.go
```diff
@@ -27,7 +27,8 @@ type IntegrationTestSuite struct {
 }
 
 func TestIntegrationTestSuite(t *testing.T) {
-	suite.Run(t, new(IntegrationTestSuite))
+	//TODO restore this when out of staging
+	//suite.Run(t, new(IntegrationTestSuite))
```
RiccardoM, line 31: As above

x/staging/posts/client/cli/cli_test.go
bragaz: Those tests are failing without the loaded modules

RiccardoM: Hmmm, I see

RiccardoM:
> Should we adapt the rest of the staging modules docs? Maybe writing something like (staging) alongside the names of the modules? Just to avoid confusion.

@bragaz, sorry, didn't see this before. I think we can move them inside a staging folder as well. So that it should be easier to distinguish them

bragaz:
@RiccardoM I don't understand why after renaming `.proto` files `option_package` the `Breakage` check is failing.
Does it fail everytime some change to the files is made?

RiccardoM (approved):

Issue #192, opened 2020-06-09
IBC-enabled social profile


kwunyeung:
## Feature description 

The decentralized profile can serve as a central reference of a user profile. There is a pain point of validators and wallets. Validators are validating multiple chains. They have to setup their profiles in different chains every time. Every time they have to input their moniker, communication channels, bio, terms and conditions etc. The decentralized profile can serve as a reference to the validators, explorers and wallets can retrieve the profile directly from the decentralized profile. This also serve the users who are using the wallets if they would like to show extra profile information in their wallet. 

## Implementation proposal

The decentralized profile can be done by sending a signature from chain A to `DESMOS` via IBC sent by the profile owner account on `DESMOS`. Once the signature is verified, `DESMOS` will update a reference of the profile to the account on the chain A. We can then verify that the profile belongs to the address on chain A. The user on chain A should keep an open IBC channel between chain A and `DESMOS`. This will be considered the liveness of the profile of the user on chain A.

Superses #56

RiccardoM:
I personally think that this feature is very interesting, and could lead Desmos to be the main repository for different profiles that can be used inside wallets, explorers and other dApps as well. 

That being said, I would like to expand the implementation proposal you have made, to make sure I understood it correctly before making any comment on it. 

What you have proposed is a system that works like this (for the sake of simplicity I will assume a Cosmos <-> Desmos communication): 

#### 1. Signature creation
First of all, Alice creates a signature object. Such object should be created as follows: 

```js
let firstSignature = sign(Alice.CosmosAddress, Alice.CosmosPrivKey)
let secondSignature = sign(firstSignature, Alice.DesmosPrivKey)
```

This way with a double signature we are verifying that Alice owns both the `cosmos1...` and `desmos1...` accounts.

#### 2. IBC packet creation
Once the signature has been created, an IBC packet will be formed. Such packet must contain: 

1. Alice's Desmos address in HEX form. 
2. Alice's Cosmos address in HEX form. 
3. Alice's Cosmos public key in HEX form.
4. The previously created `firstSignature` and `secondSignature` in HEX format. 


An example could be the following JSON packet: 

```json
{ 
  "source_address": "hex(Alice.CosmosAddress)",
  "source_pub_key": "hex(Alice.CosmosPubKey)",
  "destination_address": "hex(Alice.DesmosAddress)",
  "first_signature": "hex(<firstSignature>)",
  "second_signature": "hex(<secondSignature>)"
}
```

[...]

PR #412, opened 2021-05-04
Implement IBC links module

dadamu:
## Description

This PR implement #192, creating account link from different chains which integrated links module by IBC protocol.

#### Links module does:
* Create link struct on destination chain after relaying successfully
* Support create-ibc-link method to create a link by one key
* Provide create-ibc-connection method to create a link by two private keys used on target chains 
* Offer query method to get a link info with source address

## Checklist
- [x] Wrote unit tests.


x/links/types/types.go
```diff
@@ -0,0 +1 @@
+package types
```
bragaz, line 1: I think you can remove this file since is empty

x/links/types/expected_keeper_ibc.go
```diff
@@ -0,0 +1,21 @@
+package types
+
```
bragaz, line 2: I think you don't need to rewrite these interfaces, you can just import the cosmos ones where you need them with this import:
```go
import (
ibcTypes "github.com/cosmos/cosmos-sdk/x/ibc/applications/transfer/types"
)
```

bragaz (changes requested):
In the current implementation it looks like it isn't possible to create more than one link from Desmos to other chains. I think we should also handle this possibility, letting users having links with multiple chains.

To do so, we could save an array of `[]link` associated with an address, instead of a single link. In doing so, we should also update the `StoreLink` method and the `GetLink` one in something similar to this:
```go
GetLinks(ctx sdk.Context, address string) []types.Link
``` 

I would like to know also your thoughts: @dadamu
CC: @RiccardoM

dadamu:
I think the most common case is google account link service. An account from a website should only have one linked account on google. So, a source address on certain chain linked to a destination address makes more sense to me.

RiccardoM:
@dadamu You're right, inside **centralized** websites and applications you often want to have a single `source -> destination` connection, as you mentioned with Google where links are represented as 1-to-1 relationships. 

Inside IBC, however, there might be cases where we want to have an n-to-1 `source -> destination` connection. As an example, consider the case in which a Cosmos validator has two different `cosmos` addresses: 

1. the first address, which represents the self-delegation address of their validator; 
2. the second address, which represents a public account that the validator uses to perform other actions. 

In this case, we should allow such validator to connect **both** their `cosmos` accounts to a single Desmos account if they wish to do so. By doing that, we allow them to verify both the `self_delegation` and the other account with a single Desmos profile, instead of requiring them to create two distinct (but identical) Desmos profiles. 

Generally speaking, we should make sure that our IBC links are `n-to-n` relationships, where we can: 

- connect multiple accounts from the same `source_chain` to a single Desmos profile
- connect a single Desmos profile to multiple accounts on different `destination_chain`

RiccardoM:
@dadamu After discussing a bit with @bragaz, we found the best way to solve all the possible problems. Before illustrating it, let me resume all the things we want to achieve, so that we are all on the same page. 

What we want to have is the ability to link external addresses (`cosmos`, `akash`, etc) to a Desmos profile. This will be called **cross-chain account links**.  
Knowing that one person might have multiple external addresses of the same chain, we should allow them to link all of those addresses to the same profile. For this reason, we should allow links such as `cosmos1 -> desmos1` and `cosmos2 -> desmos1`. 

Now, inside Desmos users might have multiple accounts as well. As inside Cosmos, a validator might have their self delegate address as well as another address that should be linked to the same profile.  
To solve this problem, I think the best way is to implement **inter-chain account links**. What this means is that I can create a link such as `desmos2 -> desmos1`. By doing so, I'm telling everyone that I own both `desmos1` and `desmos2`, and that the profile information will be stored only inside `desmos1`. 

To make this practical let's assume we have the following situation: 

1. Bob is a user with multiple account on different chains. Particularly, he has two Cosmos accounts:  
   - a `cosmos_val` account, which is the account he uses when running validator-oriented business
   - a `cosmos_priv` account, which he uses to store his personal tokens.
2. Bob also has two different Desmos accounts:
   - a `desmos_val` account, used to run a validator business
   - a `desmos_priv` account, which holds his personal DSM tokens
3. Bob wants to create a single Desmos profile, connecting to it all his accounts

To achieve what Bob wants at point (3), what he should do is: 

1. Create a single Desmos profile using the `desmos_priv` account. 
2. Connect `cosmos_val -> desmos_priv` (cross-chain link)
3. Connect `cosmos_priv -> desmos_priv` (cross-chain link)
4. Connect `desmos_val -> desmos_priv` (cross-chain link) 

The best way to represent this inside the chain storage is using a slice of `Link` objects (`[]Link`) where each link is defined as: 
```golang
type Link struct {
  ChainID string  // The chain id of the source chain (eg. cosmos-hub4, desmos-mainnet1, etc)
  Address string  // The address inside the source chain (eg. cosmos_priv, desmos_val, etc)
  Proof   string  // The cryptographic proof of the connection
}
```

Such slice can then be simply stored inside a `Profile` object: 
```golang
type Profile struct {
  ...
  Link []Link
}
```


This will make it super easy to get all the links of a user. 

Also, we should make sure that **one address can be linked only to a single profile**. To do this, the best way is to also store the following associations on-chain: 

```
cosmos_priv -> desmos_priv
cosmos_val -> desmos_priv
desmos_val -> desmos_priv
```

By doing so, making sure that an external address is linked only to a Desmos profile will result in an O(1) query. 

Special precautions must be taken when **deleting** a link, since we want to remove it from the Desmos profile as well as this table.

Please @dadamu let me know if there is anything missing from here or if you think we can improve this somehow

dadamu:
@RiccardoM It is interesting idea to have an inter-chain link. What we would like to do is making sure that one address only links to  a profile, and let profile be able to show all linked addresses, right?

Based on it, there may be a problem on inter-chain link mechanism As I know, every desmos address can create a profile. Then, if an address `desmos1` links to `desmos2` after creating its profile, `desmos1` address will have two profiles.

image

Subsequently, if addresses from other chains links to `desmos2` address, the chain can't identify which profile should be linked, because both `desmos1` and `desmos2` have a profile and `desmos1` links to `desmos2` already.

image

Moreover, links module will couple profile modules so that other chains can't import links module directly.

#### My solution

image

Remaining the cross-chain link in the address layer, and having inter-chain link in the profile layer to make sure a desmos address only have one profile. The structure will be a tree but height is only 3.

And let `Profile`  object be related to `Address`.
```
type Profile struct {
  ...
  Address []Address
}
```
As the result, BFS search can be used to get all internal addresses and external addresses in the profile.

bragaz:
@dadamu I would prefer to add a `link` slice instead of a simple `addresses` one, it will make the whole profile representation more explicit and better understandable by other developers. 
For the inter-chain link I think we can also let the two desmos addresses linked together and perform these checks below to be sure the double-profile problem you mentioned will never appear:
1) Inside the `x/links` module before creating the inter-chain link:   
     -  Check if any of the two addresses own a profile (which should be O(1) complexity)
     -  If both has it, doesn't allow to create the link between them, otherwise create the link.
2) Inside the `x/profiles` module before creating a profile:
      - Check if there's an active inter-chain link
      - If this link exists, check if the other linked address has a profile, if so return an error, else create it.
Did this make sense to you @dadamu ? CC @RiccardoM

dadamu:
@bragaz Great, it is good way to prevent from double-profile problem.

It seems that the chain which want to import `x/link` module should also import `x/profile` module because it checks if the address has profile or not before creating a link.
Could other chain import the `x/link` module without `x/profile` module? or they should build a their `x/link` to communicate with ours?

RiccardoM:
I personally think that @bragaz solution is the best. This should allow us to check very quickly for all the conditions to be satisfied. 

Suppose we want to represent the following connections: 

- `cosmos_priv -> desmos_priv`
- `cosmos_val -> desmos_priv`
- `desmos_val -> desmos_priv`


and that `desmos_priv` has a Profile. Then we should store things as 

```
desmos_priv Profile {
  Links: [
    cosmos_priv
    cosmos_val
    desmos_val
  ]
}


LinkKey + desmos_val -> desmos_priv
```

The link `desmos_val -> desmos_priv` must be represented in two ways because we will later need the second one to perform the checks that @bragaz proposed in the most efficient way possible. 

---

@dadamu Aside from this, honestly I **would not** split things into two modules (`x/link` and `x/profiles`). I would simply keep everything inside the `x/profiles` module without complicating things. 

This because other chains do not need to import the whole module inside their code to interact with us. They will simply have to import the types that represent IBC packets and they should be good to go. No need to have all the logic, they simply need the different types to be able to build such packets and send them over to our chain to be processed.

dadamu:
@RiccardoM @bragaz  Thanks for answering my questions. I agree with your idea, just I think importing whole module is convenient to other chains, like integrating with `ibc/transfer` module, we don't need to build the whole keeper and storage logic by our hand.

<hr>

Let me clean up the processes of links with some cases by the following image.

IBC5

### Create a link from `Desmos4` to `Desmos3`

1. Check `Desmos4` do not have its own profile
2. Make sure `Desmos4` does not link to other profile
3. Link `Desmos4` to `Desmos3`
4. Add a new link to `Profile1`
5. Traversal all child links ( `Link3` ) then insert to `Profile1`

### Create a link from `Desmos3` to `Desmos1`

1. Check `Desmos3` do not have its own profile
2. Stop the process because `Desmos3` links to a profile

### Remove `Link1`

1. Remove `Link1` in storage and in the profile
2. Traversal all links related `Link1`
3. Remove all links ( `Link2`, `Link3` ) in the profile

Please @RiccardoM  @bragaz  let me know if there is anything missing from here.

RiccardoM:
@dadamu Wow, that example actually makes it look like the linking is very complicated. What if we allow only already existing profiles to be linked to other addresses? In this case we require the user to only link everything to `Profile1` instead of having a multi-depth tree of links

dadamu:
@RiccardoM Yeah, That is why I think inter-chain link should be implemented on the profile layer. We only need to check whether a address has linked profile or not.

I think allowing only already existing profiles to be linked to other addresses will be like my solution mentioned above because it should be ensured that a address can have only one profile.

In the following example, only `Desmos2` status should be concerned. It is easier to create inter-chain link.

IBC6 (1)

Next, insert the address to `Profile1`, the `Profile1` statement will be like:

```
Profile1 {
  ...
  Addresses: [ "Desmos1", "Desmos2" ]
}
```

Then, actions about CRUD will be like:

### Get whole related addresses from `Profile1`:

1. List all `inter-link` addresses from profile
2. Traversal all links related the addresses in the list
3. Put `cross-chain` addresses to the list

It will be O(n) complexity, assume if n is small then it is enough in the case.

### Create a `inter-link` to `Profile1`

1. Check if the account `Desmos4` has not linked to other profile
2. Push it to the profile addresses storage

It will be O(1) complexity.

### Remove a `inter-link`

1. Remove the certain address from `Profile1`

It will be also O(n) complexity because the addresses storage in `Profile` is array. However, we don't need to check each addresses links which is related to the `removed address`.

<hr>

With this structure, the `cross-link` will be completely isolated from profile. However, it makes the system complicated as you mentioned indeedly. As the result, I think the advantages overweigh disadvantages.

RiccardoM:
@dadamu I see what you mean. What I was thinking, though, was an even simpler implementation of both cross-chain and inter-chain links that require exactly 1 Desmos address to be associated with a Profile in order to create a valid link. Such implementation can be easily summarized by the following diagrams. 


#### Inter-chain connection flow
Untitled Diagram(1)

#### Cross-chain connection flow
Untitled Diagram(2)

--- 

This solution should make it very easy to implement all the checks and make sure that only a Profile can be linked to other addresses as well as there are no profiles linked together. 


The links can then be stored inside the `Profile` structure as an array of `Link` objects. We need such array because we might also be interested in knowing when a link was created and the proof associated with it. 

Finally, if we want to make this system easily usable by other chains what we can do is create a new module called `x/ibc/profiles` that contains only the definition of the IBC packets and methods that other chains can use to send over their request. In this way we won't require them to create an `x/profiles` keeper but we can still handle all those IBC packets inside the `x/profiles` module which would make it easier for us.

dadamu:
@RiccardoM It is awesome design to prevent from all what I concerned and provide the scalability for other chains who would like to create a link to profile on Desmos safely.

The last small question. 
What does the `Proof` you mentioned in the `Link` struct look like? 
In my current way, I use the signature which binary link signed by private keys as a Proof.

RiccardoM:
> What does the `Proof` you mentioned in the `Link` struct look like?
> In my current way, I use the signature which binary link signed by private keys as a Proof.

@dadamu I think that the overall structure of the `Link` object might be something like this one: 

```golang
// Link contains the data representing either an inter- or cross- chain link
type Link struct {
  AddressHex string       // Hex address of the account inside the other chain
  ChainConfig ChainConfig // Configuration of the other chain
  CreationTime time.Time  // Time in which the link was created 
  Proof Proof             // Signature proof
}

// ChainConfig contains the data of the chain on the other end of the link. 
// This will be used to get useful information about the link (eg. the Bech32 address instead of the HEX one)
type ChainConfig struct {
  ID string              // ID of the other chain
  Bech32AddrPrefx string // Bech32 prefix to be used for the address
}

// Proof contains all the data used to verify the signature when linking an account to a profile
type Proof struct {
  PubKeyHex string    // Hex representation of the public key used to sign the link proof
  SignatureHex string // Hex representation of the signed data
}
```

The `SignatureHex` should be the HEX-encoded signature of the `Adress` that is placed inside the `Link` object. 

#### Example forkflow
Suppose we want to connect a `cosmos` account to a `desmos` account. Then what the user needs to do on the Cosmos chain is: 

1. Get their HEX address
2. Sign the HEX address using the private key associated with their Cosmos account
3. Send the following data over:   
   - HEX address
   - HEX public key
   - Signed HEX address
   - Chain ID 
   - Bech32 address prefix

Then we will need to verify the following constraints are matched:

1. The signature is valid using the given public key 
2. The given public key is associated with the provided address

dadamu:
@RiccardoM Thanks for answering my question. I agree with the structure of `Link`.

I have another idea about the linking structure.  Instead of checking if desmos account associated to profile or not, is that changing to check if desmos account has own its profile good idea?

The linking structure will be like:

IBC8

Instead of:

IBC7

RiccardoM:
@dadamu Yes, that's perfectly fine and it's the best way to go in my opinion. 

Since a `Profile` is just an implementation of the Cosmos' `AccountI` interface, you can also check if a Desmos account has its own profile as such: 

```golang
account := accountkeeper.GetAccount(address)
profile, ok := account.(*types.Profile) 
if !ok {
  // The Desmos address does not have a profile
}

// The Desmos address has a profile 
```

dadamu:
@RiccardoM That's great. I will implement them.

- [ ] Migrate `x/links` to `x/ibc/profile`
- [ ] Reconstruct `Link` and `Profile` object in `x/profile`
- [ ] Implement `Link` storage and the logic associated `Profile`
- [ ] Write unit tests

Thanks for your advice @RiccardoM @bragaz.

dadamu:
@RiccardoM Hi, I have some question when implementing `Link` struct.

Why the encode type of `Address` is Hex instead of Bech32?
If using hex encoded address, creating links from different chains with one keypair may not easy to be checked if it duplicated.

It is `Link` object struct in my imagination.
```
type Link struct {
  SourceAddress string       // Bech32 address of the account inside the other chain
  DestinationAddress string  // Bech32 address of the account inside Desmos chain
  Proof Proof                // Signature proof
  ChainConfig ChainConfig    // Configuration of the other chain
  CreationTime time.Time     // Time in which the link was created 
}
```

RiccardoM:
@dadamu If you think it would be better, I guess we can use Bech32 without any problem. I used the hex representation to make it chain-independent, but if we provide the `ChainConfig` object then it should be the same to use the Bech32 format instead. 

Also, I would avoid putting the `SourceAddress` field inside a `Link` object. Since links will be present inside a `Profile`, it would be redundant to put a `SourceAddress` inside each link. This value would be identical to the profile address, and so it can be easily removed.

dadamu:
@RiccardoM I got it, the cross-chain and inter-chain links are associated with Profile directly, not account. So desmos address is not needed in `Link` object. Thanks!

RiccardoM:
Hey @dadamu, since this PR got quite big (more than 100 files to be checked), I think we should split it into multiple ones, each one with its review process.

To do this, what I would ask you to do is the following: 

1. Create a new branch named `paul/ibc-profiles` starting from the current `master` branch. 
2. Using that branch, create multiple consecutive PRs each one implementing one small change: 
   1. The addition of Proto files
   2. The logic behind storing the links
   3. Tests
   4. Documentation
   
At each PR me and @bragaz will review it and once everything is done, merge it into the destination branch. 

The overall process will be the following: 

1. Fork `master` and create `paul/ibc-profiles`. 
2. Put inside the `paul/ibc-profiles` the updated Proto files. 
3. Open a new PR and ask me and @bragaz the review. 
4. Once the review is approved and merged, add to the `paul/ibc-links` branch the logic to store the links.
5. Open a new PR and ask me and @bragaz the review.
6. Once the review is approved and merged, add to the `paul/ibc-links` branch the tests. 
7. Open a new PR and ask me and @bragaz the review. 
8. Once the review is approved and merged, add to the `paul/ibc-links` branch the documentation. 
9. Open a new PR and ask me and @bragaz the review. 

Ideally, all the PRs should be made towards a new branch called `paul/ibc-profiles` that will have all the changes. Finally, we will create a `paul/ibc-profiles -> master` PR and merge everything when it's all clear and finalized.

This way we will be able to review all the things faster and give you a more rapid feedback. 

If all of this is clear, please start with creating the branches and close this PR.

PR #420, opened 2021-05-13
Leonardo/subspaces

### LINKED ISSUE #392: Implement manageable subspaces
## Feature description 
Following the [_"Extend the concept of subspace"_ discussion](https://github.com/desmos-labs/desmos/discussions/375_), we should implement editable subspaces.

## Implementation proposal
We would like to implement a new module called `x/subspaces` inside which every subspace is represented by an entity that has the following attributes: 

- and `id`, which is a unique SHA-256 hash that can be used to reference such subspace while creating posts;
- an `owner` which is going to be the profile that first registers that subspace; 
- an optional list of `admins` that represent the users allowed to perform special operations inside such subspace.

I think that what we should do is the same that Band Protocol does with data sources. Citing their documentation website: 

> When registering the data source, the message sender can choose whether to specify an owner of the source. If an owner is specified, only the owner can make any changes to the data source once it is registered

This should be the same behavior that we should implement with subspaces. The user that registers a subspace should be allowed to specify a different owner of such subspace. If they do not specify an owner, the subspace will not be editable later. Otherwise, only the owner will be able to edit it. 

### Admins
Since we want the subspace to be manageable, we should allow the owner to add or remove admins to the subspace. 

This should be done using two different messages that can be sent **only by the creator of the subspace**: `MsgAddAdmin` and `MsgRemoveAdmin`. 

Each admin should be able to perform some administering operations. These initially can be limited to: 

- allowing a specific user to use the subspace (`MsgAllowUser`); 
- removing the ability for a specific user to use the subspace (`MsgRemoveUser`).

#### Notes
We should edit the existing `posts` and `reports` modules to make sure that users who are blocked from using a subspace are not able to:

- create, comment or react to posts present inside such subspace, and
- report posts present inside such subspace

bragaz:
## Description
This PR introduce the expansion of the original `subspace` field we used inside `x/posts`.
Now the whole subspaces are managed inside `x/subspaces` module.
Closes #392.
## Checklist
- [x] Wrote unit tests.


bragaz:
@RiccardoM I think this is a good point where to start a first review before testing.
Other stuff that I want to introduce later but not it this PR will be:
1. Adapt other modules to the subspaces one.
2. Add the support for IBC

proto/desmos/subspaces/v1beta1/genesis.proto
```diff
@@ -0,0 +1,34 @@
...
+
+// GenesisState contains the data of the genesis state for the subspaces module
+message GenesisState {
+  repeated desmos.subspaces.v1beta1.Subspace subspaces = 1
+        [ (gogoproto.nullable) = false ];
+  repeated SubspaceAdminsEntry subspace_admins = 2
+        [ (gogoproto.nullable) = false ];
+  repeated BlockedUsersEntry blocked_to_post_users = 3
+        [ (gogoproto.nullable) = false ];
+}
```
RiccardoM, line 18: Since the admins and allowed users won't change that much, can we have those data directly inside a subspace? I think it would be much easier to handle as well as better for anyone querying a subspace to get all the data directly. 

This would make the genesis state a lot simpler: 

```protobuf
message GenesisState {
  repeated Subspaces subspaces = 1
}
```

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,153 @@
...
+option go_package = "github.com/desmos-labs/desmos/x/staging/subspaces/types";
+
+// Msg defines subspaces Msg service.
+service Msg {
+
+  // CreateSubspace defines the method to create a subspace
+  rpc CreateSubspace(MsgCreateSubspace) returns (MsgCreateSubspaceResponse);
+
+  // TransferOwnership defines the method to transfer a subspace ownership
+  rpc TransferOwnership(MsgTransferOwnership) returns (MsgTransferOwnershipResponse);
```
RiccardoM, line 15: Instead of something like `MsgTransferOwnership`, I think it would be much better to implement a more genesis `MsgEditSubspace` that allows to edit the different things (except the `id`). This would also allow to change the name and the owner with a single message. We could then use the usual `[do-not-modify]` value to mean something shouldn't be edited, the same way we do with profiles.

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,153 @@
...
+  rpc AddAdmin(MsgAddAdmin) returns (MsgAddAdminResponse);
+
+  // RemoveAdmin defines the method to remove an admin from a specific subspace
+  rpc RemoveAdmin(MsgRemoveAdmin) returns (MsgRemoveAdminResponse);
+
+  // EnableUserPosts defines the method to let user posts inside a specific subspace
+  rpc EnableUserPosts(MsgEnableUserPosts) returns (MsgEnableUserPostsResponse);
+
+  // DisableUserPosts defines the method to block user posts inside a specific subspace
+  rpc DisableUserPosts(MsgDisableUserPosts) returns (MsgDisableUserPostsResponse);
```
RiccardoM, line 27: Instead of `MsgEnableUserPosts` and `MsgDisableUserPosts` I think it would be much clearer if we used more widely-used terms such as `MsgRegisterUser` and `MsgBlockUser` that might make it easier for external developers to immediately understand what those are used for

proto/desmos/subspaces/v1beta1/subspace.proto
```diff
@@ -0,0 +1,31 @@
...
+
+  string id = 1 [
+    (gogoproto.customname) = "ID",
+    (gogoproto.jsontag) = "subspace_id",
+    (gogoproto.moretags) = "yaml:\"subspace_id\""
+  ];
+
+  string name = 2 [ (gogoproto.moretags) = "yaml:\"name\"" ];
+
+  string owner = 3 [ (gogoproto.moretags) = "yaml:\"owner\""];
```
RiccardoM, line 22: Watching this, I think we should have another field named `creator` that allows to tell who was the creator of the subspace. We should also leave the `owner` as that might change over time while the creator stays the same. This would also serve the purpose to always identify a user that has created a subspace even when they do not specify an owner (if they want to make it immutable).

proto/desmos/subspaces/v1beta1/subspace.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+package desmos.subspaces.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/desmos-labs/desmos/x/staging/subspaces/types";
+
+// Subspace contains all the data of a Desmos subspace
+message Subspace {
```
RiccardoM, line 10: I think we should add another field to the subspace named like `open` or something similar, that allows to specify whether posting inside such subspace is allowed without any registration. This would allow users to create open communities that do not require subspace owners to manually approve people to post inside them

bragaz: wouldn't be simpler if we let subspaces opened by default and then let admin/owner to block someone if they're not respecting the rules?

RiccardoM: @bragaz No, I think it's better to have subspaces closed by default, optionally open to everyone, and allow admins to both: 

- register new users
- block existing users

This would be an implementation that's more similar to current social networks, where by default you are not registered and you have to go through a custom procedure to be allowed to use them. If we set subspaces open by default, developers would not be able to do this cause they would have to block all existing users and then unblock them later. It's much easier if we: 

1. implement an `Open` field that tells that subspace does not require a registration 
2. allow subspace admins to allow individual Desmos accounts to post inside it

bragaz: Ok gotcha!

bragaz:
@RiccardoM besides tests I should've fixed everything as we discussed. 
Let me know what do you think!

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+  rpc AddAdmin(MsgAddAdmin) returns (MsgAddAdminResponse);
+
+  // RemoveAdmin defines the method to remove an admin from a specific subspace
+  rpc RemoveAdmin(MsgRemoveAdmin) returns (MsgRemoveAdminResponse);
+
+  // RegisterUser defines the method to let user posts inside a specific subspace
+  rpc RegisterUser(MsgRegisterUser) returns (MsgRegisterUserResponse);
+
+  // BlockUser defines the method to block user posts inside a specific subspace
+  rpc BlockUser(MsgBlockUser) returns (MsgBlockUserResponse);
```
RiccardoM, line 27: Since we're having `RegisterUser` and `BlockUser`, I think we should also have `UnregisterUser` and `UnblockUser` here. So that social networks can implement those functions as well.

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+
+  string new_owner = 2 [
+    (gogoproto.jsontag) = "new_owner",
+    (gogoproto.moretags) = "yaml:\"new_owner\""
+  ];
+
+  string new_name = 3 [
+    (gogoproto.jsontag) = "new_name",
+    (gogoproto.moretags) = "yaml:\"new_name\""
+  ];
```
RiccardoM, line 76: I think there's no need to use the `new_` prefix here. Simply using `owner` and `name` is sufficient IMO

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+
+// ___________________________________________________________________________________________________________________
+
+// MsgEditSubspace represents the message used to edit a subspace fields
+message MsgEditSubspace {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string id = 1 [
+    (gogoproto.customname) = "SubspaceID",
```
RiccardoM, line 63: I think this can be simply `ID` instead of `SubspaceID`. It should be clear, since we're editing a subspace, the id refers to a subspace

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+    (gogoproto.moretags) = "yaml:\"subspace_id\""
+  ];
+
+  string name = 2 [
+    (gogoproto.moretags) = "yaml:\"name\""
+  ];
+
+  string creator = 3 [ (gogoproto.moretags) = "yaml:\"creator\"" ];
+
+  bool open = 4 [ (gogoproto.moretags) = "yaml:\"open\""];
```
RiccardoM, line 49: Can we make `open` the third field and `creator` the last one?

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string subspace_id = 1 [
+    (gogoproto.customname) = "SubspaceID",
+    (gogoproto.jsontag) = "subspace_id",
+    (gogoproto.moretags) = "yaml:\"subspace_id\""
+  ];
+
+  string new_admin = 2 [(gogoproto.moretags) = "yaml:\"new_admin\""];
```
RiccardoM, line 97: I think this can simply be `admin` without the `new_` prefix.

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string user = 1 [ (gogoproto.moretags) = "yaml:\"user\"" ];
+  string subspace_id = 2 [
+    (gogoproto.customname) = "SubspaceID",
+    (gogoproto.jsontag) = "subspace_id",
+    (gogoproto.moretags) = "yaml:\"subspace_id\""
+  ];
+  string admin = 3 [ (gogoproto.moretags) = "yaml:\"admin\"" ];
```
RiccardoM, line 137: I think this should better be `user` instead of `admin`

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,160 @@
...
```
RiccardoM, line 156: I think this should better be `user` instead of `admin`

x/staging/subspaces/client/cli/query.go
```diff
@@ -0,0 +1,55 @@
...
+		GetCmdQuerySubspace(),
+	)
+	return subspaceQuerycmd
+}
+
+// GetCmdQuerySubspace returns the command to query the subspace with the given id
+func GetCmdQuerySubspace() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "subspace [id]",
+		Short: "Retrieve the subspace with the given [id] alongside its admins and the users that cant post in it",
```
RiccardoM, line 31: No need to specify all the returned data, as it might change in the future and we don't want to keep updating the description
```suggestion
		Short: "Get the subspace with the given [id]",
```

proto/desmos/subspaces/v1beta1/query.proto
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// ___________________________________________________________________________________________________________________
+
+// QuerySubspace is the request type for the Query/Subspace RPC method
+message QuerySubspaceRequest {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string subspace_id = 1;
```
RiccardoM, line 27: We should use a custom Go name making this `SubspaceID` instead of `SubspaceId`

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return subspacesTxCmd
+}
+
+// GetCmdCreateSubspace returns the command used to create a subspace
+func GetCmdCreateSubspace() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [subspace-id] [name] [[open]]",
+		Args:  cobra.RangeArgs(2, 3),
+		Short: "Create a subspace with the given [subspace-id], [name] and [open]",
```
RiccardoM, line 44: No need to specify all the flags
```suggestion
		Short: "Create a new subspace",
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+		Long: fmt.Sprintf(`The [subspace-id] must be a sha256 string identifying the subspace, 
+the [name] a human readable name and the [open] a boolean identifying whether the subspace allow users posts without registration.
+Leaving [open] empty will set it to false by default. 
+e.g 1) %s tx subspaces create 4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e "mooncake"
+	2) %s tx subspaces create 4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e "mooncake" true
```
RiccardoM, line 49:
```suggestion
		Long: fmt.Sprintf(`Create a new subspace.
		
The id must be a valid SHA-256 hash uniquely identifying the subspace. 
The name shall be a human readable name, while the open can be used to tell whether the subspace allow users to post messages freely.

Examples:
%s tx subspaces create 4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e "mooncake"
%s tx subspaces create 4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e "mooncake" true
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+		GetCmdEditSubspace(),
+	)
+
+	return subspacesTxCmd
+}
+
+// GetCmdCreateSubspace returns the command used to create a subspace
+func GetCmdCreateSubspace() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [subspace-id] [name] [[open]]",
```
RiccardoM, line 42: Instead of making `open` an argument, I think it would be better to have it as a flag since it's optional with a default value of `true`: 

```
create <hash> <name>
create <hash> <name> --open
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func GetCmdAddSubspaceAdmin() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-admin [address] [subspace-id]",
+		Args:  cobra.ExactArgs(2),
+		Short: "Add a new admin to the subspace with the given [subspace-id]",
```
RiccardoM, line 84:
```suggestion
		Short: "Add a new admin to the subspace with the given id",
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func GetCmdRegisterUser() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "register-user [address] [subspace-id]",
+		Args:  cobra.ExactArgs(2),
+		Short: "Register a user inside the subspace with the given [subspace-id] to let him post in it",
```
RiccardoM, line 148: No need to tell what registration does, since it might change in the future and we don't want to have to update this doc 
```suggestion
		Short: "Register a user inside the subspace with the given id",
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func GetCmdBlockUser() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "block-user [address] [subspace-id]",
+		Args:  cobra.ExactArgs(2),
+		Short: "Block a user to post inside the subspace with the given [subspace-id]",
```
RiccardoM, line 177:
```suggestion
		Short: "Block a user to post inside the subspace with the given id",
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func GetCmdEditSubspace() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "edit [subspace-id] [[new_owner]]",
```
RiccardoM, line 204: Instead of making the `owner` an argument, I would make this a flag with the default value of an empty string (`""`). This would require the user to use it everytime they want to edit a subspace, but I think it would be better than making it an argument.

x/staging/subspaces/client/cli/keys.go
```diff
@@ -0,0 +1,6 @@
+package cli
+
+// Subspaces flag
+const (
+	FlagNewName = "new-name"
```
RiccardoM, line 5: I think we should have simpler flag name such as `--name` instead of `--new-name`. Also, is it necessary to put the flags here? Can't we declare them in the same place where the command is used?

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+		Args:  cobra.RangeArgs(1, 2),
+		Short: "Edit an existent subspace with the given [subspace-id]",
```
RiccardoM, line 206:
```suggestion
		Short: "Edit an existent subspace with the given id",
```

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,249 @@
...
+		Long: fmt.Sprintf(`Edit a subspace with the given [subspace-id].
```
RiccardoM, line 207:
```suggestion
		Long: fmt.Sprintf(`Edit a subspace with the given id.
```

proto/desmos/subspaces/v1beta1/msgs.proto
bragaz: mmmh, the fact is that we've already the `owner` field identifying the actual owner who's doing the message

RiccardoM: We can use `editor` to identify the user that's editing the subspace. In this way we can use `owner` to identify the subspace owner

x/staging/subspaces/client/cli/tx.go
RiccardoM: I revised my stance. I think the best way is to set the default value of a flags to `[do-not-modify]` instead to be more consistent with other edit messages (eg. `MsgSaveProfile` or others)

proto/desmos/subspaces/v1beta1/msgs.proto
bragaz: Since it's the `admin` that's registering the user I don't think so. 
Also we will have two `user` fields which I think it's not possible.

proto/desmos/subspaces/v1beta1/msgs.proto
bragaz: same as above. Maybe we should name the `user` field `address` and `admin` -> `user`?

x/staging/subspaces/client/cli/keys.go
bragaz: we declared them like this inside the post's cli commands, I was following the same pattern

proto/desmos/subspaces/v1beta1/msgs.proto
RiccardoM: @bragaz Didn't see we already have a `user` field. My bad

x/staging/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,306 @@
...
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			subspaceID := args[0]
+			subspaceName := args[1]
+			open := viper.GetBool(FlagOpen)
+
+			msg := types.NewMsgCreateSubspace(subspaceID, subspaceName, clientCtx.FromAddress.String(), open)
```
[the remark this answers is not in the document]
bragaz: I think it's better to create the message and use its `ValidateBasic()` method handles the checks like the one you've mentioned. If the user send some wrong fields he will eventually be notified by a clear print statement like this one
```bash
Error: message validation failed: subspace id must be a valid sha-256 hash: invalid subspace [desmos-labs/desmos/x/staging/subspaces/types/msgs.go:36]
```

proto/desmos/subspaces/v1beta1/models.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+package desmos.subspaces.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/desmos-labs/desmos/x/staging/subspaces/types";
+
+// Users represents a slice of users inside a specific subspace
+message Users {
```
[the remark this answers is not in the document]
bragaz: Yes I think so. I've added this type when I originally intend the `Admins` and `RegisteredUsers` as separated entities from the subspace and needed it because Protobuf won't let you marshal something that doesn't implement its `marshaler` methods.

proto/desmos/subspaces/v1beta1/query.proto
```diff
@@ -0,0 +1,37 @@
...
+
+import "gogoproto/gogo.proto";
+import "google/api/annotations.proto";
+import "desmos/subspaces/v1beta1/subspace.proto";
+import "desmos/subspaces/v1beta1/models.proto";
+
+option go_package = "github.com/desmos-labs/desmos/x/staging/subspaces/types";
+
+// Query defines the gRPC querier service
+service Query {
```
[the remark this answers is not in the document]
bragaz: 👍 I will add it ASAP

x/staging/subspaces/keeper/common_functions.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	return subspace, nil
+}
+
+// CheckSubspaceAndOwner check if the subspaces with the given id exists and
+// if the address belongs to the owner of the subspace.
+// It returns an error or the subspace itself if everything's fine.
+func (k Keeper) CheckSubspaceAndOwner(subspaceBytes []byte, subspaceID, address string) (types.Subspace, error) {
```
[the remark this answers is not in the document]
bragaz: 4. I can't test them if I don't make them public unfortunately :/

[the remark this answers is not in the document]
RiccardoM: @bragaz I think that those methods are simple enough we don't need to test them independently. We can simply rely on the tests of the methods that used them (i.e `AddAdmin` etc), so it's fine to keep them private and untested

x/staging/subspaces/client/cli/tx.go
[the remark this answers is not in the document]
RiccardoM: @bragaz Ok seems fine then :+1:

PR #428, opened 2021-05-17
Small fixes

### LINKED ISSUE #423: Rename all `Attribute` into `AttributeKey`
Currently inside the different modules, when emitting the different events we are using some attribute keys that have name starting with `Attribute` while other times (eg. inside `x/posts`) we are using the `AttributeKey` prefix. 

Example:
- `AttributeKeyPostID` 
- `AttributeRequestReceiver`

We should change all the ones missing the `Key` part to include it. So `AttributeRequestReceiver` becomes `AttributeKeyRequestReceiver`.

### LINKED ISSUE #424: Improve `DTag` regex docs
Currently inside the documentation there's a wrong `DTag` regex (see [here](https://docs.desmos.network/developers/types/profiles/profile.html#dtag)) that's different from the chain one.

### LINKED ISSUE #425: Improve `DTag` regex
Currently inside the chain (and the default params of the `x/profiles` module) we are using a RegEx that does **not** end with `$`. I think we should add this since it would make for a better control of the value overall.

bragaz:
## Description
This PR solves small stuff like the attribute renaming, the update regEx for docs.
Closes #423.
Closes #424.
Closes #425.
## Checklist
- [ ] Wrote unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 - Removed `--allows-comments` in favor of `--disable-comments` ([#386](https://github.com/desmos-labs/desmos/issues/386))
 - Standardized `PostID` naming convetions ([#419](https://github.com/desmos-labs/desmos/issues/419))
 - Renamed all events `Attribute` into `AttributeKey` ([#423](https://github.com/desmos-labs/desmos/issues/423))
-- Updated `DTag` regEx docs ([#424](https://github.com/desmos-labs/desmos/issues/424))
 
 # Version 0.16.2
 ## Changes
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - Capped the `OptionalDataEntry.Key` length ([#393](https://github.com/desmos-labs/desmos/issues/393)])
 - Removed `--allows-comments` in favor of `--disable-comments` ([#386](https://github.com/desmos-labs/desmos/issues/386))
 - Standardized `PostID` naming convetions ([#419](https://github.com/desmos-labs/desmos/issues/419))
+- Renamed all events `Attribute` into `AttributeKey` ([#423](https://github.com/desmos-labs/desmos/issues/423))
+- Updated `DTag` regEx docs ([#424](https://github.com/desmos-labs/desmos/issues/424))
```
RiccardoM, line 7: No need to mention docs changes
```suggestion
```

RiccardoM (approved):

PR #435, opened 2021-05-24
Implement Post OptionalData improvements

### LINKED ISSUE #422: Post `OptionalData` improvements
Currently inside the `x/posts` module we have defined two types to handle the optional data that's associated with a post: 

- `OptionalDataEntry` that represents a single optional data, and
- `OptionalData` that represents a slice of optional data

In my opinion, these very similar names might be confusing for external developers. Also, I think the current naming of `OptionalData` might be confusing as well. For this reason, I propose the following changes:

1. Remove the `OptionalData` type and use only the `[]OptionalDataEntry`
2. Rename the `OptionalDataEntry` type to be `Attribute` 
3. Rename the `OptionalData` field of a `Post` into `AdditionalAttributes`

This way, the final `Post` object would then be something like

```go
type Post struct {
  AdditionalAttributes []Attribute
}
```

IMO, this would make it very easy for external developers to understand that they can insert additional things inside that array.

dadamu:
## Description

This PR is the implementation of #422, the features include:
1. Remove the `OptionalData` type and use only the `[]OptionalDataEntry`
2. Rename the `OptionalDataEntry` type to be Attribute
3. Rename the `OptionalData` field of a Post into `AdditionalAttributes`
4. Rename all `OptionalData` related fields of the Params into `AdditionalAttributes`

Closes #422

## Checklist
- [ ] Wrote unit tests.


RiccardoM (approved):
Looks great to me! :100:

x/staging/posts/types/posts.go
```diff
@@ -13,21 +13,21 @@ import (
 // NewPost allows to build a new Post instance with the provided data
 func NewPost(
 	postID string, parentID string, message string, disableComments bool, subspace string,
-	optionalData OptionalData, attachments []Attachment, pollData *PollData,
+	additionAttributes []Attribute, attachments []Attachment, pollData *PollData,
```
bragaz, line 16:
```suggestion
	additionalAttributes []Attribute, attachments []Attachment, pollData *PollData,
```

bragaz (changes requested):
small typo

bragaz (approved):

PR #436, opened 2021-05-25
Built proto files of IBC profiles

dadamu:
## Description

This PR is proto files of the implementation related to #192, it includes:

* Build `x/ibc/profiles` basic proto files ( genesis, msg, query )
* Conduct `x/ibc/profiles` packet.proto for communicating between two chains
* Add `Link` related struct to `x/profiles` models.proto
* Define `Link` and `Unlink` functions to `x/profiles` msg.proto

## Checklist
- [ ] Wrote unit tests.

### DIFF
--- a/proto/desmos/profiles/v1beta1/models.proto
+++ b/proto/desmos/profiles/v1beta1/models.proto
@@ -119,8 +119,8 @@ message Proof {
   // signature represents the hex string produced from value by the private key
   string signature = 2 [ (gogoproto.moretags) = "yaml:\"signature\"" ];
 
-  // value represents the plain text in order to produce the signature
-  string value = 3 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+  // plain text represents the value in order to produce the signature
+  string plain_text = 3 [ (gogoproto.moretags) = "yaml:\"plain_text\"" ];
 }
 
 // ChainConfig contains the data of the chain with which the link is made.


RiccardoM (commented):
I'm having some hard time understanding the following things: 

1. Why are there two messages (`MsgCreateIBCAccountConnection` and `MsgCreateIBCAccountLink`) with a very similar name that contain pretty much the same data?
2. Why are there two different packet data (`IBCAccountConnectionPacketData` and `IBCAccountLinkPacketData`) that contain pretty much the same data?
3. I'm seeing a `PortID` being added in both the `x/profiles` genesis state and the `x/ibc/profiles` genesis state. Is this really necessary? Can't we simply add it to the `x/profiles` one? 


Also, lately I've been thinking about how to organize the different files in the appropriate folders, and I've came to the conclusion that it's probably best to put all the profiles-related methods inside the `x/profiles` module. 

Another chain that likes to integrate Desmos IBC profiles will simply have to import the `x/profiles/types` folder. There is no need for the other chain to create a keeper or do some fancy stuff. All they have to do is create the packet data that we define and then perform a simple IBC transaction with it inside. So we can avoid complicating out code organization and simply put everything inside the `x/profiles` module. I would simply make sure that IBC-related files are identifiable by an `ibc_` prefix on the file name itself (eg. the IBC-related keeper methods should be put inside `ibc_keeper.go` and the IBC connections should be inside `ibc_connections_keeper.go`).

dadamu:
@RiccardoM  Thanks for your reply.

For answering the first two questions, I followed the packet types the following issue mentioned. 
https://github.com/desmos-labs/desmos/issues/192#issuecomment-642521980
It is easy to implement but occurs duplicated data indeed. They are used for two different cases to create the link, so using two different types makes sense to me, too. If combining them to one type,  then the handler function should contains two logic to deal with it.

About the third question. If other chain would like to integrate the IBC profiles feature, besides importing the packet and msg type, they still need to create a keeper to handle the IBC things like port and channel. So my idea is to maintain `x/ibc/profiles` ourselves to communicate with Desmos `x/profiles` for creating links.

RiccardoM:
> For answering the first two questions, I followed the packet types the following issue mentioned.
> [#192 (comment)](https://github.com/desmos-labs/desmos/issues/192#issuecomment-642521980)
> It is easy to implement but occurs duplicated data indeed. They are used for two different cases to create the link, so using two different types makes sense to me, too. If combining them to one type, then the handler function should contains two logic to deal with it.

I've re-read the comment and all the discussion linked to it. Honestly, I think that we can simplify everything by simply ditching the second case and leaving only a generic packet that assumes the private keys are always different. This because Desmos has a derivation path that is different from other chains (`m'/852'/0'/0/0`) which results in different private keys being generated from the same mnemonic. For this reason, the case in which the private keys will be the same will never happen (in this case it means we are trying to link two Desmos accounts, which do not require IBC to be done). 

For this reason, I think we can create a single `AccountLinkPacketData` structure: 

```proto
message AccountLinkPacketData {
  SrcAddress string
  SrcProof Proof
  SrcChainConfig ChainConfig
  DestAddress string
  DestProof Proof
}
```

This re-uses already defined types and makes it simpler for developers to implement the logic.

> About the third question. If other chain would like to integrate the IBC profiles feature, besides importing the packet and msg type, they still need to create a keeper to handle the IBC things like port and channel. So my idea is to maintain `x/ibc/profiles` ourselves to communicate with Desmos `x/profiles` for creating links.

You're right, if an external chain wants to integrate with Desmos they need to create a keeper to perform IBC-related operations. However, we should not care about this detail from our point of view. This is because if someone is integrating with us they most likely are going to define custom a logic on top of our common IBC implementations. 

Let's consider for example the case in which Bitsong wants to integrate with Desmos. There are two things that might happen: 

1. Bitsong only wants to provide its users the ability to link their Bitsong account with their Desmos account. In this case, they might want to keep track of the various connections as well, which requires a custom keeper. 
2. Bitsong wants to provide its users the ability to perform other Desmos-related operations through IBC. In this case, they will surely need a custom keeper to implement additional checks and also send posts across IBC. 

In both cases, I think it's not necessary for us to give them a pre-built solution. Most zone developers will still implement their custom logic to perform additional checks and other things. For this reason, I think it's much more worth for us to have a cleaner code structure rather than a pre-built keeper that the other chains can use.

dadamu:
@RiccardoM 
Great, I agree with both of them.
We can provide a doc to explain how to integrate and create a custom keeper with ibc-profiles instead of maintain `x/ibc/profiles`
Just I also use `x/ibc/profiles` for testing ibc features, maybe It is needed to have a test msg to send packet in `x/profiles`.

Let me reconstructed the proto files.

proto/desmos/profiles/v1beta1/genesis.proto
```diff
@@ -30,4 +30,12 @@ message GenesisState {
...
     (gogoproto.moretags) = "yaml:\"params\""
   ];
+
+  string port_id = 5 [
+    (gogoproto.moretags) = "yaml:\"port_id\"",
+    (gogoproto.customname) = "PortID"
+  ];
+
+  repeated desmos.profiles.v1beta1.Link links = 6
+      [ (gogoproto.moretags) = "yaml:\"links\"", (gogoproto.nullable) = false ];
```
RiccardoM, line 40: Since the links are also included inside each `Profile`, I don't think we need them here as well.

proto/desmos/profiles/v1beta1/genesis.proto
```diff
@@ -30,4 +30,12 @@ message GenesisState {
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"params\""
   ];
+
+  string port_id = 5 [
```
RiccardoM, line 34: Maybe we can rename this to be `ibc_port_id` to make it more clear this is IBC-related?

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -46,6 +47,17 @@ service Msg {
...
+  // CreateIBCAccountLink defines a method for creating an account link to other
+  // chain
+  rpc CreateIBCAccountLink(MsgCreateIBCAccountLink)
+      returns (MsgCreateIBCAccountLinkResponse);
+
+  // Link defines a method for linking an account
+  rpc Link(MsgLink) returns (LinkResponse);
+
+  // Unlink defines a method for unlink an account from a profile
+  rpc Unlink(MsgUnlink) returns (UnlinkResponse);
```
RiccardoM, line 60: I think it's better to rename the methods, messages and responses to be: 

- `LinkAccount`, `MsgLinkAccount`, `LinkAccountResponse`
- `UnlinkAccount`, `MsgUnlinkAccount` and `UnlinkAccountResponse` 

This is just to make it easier we are linking accounts and not other things (eg. third-party applications)

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -207,10 +219,67 @@ message BlockUserResponse {}
...
 }
 
 // UnblockUserResponse defines the Msg/UnblockUser response type.
 message UnblockUserResponse {}
+
+// ___________________________________________________________________________________________________________________
+
+// MsgIBCCreateAccountLink represents the message to be used to create a
+// link.
+message MsgCreateIBCAccountLink {
```
RiccardoM, line 234: What is this used for? Can this be removed?

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -207,10 +219,67 @@ message BlockUserResponse {}
...
+      [ (gogoproto.moretags) = "yaml:\"timeout_timestamp\"" ];
+}
+
+// MsgCreateIBCAccountLinkResponse defines the
+// Msg/CreateIBCAccountLink response type.
+message MsgCreateIBCAccountLinkResponse {}
+
+// ___________________________________________________________________________________________________________________
+
+// MsgLink represents a message to link an account.
```
bragaz, line 257:
```suggestion
// MsgLink represents a message to link an account to a profile.
```

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -207,10 +219,67 @@ message BlockUserResponse {}
...
+  string destination_signature = 4
+      [ (gogoproto.moretags) = "yaml:\"destination_signature\"" ];
+}
+
+// LinkResponse defines the Msg/Link response type.
+message LinkResponse {}
+
+// ___________________________________________________________________________________________________________________
+
+// MsgUnLink represents a message to unlink an account from a profile.
```
bragaz, line 274:
```suggestion
// MsgUnlink represents a message to unlink an account from a profile.
```

proto/desmos/profiles/v1beta1/msgs.proto
dadamu: I would like to use it to test the ibc packet with docker.

proto/desmos/profiles/v1beta1/genesis.proto
dadamu: It is for checking if link exists or not, used for preventing from double profiles problem.

proto/desmos/profiles/v1beta1/query.proto
```diff
@@ -44,6 +44,12 @@ service Query {
     option (google.api.http).get =
         "/desmos/relationships/v1beta1/blocks/{user}";
   }
+
+  // Link queries the user blocks for the user having the given address
```
bragaz, line 48:
```suggestion
  // Link queries the user links for the user having the given address
```

bragaz (changes requested):

proto/desmos/profiles/v1beta1/genesis.proto
RiccardoM: I think we should remove them. The double checking can still be done when initializing the data and storing them (eg. inside the keeper's method). Having a double list of links will result in duplicated code that needs to be written inside the genesis state

proto/desmos/profiles/v1beta1/msgs.proto
RiccardoM: IMO this should be removed. If we keep this message everyone will be able to use it, users included. For this reason, any test message that might be useful when performing automated tests should be declared as an internal class and defined inside the test classes themselves. We should not define something only because it might be useful inside tests

proto/desmos/profiles/v1beta1/genesis.proto
dadamu: Got it, will be store links here instead of using genesis file directly.
https://github.com/desmos-labs/desmos/blob/bd117008a64159184a57cdb9e43f986fec76c847/x/profiles/keeper/genesis.go#L27

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -207,10 +214,42 @@ message BlockUserResponse {}
...
+  string subspace = 4 [ (gogoproto.moretags) = "yaml:\"subspace\"" ];
 }
 
 // UnblockUserResponse defines the Msg/UnblockUser response type.
 message UnblockUserResponse {}
+
+// ___________________________________________________________________________________________________________________
+
+// MsgLinkAccount represents a message to link an account to a profile.
+message MsgLinkAccount {
```
RiccardoM, line 228: I was thinking again to this message, and I was wondering whether or not we should extend this to also allow users to link their other chain accounts with it. Maybe something like this: 

```proto
message MsgLinkAccount {
  string source_address = 1;
  Proof source_proof = 2;
  ChainConfig source_chain_config = 3;
  string destination_address = 4;
  Proof destination_proof = 5;
}
```

This should allow to connect another chain account (eg. `cosmos1...`) even if the other chain does not yet have implemented the IBC communication with Desmos. The handling will also be identical to the one of the IBC packet. 

What do you think @dadamu @bragaz?

proto/desmos/profiles/v1beta1/packets.proto
```diff
@@ -0,0 +1,29 @@
...
+  desmos.profiles.v1beta1.Proof source_proof = 2
+      [ (gogoproto.moretags) = "yaml:\"source_proof\"" ];
+
+  desmos.profiles.v1beta1.ChainConfig source_chain_config = 3
+      [ (gogoproto.moretags) = "yaml:\"source_chain_config\"" ];
+
+  string destination_address = 4
+      [ (gogoproto.moretags) = "yaml:\"destination_address\"" ];
+
+  string destination_proof = 5
```
RiccardoM, line 24: If I'm not mistaken, this should be a `Proof`, not a `string`

dadamu: Oh, My bad, let me check all things again.

proto/desmos/profiles/v1beta1/msgs.proto
dadamu: I think it is a good idea. Just concern that it seems that we should support different key base type on any other chains ( e.g. `Solana`, `Polkadot`... ) in Desmos cli tool.

bragaz: @RiccardoM I agree with this. @dadamu maybe we can add a field like `key_type` to face this problem....I was already thinking it before reading this comment. Thoughts?

bragaz (commented):
Everything looks good, just need to finish the discussion around `MsgLink`

proto/desmos/profiles/v1beta1/msgs.proto
RiccardoM: I don't get why we should add support for different keys. Aren't those chains using `Secp256k1` keys as well?

bragaz: I'm not sure (@dadamu may know this better) but I remember Solana has a different key type. Correct if I'm wrong

dadamu: As I know, Solana use `ed25519` keys. And it use ".json" key-file while cosmos-based chain use a key base folder to manage all keys.

RiccardoM: Thanks @dadamu and @bragaz. Knowing this, what I think we can do is handle public keys the same way that Cosmos does in order to handle public keys while signing the data: 

- use `Any` to represent a generic key inside Proto files ([here](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/tx/signing/v1beta1/signing.proto#L41))
- use the `PubKey` interface to represent a generic key inside Go files ([here](https://github.com/cosmos/cosmos-sdk/blob/master/crypto/types/types.go#L9))

By doing this we can leverage the Cosmos support of all keys types (`secp256k1`, `ed25516` and `secp256r1`) that is included inside the `crypto` package ([here](https://github.com/cosmos/cosmos-sdk/tree/master/proto/cosmos/crypto)) without having to duplicate our code.

dadamu: @RiccardoM It still can't know which signing algorithm is used when receiving a message, I think we can add a filed called algorithm to record it, and remain hex string on pub_key, like this:
message Proof{
  string pub_key = 1;
  string signature = 2;
  string algorithm = 3;
}

RiccardoM: @dadamu What do you mean with __signing algorithm__? Are other chains using signature algorithms that differ from the Cosmos one (`sign(sha256(message))`)?

dadamu: @RiccardoM like `secp256k1`, `ed25519 ` and `secp256r1`. or it is already included in the proto message?

RiccardoM: @dadamu There's no need to externally specify the key type. It will be already included inside the JSON encoding of the public key itself as you can see from the following test that I quickly wrote now: 

Proto file definition:
```proto
message Proof {
  google.protobuf.Any public_key = 1 [ (gogoproto.moretags) = "yaml:\"pub_key\"" ];
  string signature = 2 [ (gogoproto.moretags) = "yaml:\"signature\"" ];
}
```

Test code:
```golang
func TestProof_Serialization(t *testing.T) {
  sdk.GetConfig().SetBech32PrefixForAccount("desmos", "desmospub")
  pubKey, err := sdk.GetPubKeyFromBech32(sdk.Bech32PubKeyTypeAccPub, "desmospub1addwnpepq0m3wg0kzfsacgd4tkj8xr622yp9sztwmu22mz0rcu4ep2x5dex65g2v97y")
  require.NoError(t, err)

  any, err := codectypes.NewAnyWithValue(pubKey)
  require.NoError(t, err)

  cdc, _ := app.MakeCodecs()
  json := cdc.MustMarshalJSON(any)
  print(string(json))
}
```

Output:
```
{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A/cXIfYSYdwhtV2kcw9KUQJYCW7fFK2J48crkKjUbk2q"}
```

RiccardoM: I've also wrote a small test for future reference to remember how to deserialize that value:

Test code
```golang
func TestKey_Deserialization(t *testing.T) {
  json := `{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A/cXIfYSYdwhtV2kcw9KUQJYCW7fFK2J48crkKjUbk2q"}`

  cdc, _ := app.MakeCodecs()

  var any codectypes.Any
  cdc.MustUnmarshalJSON([]byte(json), &any)

  var pubkey cryptotypes.PubKey
  err := cdc.UnpackAny(&any, &pubkey)
  require.NoError(t, err)

  print(pubkey.String())
}
```

Output:
```
PubKeySecp256k1{03F71721F61261DC21B55DA4730F4A510258096EDF14AD89E3C72B90A8D46E4DAA}
```

dadamu: @RiccardoM I test it with ed25519 key. And it returns error, it shows that the pubkey of ed25519 doesn't implement proto.Message

Code:
```go
func main() {
	priv := ed25519.GenPrivKey()
	pubKey := priv.PubKey()

	any, _ := codectypes.NewAnyWithValue(pubKey)

	cdc, _ := simapp.MakeCodecs()
	json := cdc.MustMarshalJSON(any)
	fmt.Println(string(json))
}
```

Error:
```
cannot use pubKey (type "github.com/tendermint/tendermint/crypto".PubKey) as type "github.com/gogo/protobuf/proto".Message in argument to "github.com/cosmos/cosmos-sdk/codec/types".NewAnyWithValue:
        "github.com/tendermint/tendermint/crypto".PubKey does not implement "github.com/gogo/protobuf/proto".Message (missing ProtoMessage method)
```

dadamu: I wrote small test for `ed25519`, it works, too.

Code:
```go
func main() {
	priv := ed25519.GenPrivKey()
	pubKey := priv.PubKey()

	any, _ := codectypes.NewAnyWithValue(pubKey)

	cdc, _ := simapp.MakeCodecs()
	json := cdc.MustMarshalJSON(any)
	fmt.Println(string(json))
}
```
Output:
```
{"@type":"/cosmos.crypto.ed25519.PubKey","key":"S1s1TvGA+RLDJViBSO7kX1lhTXAbV3h7rMBvb8FI7DE="}
```

dadamu: @RiccardoM So, other non cosmos-based chains can also import the packet and create a proof whose public key includes type like `{"@type":"/cosmos.crypto.ed25519.PubKey","key":"S1s1TvGA+RLDJViBSO7kX1lhTXAbV3h7rMBvb8FI7DE="}`, then it can be deserialized the public key to verify the signature on Desmos, right?

RiccardoM: @dadamu That's correct

proto/desmos/profiles/v1beta1/genesis.proto
```diff
@@ -30,4 +30,9 @@ message GenesisState {
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"params\""
   ];
+
+  string ibc_port_id = 5 [
+    (gogoproto.moretags) = "yaml:\"ibc_port_id\"",
+    (gogoproto.customname) = "IBCPortID"
```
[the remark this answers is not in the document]
RiccardoM: @bragaz That's not correct. The `customname` is ignored only if used along with the `grpc-gateaway` Protobuf plugin, which is used only when generating the querier code. It can (and should) still be used elsewhere. Please @dadamu do not remove this line.

bragaz (approved):

RiccardoM:
While reviewing the naming used here I thought that we might want to slightly change some wordings. Since we're going to implement application links (eg. Twitter, GitHub, etc) with #427, I think we should make it very clear that IBC links are different from application links. For this reason, I think we can: 

- rename the `Link` object to `ChainLink` 
- rename the `LinkAccount` and `UnlinkAccount` server methods to `LinkChainAccount` and `UnlinkChainAccount` 
- rename the `MsgLinkAccount` and `MsgUnlinkAccount` messages to `LinkChainAccount` and `UnlinkChainAccount`

This way we will be able to implement with #427 the related `ApplicaitonLink` / `LinkApplication` / `MsgLinkApplication` names.

What do you think @bragaz @dadamu?

bragaz:
@RiccardoM Yeah it's better to be as specific as possible with naming, I agree!

dadamu:
@RiccardoM 
Yeah, I agree with it, too.

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -43,6 +43,8 @@ message Profile {
     (gogoproto.jsontag) = "creation_date",
     (gogoproto.nullable) = false
   ];
+  repeated ChainLink chain_links = 7
+      [ (gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"chain_links\"" ];
```
RiccardoM, line 47: I think we should name it using the plural form
```suggestion
  repeated ChainLink chains_links = 7
      [ (gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"chains_links\"" ];
```

proto/desmos/profiles/v1beta1/msgs.proto
```diff
@@ -46,6 +47,12 @@ service Msg {
 
   // UnblockUser defines a method for unblocking a user
   rpc UnblockUser(MsgUnblockUser) returns (UnblockUserResponse);
+
+  // LinkChainAccount defines a method to link an account to a profile
+  rpc LinkChainAccount(MsgLinkChainAccount) returns (LinkChainAccountResponse);
+
+  // UnlinkChainAccount defines a method to unlink an account from a profile
+  rpc UnlinkChainAccount(MsgUnlinkChainAccount) returns (UnlinkChainAccountResponse);
```
RiccardoM, line 55:
```suggestion
  // LinkChainAccount defines a method to link an external chain account to a profile
  rpc LinkChainAccount(MsgLinkChainAccount) returns (LinkChainAccountResponse);

  // UnlinkChainAccount defines a method to unlink an external chain account from a profile
  rpc UnlinkChainAccount(MsgUnlinkChainAccount) returns (UnlinkChainAccountResponse);
```

proto/desmos/profiles/v1beta1/packets.proto
```diff
@@ -0,0 +1,16 @@
...
+
+import "gogoproto/gogo.proto";
+import "desmos/profiles/v1beta1/msgs.proto";
+
+option go_package = "github.com/desmos-labs/desmos/x/profiles/types";
+
+// LinkChainAccountPacketData defines a struct for the packet payload
+message LinkChainAccountPacketData {
+  desmos.profiles.v1beta1.MsgLinkChainAccount msg = 1
+      [ (gogoproto.moretags) = "yaml:\"msg\"" ];
```
RiccardoM, line 12: I would vouch against including a message inside a packet data. Even thought it might seem counter-intuitive or a bad thing, I think we should simply copy/paste the same fields that `MsgLinkChainAccount` has inside the packet itself, in order to avoid having to create multiple objects just to send a packet: 
```suggestion
  string source_address = 1      [ (gogoproto.moretags) = "yaml:\"source_address\"" ];  
  desmos.profiles.v1beta1.Proof source_proof = 2      [ (gogoproto.moretags) = "yaml:\"source_proof\"" ];  
  desmos.profiles.v1beta1.ChainConfig source_chain_config = 3      [ (gogoproto.moretags) = "yaml:\"source_chain_config\"" ];  
  string destination_address = 4      [ (gogoproto.moretags) = "yaml:\"destination_address\"" ];  
  desmos.profiles.v1beta1.Proof destination_proof = 5      [ (gogoproto.moretags) = "yaml:\"destination_proof\"" ];
```

proto/desmos/profiles/v1beta1/query.proto
```diff
@@ -44,6 +44,12 @@ service Query {
     option (google.api.http).get =
         "/desmos/relationships/v1beta1/blocks/{user}";
   }
+
+  // ChainLink queries the user links for the user having the given address
+  rpc ChainLink(QueryChainLinkRequest) returns (QueryChainLinkResponse) {
+    option (google.api.http).get =
+        "/desmos/relationships/v1beta1/links/{user}/{chain_id}";
+  }
```
RiccardoM, line 52: Instead of a `ChainLink` that returns a single chain link for a single user, I think we should create instead: 

- `UserChainLinks` that returns all the (paginated) chain links for a user 
- `ChainLinks` that returns the paginated list of all chain links for all users

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -92,15 +94,48 @@ message UserBlock {
...
+// account to a profile
+message Proof {
+  google.protobuf.Any pub_key = 1 [ (gogoproto.moretags) = "yaml:\"pub_key\"" ];
+  string signature = 2 [ (gogoproto.moretags) = "yaml:\"signature\"" ];
+}
+
+// ChainConfig contains the data of the chain with which the link is made.
+message ChainConfig {
+  string id = 1
+      [ (gogoproto.moretags) = "yaml:\"id\"", (gogoproto.customname) = "ID" ];
```
RiccardoM, line 122: Yesterday I was thinking more about this ID and whether we should keep it as such or changing it to something more like `name`. I thought of this cause it might happen that some chains (.ie. Solana, Polkadot, etc) might have different ids for different versions while the name (i.e. `solana`, `polkadot`) stays the same. 

What do you guys think @dadamu @bragaz? Should we keep it referencing the chain id (e.g `desmos-mainnet-1`) or should we switch to a more generic chain name (e.g. `desmos`)?

dadamu: I think generic chain name is better because a key would generate same address string in the same chain system no matter it is test-net or main-net.

bragaz: Yes I think a `name` field is even better understandable from the devs that will use it later during implementations.

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -92,15 +94,48 @@ message UserBlock {
...
-}
\ No newline at end of file
+  repeated UserBlock blocks = 1 [ (gogoproto.nullable) = false ];
+}
+
+// ___________________________________________________________________________________________________________________
+
+// Proof contains all the data used to verify the signature when linking an
+// account to a profile
+message Proof {
```
RiccardoM, line 114: I think we should add an additional fields here named `value` that represents the plain text value that has been signed with the private key in order to produce the signature. Otherwise, without it we will not be able to check the validity of the signature

RiccardoM (commented):
A part from a small change inside the `Proof` type, it seems good to me. Could you please add the comments to the various Protobuf fields?

proto/desmos/profiles/v1beta1/models.proto
bragaz: @RiccardoM why not naming it `plain_text` directly?

dadamu: I agree with @bragaz, naming it `plain_text` would be more explicit.

RiccardoM: @bragaz @dadamu Yeah that's perfect

RiccardoM (approved):
Great job, looks good to me

bragaz (approved):

PR #439, opened 2021-05-25
Implement swagger documentation generation

### LINKED ISSUE #438: Add the Swagger documentation generation
Currently we have no documentation about the API endpoints that our chains provides. We should generate a Swagger file and publish it somewhere properly. Inside Cosmos this is done using the [`protoc-gen-swagger` script](https://github.com/cosmos/cosmos-sdk/blob/master/scripts/protoc-swagger-gen.sh) which is then called [from the `Makefile`](https://github.com/cosmos/cosmos-sdk/blob/master/Makefile#L390).

dadamu:
## Description

This PR is the implementation of #438, it includes:
* Fix `proto-swagger-gen` script
* Implement `proto-swagger-gen` function in Makefile
* Set proto targets to `client/docs/config.json`
* Generate `client/docs/swagger-ui/swagger.yaml` by `proto-swagger-gen`

Closes #438 

## Checklist
- [ ] Wrote unit tests.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -60,5 +60,5 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@3.7.1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          BRANCH: gh-pages
+          BRANCH: swagger-pages
           FOLDER: client/docs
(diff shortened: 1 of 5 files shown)


RiccardoM (commented):
Looks great to me. Can you please also add a GitHub workflow to build the docs properly?

.github/workflows/docs.yml
```diff
@@ -43,3 +43,22 @@ jobs:
...
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Build 🔧
+        run:
+          make proto-swagger-gen
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@3.7.1
```
RiccardoM, line 60: Can we add an `if` condition to only deploy if the branch is the `master`? I would avoid deploying the docs straight from the PRs

RiccardoM (approved):

PR #441, opened 2021-05-26
Subspaces types

bragaz:
## Description
This PR introduce the `x/subspaces` types.

## Checklist
- [x] Wrote unit tests.


RiccardoM (commented):
Can you also add comments to all Proto fields?

proto/desmos/subspaces/v1beta1/msgs.proto
```diff
@@ -0,0 +1,206 @@
...
+  string creator = 4 [ (gogoproto.moretags) = "yaml:\"creator\"" ];
+}
+
+// MsgCreateSubspaceResponse defines the Msg/CreateSubspace response type
+message MsgCreateSubspaceResponse {}
+
+// ___________________________________________________________________________________________________________________
+
+// MsgEditSubspace represents the message used to edit a subspace fields
+message MsgEditSubspace {
```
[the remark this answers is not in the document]
bragaz: I was thinking the same and I stopped while thinking about how to handle a default value of it, if the user doesn't specify it while making the tx.
Suppose we have an open subspace and we want to edit only its owner.   
What if the default value of the `--open` flag is `false`?
It will edit the open field anyway even though the user doesn't want that.
Do you think we can use a flag type `string` and converting it later inside the message itself to solve this?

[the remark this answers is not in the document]
RiccardoM: @bragaz You're right, I didn't think about this. What I think we can do to solve this is, instead of using an `open` field, we can use a `type` field. We can then have `SUBSPACE_TYPE_OPEN` and `SUBSPACE_TYPE_CLOSED` to identify the various types of subspaces. These should be implemented as an `enum` similar to [the `x/gov` one inside the Cosmos SDK](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/gov/v1beta1/gov.proto#L17)

proto/desmos/subspaces/v1beta1/subspace.proto
```diff
@@ -0,0 +1,74 @@
...
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/desmos-labs/desmos/x/staging/subspaces/types";
+
+// SubspaceType contains all the possible subspace types
+enum SubspaceType {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // SUBSPACE_TYPE_UNSPECIFIED identifies an unspecified type of subspace (used in errors)
+    SUBSPACE_TYPE_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "Unspecified"];
```
RiccardoM, line 14:
```suggestion
  SUBSPACE_TYPE_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "Unspecified"];
```

x/staging/posts/keeper/keeper_bench_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/desmos-labs/desmos/x/staging/posts/types"
 )
 
-// RandomPost returns a post with a 50% chance to have random medias and random poll
+// RandomSubspace returns a post with a 50% chance to have random medias and random poll
```
RiccardoM, line 14:
```suggestion
// RandomPost returns a post with a 50% chance to have random medias and random poll
```

x/staging/subspaces/types/errors.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+import sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+// RootCodespace is the codespace for all errors defined in this package
+const RootCodespace = "subspaces"
+
+var (
+	// ErrInvalidSubspace is returned if a subspace id is not valid
+	ErrInvalidSubspace = sdkerrors.Register(RootCodespace, 1, "invalid subspace")
```
RiccardoM, line 10: Probably we can rename this to something like `ErrInvalidSubspaceID`?

x/staging/subspaces/types/errors.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+import sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+// RootCodespace is the codespace for all errors defined in this package
+const RootCodespace = "subspaces"
```
RiccardoM, line 6: I think we can remove this and use `ModuleName` instead

RiccardoM (commented):
I'm having a really hard time deciding whether or not we should really have the `NameParams`. Subspaces will be limited and ideally the name will be never used elsewhere except when the user might want to know it. Should we really cap it?

x/staging/subspaces/types/errors.go
bragaz: yeah it will be better

bragaz:
> I'm having a really hard time deciding whether or not we should really have the NameParams. Subspaces will be limited and ideally the name will be never used elsewhere except when the user might want to know it. Should we really cap it?

I added them when I was adding a check on the `name` length inside the `MsgCreateSubspace` and `MsgEditSubspace` and remembered we did a similar thing with `profiles` parameters, first we had the hard coded checks, then we switched to `params`. I know that the usage will be different, so yes, we can replace them with some checks inside each of the `ValidateBasic()` methods. Let me know what d you think @RiccardoM .

RiccardoM:
@bragaz I think we can simply remove all checks from subspace names. We should not have a min/max length as well as any kind of check of the values themselves (eg. regex). Allowing people to create subspaces with arbitrary names won't be a problem as far as I can think of.

RiccardoM (approved):
Changes look great. Can you please create a new branch (maybe something like `leonardo/manageable-subspaces` and use that as the target branch of this PR instead of `master`? You can change it without having to close the PR

bragaz:
@RiccardoM done

Issue #447, opened 2021-05-28
Remove legacy REST endpoints


RiccardoM:
With the introduction of [gRPC](https://grpc.io/) inside Cosmos, the legacy REST endpoints have slowly became unused. Currently, all queries and actions can be performed using the gRPC endpoints instead. This allow for faster operations and easier maintenance of the code.

For this reason, I think we should remove all the legacy REST endpoints from Desmos as well. We can remove all the `clint/rest` folders and leave only the CLI interface. Clients will still be able to use gRPC to interact with any module anyway.

PR #449, opened 2021-05-28
Implement the logic of ibc profiles features

dadamu:
## Description

This PR is the ibc logic part of the implementation related to #192.

The features include:
* Implementation of chain link related type ( proof, chain link, chain config )
* IBC connection and relay functions
* Packet and msg handlers
* The keeper functions to store chain link
* Chain link query methods and cli command functions

## Checklist
- [ ] Wrote unit tests.


bragaz:
Hey @dadamu, you can convert this to a PR now

x/profiles/client/cli/query.go
```diff
@@ -175,3 +175,78 @@ func GetCmdQueryUserBlocks() *cobra.Command {
 
 	return cmd
 }
+
+// GetCmdQueryChainsLinks returns the command allowing to query all the links
+func GetCmdQueryChainsLinks() *cobra.Command {
```
RiccardoM, line 180: Instead of having two different commands (`GetCmdQueryChainsLinks` and `GetCmdQueryUserChainsLinks`), I think we should only have one (`GetCmdQueryUserChainsLinks`). 

This then should have an optional parameter named `[user]` that: 
- if specified, allows to query the links for such user
- if not specified, allows to query all the links for all the users.

This can be done using the `cobra.RangeArgs(0, 1)` and then checking whether the optional parameter is empty or not. You can see an example [here](https://github.com/cosmos/cosmos-sdk/blob/master/x/distribution/client/cli/query.go#L255)

x/profiles/keeper/keeper.go
```diff
@@ -32,7 +39,13 @@ type Keeper struct {
...
 //    This is used to get the address of a user based on a DTag
 func NewKeeper(
-	cdc codec.BinaryMarshaler, storeKey sdk.StoreKey, paramSpace paramstypes.Subspace, ak authkeeper.AccountKeeper,
+	cdc codec.BinaryMarshaler,
+	storeKey sdk.StoreKey,
+	paramSpace paramstypes.Subspace,
+	ak authkeeper.AccountKeeper,
+	channelKeeper ibctypes.ChannelKeeper,
+	portKeeper ibctypes.PortKeeper,
+	scopedKeeper capabilitykeeper.ScopedKeeper,
```
RiccardoM, line 48: Instead of depending on the `ibctypes` and `capabilitykeeper` packages, I think it's better to define the expected keeper definitions inside a new file named `expected_keeper` inside the `types` folder the same way that is being done [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/expected_keepers.go)

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkquery "github.com/cosmos/cosmos-sdk/types/query"
+
+	"github.com/desmos-labs/desmos/x/profiles/types"
+)
+
+// StoreChainLink sotres the given chain link inside the current context.
```
RiccardoM, line 13:
```suggestion
// StoreChainLink stores the given chain link inside the current context.
```

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,93 @@
...
+// GetChainLink returns the chain link corresponding to the given address and the given chain name inside the current context.
+func (k Keeper) GetChainLink(ctx sdk.Context, address string, chainName string) (link types.ChainLink, found bool) {
+	store := ctx.KVStore((k.storeKey))
+
+	bz := store.Get(types.ChainsLinksStoreKey(address, chainName))
+	if bz != nil {
+		k.cdc.MustUnmarshalBinaryBare(bz, &link)
+		return link, true
+	}
+	return types.ChainLink{}, false
```
RiccardoM, line 36: I think it's better to reverse the check here, making it more similar to other code parts: 

```suggestion
	bz := store.Get(types.ChainsLinksStoreKey(address, chainName))
	if bz == nil {
		return types.ChainLink{}, false
	}
	
	k.cdc.MustUnmarshalBinaryBare(bz, &link)
	return link, true
```

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,117 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/desmos-labs/desmos/x/profiles/types"
+)
+
+func (k msgServer) LinkChainAccount(goCtx context.Context, msg *types.MsgLinkChainAccount) (*types.LinkChainAccountResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, err
+	}
```
RiccardoM, line 18: This check can be skipped as it will be performed from the chain itself automatically

```suggestion
```

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Check if address has the profile
+	profile, found, err := k.GetProfile(ctx, msg.DestinationAddress)
+	if err != nil {
+		return nil, err
+	}
+	if !found {
+		return nil, fmt.Errorf("address does not have any profile")
+	}
```
RiccardoM, line 35: This should be moved inside the `StoreLink` method to make sure the checks are always performed even if the method is called from other places

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := k.StoreChainLink(ctx, link); err != nil {
+		return nil, err
+	}
+
+	// Store chain link to the profile
+	profile.ChainsLinks = append(profile.ChainsLinks, link)
+	if err := k.StoreProfile(ctx, profile); err != nil {
+		k.DeleteChainLink(ctx, link.ChainConfig.Name, link.Address)
+		return nil, err
+	}
```
RiccardoM, line 53: This should be included inside the `StoreChainLink` to avoid having to remember that they should be called along with it.

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &types.LinkChainAccountResponse{}, nil
+}
+
+func (k msgServer) UnlinkChainAccount(goCtx context.Context, msg *types.MsgUnlinkChainAccount) (*types.UnlinkChainAccountResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if err := msg.ValidateBasic(); err != nil {
+		return nil, err
+	}
```
RiccardoM, line 71: As above, this can be removed as it will be performed by the chain automatically

```suggestion
```

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,117 @@
...
+	if !isTargetExist {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, ("non existent target chain link in the profile"))
+	}
+
+	// Update profile status
+	profile.ChainsLinks = newChainsLinks
+	err = k.StoreProfile(ctx, profile)
+	if err != nil {
+		return nil, err
+	}
```
RiccardoM, line 105: All of this should be handled inside the `DeleteChainLink` method to make sure the logic is all there and we never forget to do anything if the method is called from other places

x/profiles/keeper/relay_ibc.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	channeltypes "github.com/cosmos/cosmos-sdk/x/ibc/core/04-channel/types"
+	"github.com/desmos-labs/desmos/x/profiles/types"
+)
+
+// OnRecvPacket processes packet reception
+func (k Keeper) OnRecvPacket(
```
RiccardoM, line 13: Most of the logic present here can be simply moved inside the `StoreChainLink` method as mentioned above as well. This would reduce the code duplication and make the code more easily maintainable

x/profiles/keeper/relay_ibc.go
```diff
@@ -0,0 +1,87 @@
...
+		return packetAck, err
+	}
+
+	packetAck.SourceAddress = data.SourceAddress
+	return packetAck, nil
+}
+
+// OnAcknowledgementPacket responds to the the success or failure of a packet
+// acknowledgement written on the receiving chain.
+func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context,
```
RiccardoM, line 64: I don't think we should care about acknowledgements from other chains. We can simply remove this method and have an empty `OnAcknowledgmentPacket` inside the `module.go` file

x/profiles/module_ibc.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	// NOTE: acknowledgement will be written synchronously during IBC handler execution.
+	return &sdk.Result{
+		Events: ctx.EventManager().Events().ToABCIEvents(),
+	}, ackBytes, nil
+}
+
+// OnAcknowledgementPacket implements the IBCModule interface
+func (am AppModule) OnAcknowledgementPacket(
```
RiccardoM, line 163: IMO we should not care about received ack packets. These are sent only by third party chains when we perform a request towards them. Since we are not doing this kind of thing, we can simply have this method body empty and avoid dealing with ack packets all together

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,158 @@
...
+		Name:             name,
+		Bech32AddrPrefix: prefix,
+	}
+}
+
+func (chainConfig ChainConfig) Validate() error {
+	if chainConfig.Name == "" {
+		return fmt.Errorf("chain name cannot be empty")
+	}
+	return nil
```
RiccardoM, line 28: We should also check to make sure the Bech32 prefix is not empty

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,158 @@
...
+func (proof Proof) Validate() error {
+	_, err := hex.DecodeString(proof.Signature)
+	if err != nil {
+		return fmt.Errorf("failed to decode hex string of signature")
+	}
+
+	if strings.TrimSpace(proof.PlainText) == "" {
+		return fmt.Errorf("plain text can not be empty or blank")
+	}
+	return nil
```
RiccardoM, line 55: I think we should also make sure the provided pub key is a valid pub key (ie. it can be deserialized to a `cyrptotypes.PubKey` instance properly).

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if err := link.ChainConfig.Validate(); err != nil {
+		return err
+	}
+
+	if err := link.Proof.Validate(); err != nil {
+		return err
+	}
+
+	return nil
```
RiccardoM, line 135: We should make sure the `CreationTime` is not zero

x/profiles/types/chain_link.go
dadamu: It might be empty if the link is from the chain without bech32 encoded address.

x/profiles/types/chain_link.go
dadamu: If do so, it need to have a unpacker pass to the function because Proof.PubKey is Any type.
Then, when calling ValidateBasic() function to validate msg or packet, the Proof.Validate() can not be used directly cause ValidateBasic() does not allow to pass value.

RiccardoM: You're right. Can we at least check if it's a valid `hex` string (just like the signature)?

x/profiles/types/chain_link.go
RiccardoM: @dadamu Are there chains that do not use the Bech 32 encoding for their addresses?

dadamu: @RiccardoM Solana is using base58 public key as address.

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	bz := store.Get(types.ChainsLinksStoreKey(chainName, address))
+	if bz == nil {
+		return types.ChainLink{}, false
+	}
+	k.cdc.MustUnmarshalBinaryBare(bz, &link)
+	return link, true
+}
+
+// GetAllChainsLinks returns a list of all the chains links inside the given context.
+func (k Keeper) GetAllChainsLinks(ctx sdk.Context) []types.ChainLink {
```
RiccardoM, line 54: Can you move this method to a new file named `alias_functions`? It's usually where we place all these kind of methods that use iterators inside them to implement a utility method

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	for ; iterator.Valid(); iterator.Next() {
+		var link types.ChainLink
+		k.cdc.MustUnmarshalBinaryBare(iterator.Value(), &link)
+		links = append(links, link)
+	}
+	return links
+}
+
+// GetChainLinksWithPagination returns a list of the links which is paginated from all the links
+func (k Keeper) GetChainsLinksWithPagination(ctx sdk.Context, page int, limit int) []types.ChainLink {
```
RiccardoM, line 69: I'm not convinced this is the best way to implement this method. The `page` and `limit` can be provided by the client empty. This is the case when the client used a particular `Key` instead. 

For this reason, I think we should just simply paginate them as it's now done inside the queries, removing this method all together.

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	if start < 0 || end < 0 {
+		pagedLinks = []types.ChainLink{}
+	} else {
+		pagedLinks = links[start:end]
+	}
+	return pagedLinks
+}
+
+// GetUserChainsLinks returns a list of links by a given address and pagination params
+func (k Keeper) GetUserChainsLinks(ctx sdk.Context, address string, page int, limit int) ([]types.ChainLink, error) {
```
RiccardoM, line 85: I think we can safely remove this query. Since chain links for a user are stored inside their profile, if someone wants to know them they can simply use the `desmos query auth account` command and it would return the entire list. 

It would be a lot more useful to have a query that allows you to get the profile based on a given chain id and profile: 

```golang
type QueryProfileByLink struct {
  ChainID string
  Address string
}
```

(This also applies to the CLI and gRPC queries)

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	sdkquery "github.com/cosmos/cosmos-sdk/types/query"
+
+	"github.com/desmos-labs/desmos/x/profiles/types"
+)
+
+// StoreChainLink stores the given chain link inside the current context.
+// It assumes that the given chain link has already been validated.
+func (k Keeper) StoreChainLink(ctx sdk.Context, link types.ChainLink, destinationAddress string) error {
```
RiccardoM, line 16: I think it would be better to: 
1. Rename the `destinationAddress` parameter to simply `user` 
2. Reorder the parameter making it `user string, link types.ChainLink`

This would make the method easier to comprehend what it does

x/profiles/types/chain_link.go
dadamu: @RiccardoM I think it can not be checked without unpacking, only can be checked if it is `nil` .

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	if _, found := k.GetChainLink(ctx, link.ChainConfig.Name, link.Address); found {
+		return fmt.Errorf("chain link already exists")
+	}
+
+	// Check if address has the profile
```
RiccardoM, line 21: I think there might be a missing check here. We should also check if the recipient address has a profile as well. This can be done by reading the address value using either the `sdk.AccAddressFromBech` or the `sdk.AccAddressFromHex` methods, and seeing if the associated Desmos account has a profile. 

As we discussed previously, we must avoid connecting together two Desmos profiles.

x/profiles/types/chain_link.go
RiccardoM: You're right, I forgot it was an `Any` and not a hex string.

x/profiles/types/chain_link.go
RiccardoM: @dadamu Ok, I see. Than instead of 

```proto
message ChainLink {
  string address = 1;
  // ...
}
```

I think it might be better to use something like 
```proto
message ChainLink {
  Address address = 1; 
  // ...
}

message Address {
  oneof sum {
    Bech32 bech_32 = 1;
    Base58 base_58 = 2;

  }

  message Bech32 {
    string value = 1;
    string prefix = 2;
  }

  message Base58 {
    string value = 1;
  }
}
```

This way we will be able to represent various encoding methods differently with each one their additional data (eg. the `prefix` for Bech32). If this approach is used, we can then remove the `bech32_addr_prefix` from the `ChainConfig` object.

x/profiles/client/cli/query.go
```diff
@@ -175,3 +176,53 @@ func GetCmdQueryUserBlocks() *cobra.Command {
...
+					context.Background(),
+					&types.QueryUserChainsLinksRequest{
+						User:       args[0],
+						Pagination: pageReq,
+					})
+				if err != nil {
+					return err
+				}
+				return clientCtx.PrintProto(res)
+			}
```
bragaz, line 209: Maybe here you could use an `else` and avoid to rewrite the following code twice
```go
if err != nil {
       return err
}				
return clientCtx.PrintProto(res)
```

x/profiles/keeper/keeper_chain_link.go
dadamu: @RiccardoM In current chain link store logic and its struct, the chain link does not include the `destAddr` that can retrieve the profile easily.  Should it have a prefix to store the relationship between chain link and profile? Say:
```go
func (k Keeper) StoreLinkToProfile(ctx, profile, link) {
    key := types.ChainLinkToProfileStoreKey(llink.Address)
    store.Set(key, []byte(profile.GetAddress().String()))
}
``` 

Or adding new field named `destAddr` to `ChainLink` is better than above one?

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,148 @@
...
+	if !found {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, ("non existent profile on owner address"))
+	}
+
+	isTargetExist := false
+	newChainsLinks := []types.ChainLink{}
+	// Try to find the target link
+	for _, link := range profile.ChainsLinks {
+		currChainName := link.ChainConfig.Name
+		currAddr := link.Address
```
bragaz, line 126: I think we can remove `currChainName` and `currAddr` since we're not using them anywhere else

x/profiles/keeper/keeper_chain_link.go
dadamu: @RiccardoM This method is for queryChainsLinks function in `querier.go`. It is not gRPC method to query here so the `sdkquery.FilteredPaginate` method can not be used.

`querier.go`:
https://github.com/desmos-labs/desmos/blob/paul/ibc-links/x/profiles/keeper/querier.go#L149

`sdkquery.FilteredPaginate` in `grpc_query.go`:
https://github.com/desmos-labs/desmos/blob/paul/ibc-links/x/profiles/keeper/grpc_query.go#L93

x/profiles/client/cli/query.go
dadamu: The `UserChainsLinks` query function will be removed.

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,130 @@
...
+	err = k.StoreProfile(ctx, profile)
+	if err != nil {
+		return err
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	key := types.ChainsLinksStoreKey(chainName, target)
+	store.Delete(key)
+	return nil
+}
```
bragaz, line 120: I think you can simplify this a little like this:
```go
func (k Keeper) DeleteChainLink(ctx sdk.Context, owner, chainName, target string) error {
	// Check if address has the profile and get the profile
	profile, found, err := k.GetProfile(ctx, owner)
	if err != nil {
		return err
	}
	if !found {
		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "non existent profile on owner address")
	}

	doesLinkExists := false
	// Try to find the target link
	for index, link := range profile.ChainsLinks {
		if link.ChainConfig.Name == chainName && link.Address == target {
			doesLinkExists = true
			newChainsLinks := append(profile.ChainsLinks[:index], profile.ChainsLinks[index+1:]...)
			profile.ChainsLinks = newChainsLinks
			// Update profile status
			if err = k.StoreProfile(ctx, profile); err != nil {
				return err
			}
			break
		}

	}

	if !doesLinkExists {
		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "non existent target chain link in the profile")
	}
	key := types.ChainsLinksStoreKey(chainName, target)
	store.Delete(key)
	return nil
}
```

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,130 @@
...
+	start, end := client.Paginate(len(links), page, limit, sdkquery.DefaultLimit)
+	if start < 0 || end < 0 {
+		pagedLinks = []types.ChainLink{}
+	} else {
+		pagedLinks = links[start:end]
+	}
+	return pagedLinks
+}
+
+// DeleteLink allows to delete a link associated with the given address and chain name inside the current context.
```
bragaz, line 83:
```suggestion
// DeleteChainLink allows to delete a link associated with the given address and chain name inside the current context.
```

x/profiles/client/cli/query.go
```diff
@@ -175,3 +176,39 @@ func GetCmdQueryUserBlocks() *cobra.Command {
 
 	return cmd
 }
+
+// GetCmdQueryChainsLinks returns the command allowing to query all the links
+func GetCmdQueryChainsLinks() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "links",
+		Short: "Retrieve all links or links from a particular user by given pagination flags",
```
bragaz, line 184:
```suggestion
		Short: "Retrieve all the links by given pagination flags",
```

x/profiles/keeper/keeper_ibc.go
```diff
@@ -0,0 +1,55 @@
...
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.IBCPortKey, []byte(portID))
+}
+
+// AuthenticateCapability wraps the scopedKeeper's AuthenticateCapability function
+func (k Keeper) AuthenticateCapability(ctx sdk.Context, cap *capabilitytypes.Capability, name string) bool {
+	return k.scopedKeeper.AuthenticateCapability(ctx, cap, name)
+}
+
+// ClaimCapability allows the module that can claim a capability that IBC module passes to it
```
bragaz, line 52:
```suggestion
// ClaimCapability wraps the scopedKeeper's ClaimCapability method
```

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,159 @@
...
+	return Proof{
+		PubKey:    pubKeyAny,
+		Signature: signature,
+		PlainText: plainText,
+	}
+}
+
+func (proof Proof) Validate() error {
+	if proof.PubKey == nil {
+		return fmt.Errorf("public key field can not be empty")
```
bragaz, line 48:
```suggestion
		return fmt.Errorf("public key field cannot be empty")
```

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	_, err := hex.DecodeString(proof.Signature)
+	if err != nil {
+		return fmt.Errorf("failed to decode hex string of signature")
+	}
+	if strings.TrimSpace(proof.PlainText) == "" {
+		return fmt.Errorf("plain text can not be empty or blank")
```
bragaz, line 55:
```suggestion
		return fmt.Errorf("plain text cannot be empty or blank")
```

x/profiles/types/chain_link.go
dadamu: I stuck on the RegisterInterfaces in `codec.go` when implementing.

When it tried to `GetSignBytes()` in link chain account message, there is a panic said `"panic: Unregistered interface types.isAddress_Sum"`.
Then I tried to register `Address_Bech_32` and `Address_Base_58` type to `codec` like:
```
registry.RegisterImplementations((*isAddress_Sum)(nil), &Address_Bech_32{}, &Address_Base_58{})
```
However, It shows the message that `cannot use &(Address_Base_58 literal) (value of type *Address_Base_58) as proto.Message value in argument to registry.RegisterImplementations: missing method ProtoMessage`.

Could you help me? @RiccardoM @bragaz

dadamu: I have an idea that is using the following struct to avoid the problem I mentioned above.
```proto
message MsgLinkAccount{
    string source_address = 1;
    string source_prefix =2;
    ...
}
```
What do you think?

x/profiles/keeper/keeper_chain_link.go
RiccardoM: As per #447 I think we can safely remove the legacy REST endpoints for queries. Clients should rely only on gRPC ones. So we can remove the entire `client/rest` folder and all the associated keeper methods (like this one)

x/profiles/keeper/keeper_chain_link.go
RiccardoM: @dadamu In my head, what I think we should do is the following: 

1. Store the chain link inside the `Profile` object.  
   This makes it super easy to get all the links given a Desmos address associated with a profile. 
2. Store the `Profile` address associating it to the chain link address.  
   This makes it super easy to get the `Profile` object given a chain address.

Summarizing, I think we can store: 

```
Profile -> []ChainLink
ChainLink.Address.String() -> Profile.Address
```

Let me know if there's something not clear

dadamu: @RiccardoM The `Address` of  `ChainLink` only stores source address now. It is like:
```proto
message ChainLink {
  Address address = 1; 
  // ...
}
message Address {
  oneof sum {
    Bech32 bech_32 = 1;
    Base58 base_58 = 2;

}
```
That's why I think It should have a new field `destination` to be related to `Profile`. Say:
```proto
message ChainLink {
  Address address = 1; 
  // ...
  string destination = 4;
  Timestamp creation_time = 5;
}
```

RiccardoM: @dadamu There's no need to store a destination address inside the `ChainLink` object itself. We can simply avoid storing such object and simply store the following association: 

Inside package `types`
```golang
var (
  ChainLinkPrefix = []byte("chain_link")
)

// ChainLinkKey returns the key used to identify the link to the given chain for the given address
func ChainLinkKey(chainName, chainAddress string) []byte {
  return appen(ChainLinkPrefix, bytes(chainName + chainAddress)
}
```

Then we can store the association like this:
```golang
// Store the link -> address association
store.Set(types.ChainLinkKey(link.ChainName, link.Address.String(), profile.GetAddress())
```

And when a user asks the profile associated with a specific chain address they can use the following command: 

```
desmos q profiles chain-link [chain-name] [chain-address]
```

Which can be implemented as follows: 

```golang
bz := store.Get(types.ChainLinkKey(chainName, chainAddress)
profile := k.accountKeeper.GetAccount(sdk.AccAddress(bz)) 
```

x/profiles/types/chain_link.go
dadamu: @RiccardoM I got it, `Message` and `Packet` should also use the `Any` type on `source_address` field. Not sure if it occurs the marshaling error.  I would try it first.

x/profiles/keeper/keeper_chain_link.go
dadamu: @RiccardoM With it, I think we should also remove `ChainsLinks` method that query all chains links cause the links detail are stored in `Profile` object.

RiccardoM: @dadamu Yes, I agree we should remove that as well

x/profiles/types/chain_link.go
dadamu: @RiccardoM I got the problem that when using msg `GetSigners()` method, it should have unpacker to unpack the `Any` type address.

Could we have a simple `AddressData` type shows if it has prefix or not? Say:
```proto
message AddressData{
  string address
  string prefix
}
```
If the chain address doesn't have prefix, just let it empty string.

dadamu: @RiccardoM I think a while about the `MsgLinkChainAccount` should have a `signer` field for `GetSigners()` function. In the case that the source account is from `Solana`. The msg `GetSigners()` will get the `Solana` address.

```proto
message MsgLinkChainAccount{
  string signer = 1;
  types.Any source_address = 2 ; // implement AddressData interface
  // ...
}
```

What do you think?

RiccardoM: @dadamu Ideally, the message signer is going to be the destination address. This will represent the address associated to the Desmos profile, and for this reason will always be a valid Desmos address.

When I link an external chain to my profile, I should be the message signer. There should not be another signer aside from me

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -125,24 +127,46 @@ message Proof {
...
-      [ (gogoproto.moretags) = "yaml:\"bech32_addr_prefix\"" ];
 }
 
 // ChainLink contains the data representing either an inter- or cross- chain link
 message ChainLink {
-  string address = 1 [ (gogoproto.moretags) = "yaml:\"address\"" ];
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+  
+  google.protobuf.Any address = 1 [ (gogoproto.moretags) = "yaml:\"address\"" ];
```
RiccardoM, line 142: Can we add the `cosmos_proto.accepts_interface` Protobuf option here the same way that [Cosmos does](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/auth/v1beta1/query.proto#L40)? I think it would help us remember which interface should be used in the future and for other developers

```protobuf
  google.protobuf.Any address = 1 [ (gogoproto.moretags) = "yaml:\"address\"", (cosmos_proto.accepts_interface) = "AddressData" ];
```

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -125,24 +127,46 @@ message Proof {
...
     (gogoproto.moretags) = "yaml:\"creation_time\"",
     (gogoproto.nullable) = false
   ];
 }
+
+ // Bech32Address represents a Bech32-encoded address
+ message Bech32Address {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
```
RiccardoM, line 161: Can we add the `cosmos_proto.implements_interface` Protobuf option here the same way that [Cosmos does](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/auth/v1beta1/auth.proto#L18? I think it will be helpful in the future:

```protobuf
option (cosmos_proto.implements_interface) = "AddressData";
```

proto/desmos/profiles/v1beta1/models.proto
```diff
@@ -125,24 +127,46 @@ message Proof {
...
+  string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+  string prefix = 2 [ (gogoproto.moretags) = "yaml:\"prefix\"" ];
+}
+
+// Base58Address represents a Base58-encoded address
+message Base58Address {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
```
RiccardoM, line 170: Same as above, I think we should add the `implements_interface` option here as well

proto/desmos/profiles/v1beta1/query.proto
```diff
@@ -46,16 +46,10 @@ service Query {
...
+  rpc ProfileByChainLink(QueryProfileByChainLinkRequest) returns (QueryProfileByChainLinkResponse) {
     option (google.api.http).get =
-        "/desmos/relationships/v1beta1/links";
-  }
-
-  // UserChainLinks queries all the links for the user having the given address
-  rpc UserChainLinks(QueryUserChainLinksRequest) returns (QueryUserChainLinksResponse) {
-    option (google.api.http).get =
-        "/desmos/relationships/v1beta1/links/{user}";
+      "/desmos/profiles/v1beta1/profiles/{chain_name}/{target}";
```
RiccardoM, line 52: I think it would be better to rename this `chain-links` instead. Also using `target_address` might be better: 

```suggestion
      "/desmos/profiles/v1beta1/chain-links/{chain_name}/{target_address}";
```

proto/desmos/profiles/v1beta1/query.proto
```diff
@@ -151,47 +145,16 @@ message QueryUserBlocksResponse {
...
-  option (gogoproto.goproto_getters) = false;
-
-  // address of the user to query the links for
-  string user = 1;
-
-  // pagination defines an optional pagination for the request.
-  cosmos.base.query.v1beta1.PageRequest pagination = 2;
+message QueryProfileByChainLinkRequest {
+  string chain_name = 1 ;
+  string target = 2 ;
```
RiccardoM, line 152: I think we should rename this `target_address` to make it easier to comprehend what it should contain

x/profiles/keeper/grpc_query.go
```diff
@@ -87,3 +87,29 @@ func (k Keeper) Params(ctx context.Context, _ *types.QueryParamsRequest) (*types
...
 	return &types.QueryParamsResponse{Params: params}, nil
 }
+
+func (k Keeper) ProfileByChainLink(ctx context.Context, request *types.QueryProfileByChainLinkRequest) (*types.QueryProfileByChainLinkResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	account, found := k.GetAccountByChainLink(sdkCtx, request.ChainName, request.Target)
+	if !found {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"No link related to this address: %s", request.Target)
```
RiccardoM, line 97:
```suggestion
			"no link related to this address: %s", request.Target)
```

x/profiles/keeper/grpc_query.go
```diff
@@ -87,3 +87,29 @@ func (k Keeper) Params(ctx context.Context, _ *types.QueryParamsRequest) (*types
...
+	}
+
+	profile, found, err := k.GetProfile(sdkCtx, account.String())
+	if err != nil {
+		return nil, err
+	}
+	if !found {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"Profile with address %s doesn't exists", account.String())
```
RiccardoM, line 106:
```suggestion
			"profile with address %s does not exist", account.String())
```

x/profiles/keeper/keeper.go
```diff
@@ -55,6 +68,12 @@ func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked string) bool {
 // It assumes that the given profile has already been validated.
 // It returns an error if a profile with the same DTag from a different creator already exists
 func (k Keeper) StoreProfile(ctx sdk.Context, profile *types.Profile) error {
+
+	if _, found := k.GetAccountByChainLink(ctx, "desmos", profile.GetAddress().String()); found {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"creator has already been associated to other profile")
```
RiccardoM, line 74:
```suggestion
			"owner address has already been linked to another profile")
```

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return nil
+}
+
+// ___________________________________________________________________________________________________________________
+
+type AddressData interface {
+	proto.Message
+	Validate() error
+	GetAddressString() string
```
RiccardoM, line 76: Can we rename this to simply be `String`? Since the interface already represents an address, I think it's redundant to name it `GetAddressString`

RiccardoM, line 76: Can you please add the docs to these methods? I think it would be important for developers to know what they should do when implementing them

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil
+}
+
+// UnpackInterfaces implements codectypes.UnpackInterfacesMessage
+func (link *ChainLink) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
+	if link != nil {
+		var pubKey cryptotypes.PubKey
+		if err := unpacker.UnpackAny(link.Proof.PubKey, &pubKey); err != nil {
+			return err
+		}
```
RiccardoM, line 156: Instead of implementing this here, we should implement the `UnpackInterfaces` method for the `Proof` object. This would make it possible to serialize/deserialize such object individually as well

x/profiles/types/msgs.go
```diff
@@ -486,3 +488,128 @@ func (msg MsgUnblockUser) GetSigners() []sdk.AccAddress {
...
+		return err
+	}
+
+	var pubKey cryptotypes.PubKey
+	if err := unpacker.UnpackAny(msg.SourceProof.PubKey, &pubKey); err != nil {
+		return err
+	}
+	if err := unpacker.UnpackAny(msg.DestinationProof.PubKey, &pubKey); err != nil {
+		return err
+	}
```
RiccardoM, line 560: There's no need to do this by hand since if the field types implement `UnpackInterfaces`, this method will call them as well automatically.

x/profiles/types/packets.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// GetBytes is a helper for serialising
+func (p LinkChainAccountPacketData) GetBytes() ([]byte, error) {
+	var modulePacket LinkChainAccountPacketData
+	return sdk.SortJSON(ProtoCdc.MustMarshalJSON(&modulePacket))
```
RiccardoM, line 55: Instead of using `ProtoCdc`, I think we should do the same that Cosmos does and use `ModuleCdc` instead as you can see [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/packet.go#L55). If we change this, then we can also get rid of the `ProtoCdc` variable

x/profiles/types/error.go
```diff
@@ -0,0 +1,13 @@
...
+// DONTCOVER
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// x/ibc/profiles module sentinel errors
+var (
+	ErrInvalidVersion      = sdkerrors.Register(ModuleName, 1501, "invalid version")
+	ErrMaxProfilesChannels = sdkerrors.Register(ModuleName, 1, "max profiles channels")
```
RiccardoM, line 12: I think we can change the error codes without any problem here: 
```suggestion
	ErrInvalidVersion      = sdkerrors.Register(ModuleName, 1, "invalid version")
	ErrMaxProfilesChannels = sdkerrors.Register(ModuleName, 2, "max profiles channels")
```

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,176 @@
+package types
```
RiccardoM, line 1: Can you add the documentation to all the methods and functions that do not have it yet please?

x/profiles/types/packets.go
dadamu: @RiccardoM Their `ModuleCdc`  is using `ProtoCdc` [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/codec.go#L32).
It is different from desmos using `AnimoCodec` and register crypto type concrete [here](https://github.com/desmos-labs/desmos/blob/master/x/profiles/types/codec.go#L56).

x/profiles/types/msgs.go
dadamu: If removing it, the msg `GetSignBytes()` will occurs panic, it said:
```
JSON marshal marshaling error for &Any{TypeUrl:/cosmos.crypto.secp256k1.PubKey,Value:[10 33 3 187 247 28 72 226 52 237 163 92 144 62 218 7 1 180 237 149 67 165 108 138 187 59 24 224 243 237 211 119 244 126 208],XXX_unrecognized:[],}, 
this is likely because amino is being used directly (instead of codec.LegacyAmino which is preferred) 
or UnpackInterfacesMessage is not defined for some type which contains a protobuf Any either directly or via one of its members.
```

x/profiles/types/chain_link.go
dadamu: @RiccardoM String() would conflict with `proto.Message`. Maybe `AddressString()` is better.

RiccardoM: @dadamu Then maybe something like `GetAddress` is better, without having to specify the type inside it

x/profiles/types/msgs.go
RiccardoM: I think this is related to this: https://github.com/desmos-labs/desmos/pull/449#discussion_r644591486. We should implement the `UnpackInterfacesMessage` for the `Proof` object directly. Then it will be called recursively on it

dadamu: @RiccardoM The panic happens after implementing `UnpackInterfacesMessage ` for the Proof. `GetSignBytes()` calls `MustMarshalJSON()` function to deserialize the proto message. Not sure if it calls `UnpackInterfacesMessage ` as well, let me check it.

RiccardoM: @dadamu Can you reproduce this on a test please? I would like to take a look

dadamu: @RiccardoM 
It seems that `UnpackInterfacesMessage` doesn't call the child property's `UnpackInterface` recursively. The interface code is [here](https://github.com/cosmos/cosmos-sdk/blob/master/codec/types/interface_registry.go#L60).

I think after implementing `UnpackInterface` for Proof, it can be:
```go
func (msg *MsgLinkChainAccount) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
	var address AddressData
	if err := unpacker.UnpackAny(msg.SourceAddress, &address); err != nil {
		return err
	}
	if err := msg.SourceProof.UnpackInterfaces(unpacker); err != nil {
		return err
	}
	if err := msg.DestinationProof.UnpackInterfaces(unpacker); err != nil {
		return err
	}
	return nil
}
```

dadamu: @RiccardoM 
I create a test branch [here](https://github.com/dadamu/desmos/tree/test/proof-unpack).

RiccardoM: @dadamu You're right, we should call the `UnpackInterfaces` for the nested fields manually using the code you have written. It's also specified [here](https://github.com/cosmos/cosmos-sdk/blob/master/codec/types/interface_registry.go#L71) as you pointed out

x/profiles/client/cli/query.go
```diff
@@ -175,3 +176,31 @@ func GetCmdQueryUserBlocks() *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+			queryClient := types.NewQueryClient(clientCtx)
+
+			res, err := queryClient.ProfileByChainLink(
+				context.Background(),
+				&types.QueryProfileByChainLinkRequest{ChainName: args[0], TargetAddress: args[1]})
```
RiccardoM, line 194:
```suggestion
			res, err := queryClient.ProfileByChainLink(
				context.Background(),
				&types.QueryProfileByChainLinkRequest{ChainName: args[0], TargetAddress: args[1]},
			)
```

x/profiles/keeper/keeper.go
```diff
@@ -55,6 +68,12 @@ func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked string) bool {
 // It assumes that the given profile has already been validated.
 // It returns an error if a profile with the same DTag from a different creator already exists
 func (k Keeper) StoreProfile(ctx sdk.Context, profile *types.Profile) error {
+
+	if _, found := k.GetAccountByChainLink(ctx, "desmos", profile.GetAddress().String()); found {
```
RiccardoM, line 72: I think we should define a constant named `DesmosChainName` inside the `types/keys.go` file and use that here instead of manually using `"desmos"`

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+
+	target := srcAddrData.GetAddress()
+	if _, found := k.GetAccountByChainLink(ctx, link.ChainConfig.Name, target); found {
+		return fmt.Errorf("chain link already exists")
+	}
+
+	// check target address has a profile or not
+	if link.ChainConfig.Name == "desmos" {
```
RiccardoM, line 27: Once the above mentioned constant is defined, we should use it here as well instead of the string value `"desmos"`

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,121 @@
...
+		if _, err := sdk.AccAddressFromBech32(target); err != nil {
+			return err
+		}
+		_, found, err := k.GetProfile(ctx, target)
+		if err != nil {
+			return err
+		}
+		if found {
+			return fmt.Errorf("the target address has profiles already")
```
RiccardoM, line 36:
```suggestion
			return fmt.Errorf("cannot link two profiles together")
```

x/profiles/keeper/keeper_chain_link.go
```diff
@@ -0,0 +1,121 @@
...
+
+// GetAccountByChainLink returns the account corresponding to the given address and the given chain name inside the current context.
+func (k Keeper) GetAccountByChainLink(ctx sdk.Context, chainName string, address string) (account sdk.AccAddress, found bool) {
+	store := ctx.KVStore((k.storeKey))
+
+	bz := store.Get(types.ChainsLinksStoreKey(chainName, address))
+	if bz == nil {
+		return nil, false
+	}
+	return account, true
```
RiccardoM, line 68: We are returning the `account` variable here, which is never initialized. This should return `sdk.AccAddress(bz)` instead. I suggest you removing the named returned values and simply return `(sdk.AccAddress, bool)` instead to avoid such errors

proto/desmos/profiles/v1beta1/query.proto
```diff
@@ -46,16 +46,10 @@ service Query {
         "/desmos/relationships/v1beta1/blocks/{user}";
   }
 
-  // ChainLinks queries all the links for all users
-  rpc ChainLinks(QueryChainLinksRequest) returns (QueryChainLinksResponse) {
+  // ProfileByChainLink the profile for the user having the given address and chain name
```
bragaz, line 49:
```suggestion
  // ProfileByChainLink queries the profile for the user having the given address and chain name
```

x/profiles/keeper/grpc_query.go
```diff
@@ -87,3 +87,29 @@ func (k Keeper) Params(ctx context.Context, _ *types.QueryParamsRequest) (*types
 	params := k.GetParams(sdkCtx)
 	return &types.QueryParamsResponse{Params: params}, nil
 }
+
```
bragaz, line 90:
```suggestion

// ProfileByChainLink implements the Query/ProfileByChainLink gRPC method
```

x/profiles/types/chain_link.go
```diff
@@ -0,0 +1,198 @@
...
+	return nil
+}
+
+// GetAddress returns the address value
+func (address Bech32Address) GetAddress() string {
+	return address.Value
+}
+
+// NewBase58Address is a constructor function for Base58Address
+func NewBase58Address(value, prefix string) *Base58Address {
```
bragaz, line 116: looks like `prefix` is never used here

x/profiles/types/packets.go
```diff
@@ -0,0 +1,56 @@
...
+		DestinationProof:   destinationProof,
+	}
+}
+
+// Validate is used for validating the packet
+func (p LinkChainAccountPacketData) Validate() error {
+	if p.SourceAddress == nil {
+		return fmt.Errorf("source address cannot be nil")
+	}
+
```
bragaz, line 35:
```suggestion
```

bragaz (commented):
a little step more before shipping this 💪

RiccardoM:
I've recently though more about the `Bech32Address` type, that currently has a `prefix` inside it. What if we simply remove that prefix? The Bech32 specification tells us that the prefix is that part of the address that comes before the first `1` we have. The `1` acts as a separator between the prefix and the actual address (see the specification [here](https://en.bitcoin.it/wiki/BIP_0173)). 

Also, we should make the `Validate` method of both `Base58` and `Bech32` verify that the provided values are actually base58 or bech32-encoded. This can be done like follows: 

```golang
// Verify it's base58 encoded
_, _, err := base58.CheckDecode(data.Value)
if err != nil {
  // address is not base58-encoded
}

// Verify it's bech32 encoded
_, _, err := bech32.Decode(data.Value)
if err != nil {
  // address is not bech32 encoded
}
```

x/profiles/keeper/keeper_chain_link.go
RiccardoM: I've thought more about this @dadamu and I think we should change this approach. Instead, what we can do without having to hard-code the Desmos chain name, is check the value of the address. A Desmos address will always be Bech32 encoded and we will always be able to read it as an `AccAddress`. For this reason, we could define the following method inside the `types` folder: 

```golang
func IsDesmosAddress(data AddressData) bool {
  // 1. Try reading the data as Bech32 address
  bech32data, ok := data.(*Bech32Address)
  if !ok {
    return false
  }
  
  // 2. Try decoding the data as an AccAddress
  _, err := sdk.AccAddressFromBech32(data.Value) 
  return err != nil
}
```

Note: the code needs to be changed to do the cast properly, but it should give you an idea on how to check this

RiccardoM: Thinking more about this approach, we might have some problems with Base58-encoded addresses. I think we might want to have only Bech32-encoded ones to avoid two profiles being linked together 🤔 Otherwise we won't be able to make sure the encoded address is not associated with any profile, and users might be able to trick the system by setting a different chain name or using a different encoding

RiccardoM: Another idea: we simply ditch the check on address and profiles all together and allow any link being created. We should have no problems in doing so since the links are stored inside each profile as a list of addresses. 

Consider we have profile `A` and profile `B`. Then if we link `A -> B` we will see inside `A` the address of `B`, but not the other way around. If we want to see the address of `A` inside `B` then we will have to create another link `B -> A`. 

This way we only keep a one-level link between different accounts and profiles, the same we do with relationships inside the `x/posts` module. What do you guys think @bragaz @dadamu?

dadamu:
@RiccardoM `Prefix` in bech32 is used for identifying if the address is encoded by the certain prefix. Bech32 decode method does not check prefix but it can be done `cosmos-sdk` package [here](https://github.com/cosmos/cosmos-sdk/blob/d14398f849ec184684b22e0dcb61fa3cd33d51b3/types/address.go#L628).

If we remove the `prefix` field, then it would be difficult to recognize `desmosa1234567890` has prefix `desmos` or `desmosa`. With it, I think the prefix should be retained.

x/profiles/keeper/keeper_chain_link.go
dadamu: @RiccardoM True, the Base58-encoded might have some problems cause there is no method to check the validity of address.

Imagine that the user want to create a link with wrong type encoded address `base58address` from `cosmos`. Then, The create link process will success. However, other users will know it is invalid address cause `cosmos` address should be `bech32` encoded and have `cosmos` prefix. As the result, the user only wastes gas fee to create an useless link.

RiccardoM: @dadamu I really think we should avoid any kind of check and let users create the links with any valid Bech32 or Base58-encoded values. If they want to waste some gas and fees creating wrong links it will be their fault. 

We should also delete the check and avoid two profiles being linked together In my opinion. It won't cause any problem and it will simplify a lot the logic of the code.

dadamu: @RiccardoM I agree with it. The multiple profile does not cause any problem. With it, Should `QueryProfileByChainLink` be replaced with `QueryProfilesByChainLink` to show multiple profiles the address is associated?

RiccardoM: @dadamu No, we should always keep one external link associated to a single profile, so the query can stay the same. Using that query, we only return the 1-level links we have, not multiple levels. 

Consider the following situation: 

Untitled Diagram(5)

If we query using the `cosmos` address, we only return the `Desmos 1` profile, not the `Desmos 2`. We only care about the first level links

dadamu: @RiccardoM I got it. Just allow `desmos` address which has profile already to link to another profile. And querying profile by chain link returns linked profile instead of itself. And keep every address has only one link.

x/profiles/keeper/keeper.go
```diff
@@ -55,6 +68,12 @@ func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked string) bool {
 // It assumes that the given profile has already been validated.
 // It returns an error if a profile with the same DTag from a different creator already exists
 func (k Keeper) StoreProfile(ctx sdk.Context, profile *types.Profile) error {
+
+	if _, found := k.GetAccountByChainLink(ctx, types.DesmosChainName, profile.GetAddress().String()); found {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"owner address has already been linked to another profile")
+	}
```
RiccardoM, line 75: We can remove this as we do not care anymore if the destination link already has a profile

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if err := msg.SourceProof.Verify(k.cdc); err != nil {
+		return nil, err
+	}
+
+	if err := msg.DestinationProof.Verify(k.cdc); err != nil {
+		return nil, err
+	}
```
RiccardoM, line 29: Maybe we can move this inside the `StoreChainLink` method? This would avoid us having to duplicate this code inside the `OnReceivPacket` method as well

RiccardoM (approved):

PR #452, opened 2021-05-31
Leonardo/subspaces logic

bragaz:
## Description
This PR adds the `subspaces` logic.
## Checklist
- [x] Wrote unit tests.


x/staging/subspaces/keeper/common_functions.go
```diff
@@ -0,0 +1,69 @@
...
+func (k Keeper) checkSubspaceAndAdmin(ctx sdk.Context, id, address string) (types.Subspace, error) {
+	subspace, found := k.GetSubspace(ctx, id)
+	if !found {
+		return types.Subspace{}, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"the subspace with id %s doesn't exist", id,
+		)
+	}
+
+	if subspace.Owner != address {
+		if !types.IsPresent(subspace.Admins, address) {
```
RiccardoM, line 39: Can we change this to be `subspace.IsAdmin(address)` instead? I think it would be easier to have it as a `Subspace` method instead of a function

x/staging/subspaces/keeper/common_functions.go
```diff
@@ -0,0 +1,69 @@
...
+			return types.Subspace{}, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+				"the user: %s is not the subspace admin or an admin and can't perform this operation on the subspaces: %s",
```
RiccardoM, line 41:
```suggestion
				"%s is not a subspace admin and can't perform this operation on the subspace: %s",
```

x/staging/subspaces/keeper/common_functions.go
```diff
@@ -0,0 +1,69 @@
...
+	subspace, found := k.GetSubspace(ctx, id)
+	if !found {
+		return types.Subspace{}, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"the subspace with id %s doesn't exist", id,
+		)
+	}
+
+	if subspace.Owner != address {
+		if !types.IsPresent(subspace.Admins, address) {
+			return types.Subspace{}, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
```
RiccardoM, line 40: Maybe we should define an error like `ErrInvalidSubspaceAdmin`, so it can be more clear what the problem is. This way we will be able to use it as 

```golang
sdkerrors.Wrap(types.ErrInvalidSubspaceAdmin, address)
```

x/staging/subspaces/keeper/common_functions_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for _, subspace := range subspaces {
+		suite.k.SaveSubspace(suite.ctx, *subspace)
+	}
+
+	var validSubspaces []*types.Subspace
+	suite.k.IterateSubspaces(suite.ctx, func(_ int64, subspace types.Subspace) (stop bool) {
+		if subspace.Name == "bad" {
+			return false
+		}
```
RiccardoM, line 53: Instead of checking the name, can we perform a check on the `index`? I think it would make the test easier to maintain in the future even if the names will change for other reasons

x/staging/subspaces/keeper/common_functions_test.go
```diff
@@ -0,0 +1,62 @@
...
+		validSubspaces = append(validSubspaces, &subspace)
+		return false
+	})
+
+	suite.Len(expSubspaces, len(validSubspaces))
+	for _, subspace := range validSubspaces {
+		suite.Contains(expSubspaces, subspace)
+	}
```
RiccardoM, line 61: Can we simply check for equality between the two slices instead of manually checking them?

x/staging/subspaces/keeper/keeper.go
```diff
@@ -0,0 +1,176 @@
...
+		return types.Subspace{}, false
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	k.cdc.MustUnmarshalBinaryBare(store.Get(types.SubspaceStoreKey(subspaceID)), &subspace)
+	return subspace, true
+}
+
+// GetAllSubspaces returns a list of all the subspaces that have been store inside the given context
+func (k Keeper) GetAllSubspaces(ctx sdk.Context) []types.Subspace {
```
RiccardoM, line 50: Can we move this to a new file named `alias_functions.go`?

x/staging/subspaces/keeper/keeper.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+// SaveSubspace saves the given subspace inside the current context.
+// It assumes that the subspace has been validated already.
+func (k Keeper) SaveSubspace(ctx sdk.Context, subspace types.Subspace) {
+	store := ctx.KVStore(k.storeKey)
+	key := types.SubspaceStoreKey(subspace.ID)
+	store.Set(key, k.cdc.MustMarshalBinaryBare(&subspace))
+}
```
RiccardoM, line 29: I think we should check for some things here. For example, with this code it appears to me that the user can save the subspace even without checking if the editor is valid. I think we should at least check if the editor matches the existing owner at least. 

Implementing this check here would make sure that it is performed also from other modules that might call this method (eg. when implementing IBC capabilities here)

x/staging/subspaces/types/subspace.go
```diff
@@ -92,9 +92,9 @@ func (sub Subspace) Validate() error {
 // SubspaceTypeFromString convert a string in the corresponding SubspaceType
 func SubspaceTypeFromString(subType string) (SubspaceType, error) {
```
RiccardoM, line 93: Is it really necessary? Where is this used?

x/staging/subspaces/keeper/common_functions.go
bragaz: It's a method cause it's generic for both admins and registered/blocked users

x/staging/subspaces/keeper/keeper.go
bragaz: I will rename `common_functions` to `alias_functions` and move it there

x/staging/subspaces/types/subspace.go
bragaz: This will be used inside cli commands to convert the strings passed into the correspondent subspace type

x/staging/subspaces/keeper/keeper.go
bragaz: You're right

bragaz: I'm thinking if it's better to check it outside the save function instead, what do you think?

RiccardoM: @bragaz Checking outside this method would mean we would have to re-implement the checks in every places we use this method, which would cause duplicated code. I think having validity checks directly inside the Save method is a lot better to preserve a more maintainable code and avoid duplication

x/staging/subspaces/types/subspace.go
RiccardoM: There's no need to implement this by hand using constant values. We can simply rely on Protobuf generated maps as Cosmos does [here](https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/types/vote.go#L89)

RiccardoM (commented):
I think we should also review how the `IsValidSubspaceType` is implemented to use a single `if` statement instead of a multiple `switch` one, like Cosmos does [here](https://github.com/cosmos/cosmos-sdk/blob/master/x/gov/types/vote.go#L120)

x/staging/subspaces/types/subspace.go
bragaz: Didn't see that! thanks

bragaz: @RiccardoM I found out that they use also [this function](https://github.com/cosmos/cosmos-sdk/blob/66ee994ce4b5951a373131f0ee5bb0436f507ab4/x/gov/client/utils/utils.go#L10) in conjunction with the one you pointed inside CLI. I don't get the point in using 2 functions instead one. Just getting the value from the first one is not feasible since users should write `SUBSPACE_TYPE_OPEN` inside the tx

RiccardoM: @bragaz They're normalizing the values so that the users do not have to care about casing when writing the option. With that function, `Yes` and `yes` are both valid, while if there was no such function then only `Yes` would be considered valid. 

I think we can implement that function as well to make it easier for users to use the CLI.

Also, we can remove the `SUBSPACE_TYPE_UNSPECIFIED` option as the subspace will always have a known type (either `open` or `closed`).

x/staging/subspaces/keeper/common_functions.go
RiccardoM: @bragaz I think we can make them different methods to be more explicit: 

```golang
func (s *Subspace) IsAdmin(user string) bool {
  return isPresent(s.Admins, user)
}

func (s *Subspace) IsBlocked(user string) bool {
  return isPresent(s.BlockedUsers, user)
}

func (s *Subspace) IsRegistered(user string) bool {
  return isPresent(s.RegisteredUsers, user)
}
```

This way we encapsulate different functionalities in different methods that all rely on the same private one (`isPresent`). We do not expose the underlying implementation but only the functionalities

x/staging/subspaces/types/subspace.go
bragaz: The problem of casing can be solved by using `strings.ToLower()` on the value the the user gives so it will always be something like `open` or `close` when checking.
Also, the `SUBSPACE_TYPE_UNSPECIFIED` is useful when someone tries to edit the subspace and don't want to edit that field, so the default value will be that one and it will tells that the field shouldn't be edited.

RiccardoM: @bragaz I think it's better to follow the Cosmos way here. Otherwise we would have to always remember to call the `string.ToLower` function on the given value. In this case we can simply use the `SubspaceTypeFromString` and do the normalization inside there. Maybe even using `strings.ToLower` inside it would be a good choice. 

Also, I think that instead of using a default value of `SUBSPACE_TYPE_UNSPECIFIED`, we should stick with `[do-not-edit]` for the subspace type as well. This would make it easier for a user to understand since it would have the same behavior of other flags

x/staging/subspaces/keeper/keeper.go
bragaz: @RiccardoM try to take a look at it now, I've made some changes also in the other keeper methods that were using the save. I opted to save it with the store to not repetitively calling the `GetSubspace()` function which is nested inside both save and the checks functions.
Alternatively I can remove the `checkSubspaceAndOwner` function, and only getting the subspace and then let the `SaveSubspace` function handle the checks on the owner...what do you think?  Doing like this will eventually ending in use the `GetSubspace` Method twice.. I'm not convinced

x/staging/subspaces/keeper/msg_server.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, err
+	}
+
+	_ = k.SaveSubspace(ctx, subspace, subspace.Owner)
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
+		types.EventTypeCreateSubspace,
+		sdk.NewAttribute(types.AttributeKeySubspaceID, msg.SubspaceID),
+		sdk.NewAttribute(types.AttributeKeySubspaceName, msg.Name),
+		sdk.NewAttribute(types.AttributeKeySubspaceCreator, msg.Creator),
```
RiccardoM, line 46: We should also add the creation time here since it's only retrievable from the block itself otherwise (which might not be handy for some clients like BDJuno)

RiccardoM (approved):
Looks great to me. Just one very small things to add

PR #459, opened 2021-06-04
Add unit tests for ibc profiles feature

dadamu:
## Description

This PR is adding tests for ibc profiles feature related #192.

## Checklist
- [x] Targeted PR against correct branch.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Wrote unit tests.
- [ ] Wrote integration tests (simulation & CLI). 
- [ ] Updated the documentation. 
- [ ] Added an entry to the `CHANGELOG.md` file.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #463, opened 2021-06-07
IBC profiles implementation

### LINKED ISSUE #192: IBC-enabled social profile

RiccardoM:
## Description
This PR implements the IBC profiles feature. Closes #192 
All props go to @dadamu for the code

## Checklist
- [x] Wrote unit tests.


bragaz (approved):
Previously approved in:
#436 
#449 
#451 
#459

PR #470, opened 2021-06-09
Subspaces checks in `x/posts`

### LINKED ISSUE #464: Add subspaces checks to `posts`
## Feature description 
With the introduction of the `x/subspaces` module it's necessary to add some checks to `x/posts` actions to make sure they're made respecting the constraints that might be applied to a specific user inside a subspace.   
These check affect all the following `x/posts` operations :

- `CreatePost`
- `EditPost`
- `AddPostReaction`
- `RemovePostReaction`
- `RegisterReaction`
- `AnswerPoll`
- `ReportPost`

## Implementation proposal
The `x/posts` module can perform these checks by calling the `x/subspaces` keeper that can be introduced inside the module as an expected keeper.

Any thoughts? Something to add?

bragaz:
## Description
This PR add the checks for subspace inside `x/posts` module.
Closes #464.

## Checklist
- [ ] Wrote unit tests.


bragaz:
@RiccardoM I did not implement the first method of the expected keeper you proposed in the issue because it wasn't used anywhere. Tell me what do you think about this and after it I will make some tests.

x/staging/subspaces/keeper/keeper.go
```diff
@@ -197,3 +197,22 @@ func (k Keeper) UnbanUserInSubspace(ctx sdk.Context, subspaceID, user, admin str
 	store.Delete(types.SubspaceBannedUserKey(subspaceID, user))
 	return nil
 }
+
+// CheckSubspaceUserPermission checks the permission of the given user inside the subspace with the
+// given id to make sure they are able to perform operations inside it
+func (k Keeper) CheckSubspaceUserPermission(ctx sdk.Context, subspaceID string, user string) error {
+	subspace, found := k.GetSubspace(ctx, subspaceID)
+	if !found {
+		return sdkerrors.Wrapf(types.ErrInvalidSubspaceID, "the subspace %s doesn't exist", subspaceID)
```
RiccardoM, line 206: I think we can avoid being more specific. The error should already tell the user what's wrong so we can remove the custom string and just reference the subspace id instead

```suggestion
		return sdkerrors.Wrap(types.ErrInvalidSubspaceID, subspaceID)
```

x/staging/subspaces/keeper/keeper.go
```diff
@@ -197,3 +197,22 @@ func (k Keeper) UnbanUserInSubspace(ctx sdk.Context, subspaceID, user, admin str
...
+	}
+
+	if k.IsBanned(ctx, subspaceID, user) {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "%s is banned from subspace %s", user, subspaceID)
+	}
+
+	if subspace.Type == types.SubspaceTypeClosed && !k.IsRegistered(ctx, subspaceID, user) {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "%s is not registered in subspace %s", user, subspaceID)
+	}
```
RiccardoM, line 215: I think we should define a new error inside the subspace to handle this cases. The error might be called something like `ErrPermissionDenied` so that it can be applied to both unregistered and banned users.

Then we will be able to change these lines with 

```suggestion
	if k.IsBanned(ctx, subspaceID, user) {
		return sdkerrors.Wrapf(types.ErrPermissionDenied, "%s is banned", user)
	}

	if subspace.Type == types.SubspaceTypeClosed && !k.IsRegistered(ctx, subspaceID, user) {
		return sdkerrors.Wrapf(types.ErrPermissionDenied, "%s is not registered", user)
	}
```

(No need to specify the subspace id each time)

RiccardoM (approved):
@bragaz Looks great, please proceed with the tests in this same branch

x/staging/posts/keeper/msgs_server_test.go
```diff
@@ -198,6 +198,8 @@ func (suite *KeeperTestSuite) TestMsgServer_CreatePost() {
 			suite.SetupTest()
 			suite.k.SetParams(suite.ctx, types.DefaultParams())
 
+			_ = suite.sk.SaveSubspace(suite.ctx, suite.testData.subspace, suite.testData.postOwner)
```
RiccardoM, line 201: Can we not ignore the error please? I think it's always worth to not ignore them: 

```suggestion
			err := suite.sk.SaveSubspace(suite.ctx, suite.testData.subspace, suite.testData.postOwner)
			suite.Require().NotErr(err)
```

This also applies to all other places

x/staging/posts/simulation/operations_posts.go
```diff
@@ -91,10 +92,19 @@ func sendMsgCreatePost(
...
-	r *rand.Rand, ctx sdk.Context, accs []simtypes.Account, k keeper.Keeper, ak authkeeper.AccountKeeper,
+	r *rand.Rand, ctx sdk.Context, accs []simtypes.Account, k keeper.Keeper, ak authkeeper.AccountKeeper, sk subspaceskeeper.Keeper,
 ) (*PostData, bool) {
 
 	postData := RandomPostData(r, accs)
+	subspaceData := RandomSubspace(r, accs)
+
+	if err := sk.SaveSubspace(ctx, subspaceData.Subspace, subspaceData.Subspace.Creator); err != nil {
+		return nil, true
+	}
```
RiccardoM, line 103: We should change how the subspace is retrieved here and in all the other places of the simulation operations as well

x/staging/posts/simulation/operations_reports.go
```diff
@@ -86,14 +88,23 @@ func sendMsgReportPost(
...
 
+	subspace, _ := subspacessims.RandomSubspace(r, sk.GetAllSubspaces(ctx))
+
 	reportsData := RandomReportsData(r, posts, accs)
+	post, found := pk.GetPost(ctx, reportsData.PostID)
+	if !found {
+		return nil, true
+	}
+	post.Subspace = subspace.ID
+	pk.SavePost(ctx, post)
```
RiccardoM, line 106: I don't get it why we need to continuously change the post subspace before performing any operations on it. What happens if we remove this?

bragaz: To ensure we won't have any errors from a non-existent subspace, do you think we can remove this?

RiccardoM: @bragaz As far as I remember, posts are created in two occasions: 

1. Inside the genesis 
2. Using `MsgCreatePost` 

What we can do is making sure that while initializing them inside those places, we subspace id is taken randomly from the slice of existing subspaces. This would make sure that we never have to check for such errors. 

So, in summary, we should: 

1. Generate a random slice of genesis subspaces. 
2. Generate a random slice of genesis posts, each one having subspace id randomly taken from the genesis subspaces. 
3. When we perform a `MsgCreatePost`, we get all the existing subspaces, extract one random subspace and get the id. Then we use that ID as the post subspace id

x/staging/posts/simulation/operations_posts.go
```diff
@@ -95,6 +95,7 @@ func randomPostCreateFields(
 ) (*PostData, bool) {
 
 	postData := RandomPostData(r, accs)
+
```
RiccardoM, line 98:
```suggestion
```

RiccardoM (approved):

PR #486, opened 2021-06-14
Added logger and logs to modules

### LINKED ISSUE #483: Improve modules logging
## Context
Currently the SDK modules are logging useful information using the `Logger` interface. This allows the node to print different levels of information when running (eg. `debug`, `info`, etc). 

To make it easier for node operators to understand how their node is doing, we should implement logging into our modules as well when needed. This can be done by defining a new method for each keeper as follows: 

```golang
func Logger() log.Logger {
  return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
}
```

Then, we can simply call the various methods (`Info`, `Debug`, ...):

```golang
logger.Info("asserted all invariants", "duration", diff, "height", ctx.BlockHeight())
```

bragaz:
## Description
This PR added the logger to modules' keeper. (And also logs in keeper methods)
Closes #483.
## Checklist
- [ ] Wrote unit tests.


bragaz:
Any other useful logs that you think could be handy?

x/profiles/keeper/keeper_dtag_transfers.go
```diff
@@ -25,6 +25,8 @@ func (k Keeper) SaveDTagTransferRequest(ctx sdk.Context, request types.DTagTrans
 
 	requests = types.NewDTagTransferRequests(append(requests.Requests, request))
 	store.Set(key, k.cdc.MustMarshalBinaryBare(&requests))
+
+	k.Logger(ctx).Info("dTag transfer request", "receiver", request.Receiver, "from", request.Sender)
```
RiccardoM, line 29: I think it's better to use the conventional `DTag` here

```suggestion
	k.Logger(ctx).Info("DTag transfer request", "receiver", request.Receiver, "from", request.Sender)
```

x/staging/subspaces/keeper/keeper.go
```diff
@@ -32,12 +38,15 @@ func (k Keeper) SaveSubspace(ctx sdk.Context, subspace types.Subspace, user stri
 
 	// Check the editor when the user is trying to edit the subspace
 	if found && storedSubspace.Owner != user {
+		k.Logger(ctx).Error("unauthorized user tried to edit subspace", "id", subspace.ID, "user", user)
```
RiccardoM, line 41: I think we can remove the logging of these kind of errors

```suggestion
```

x/staging/fees/keeper/keeper.go
```diff
@@ -56,6 +62,7 @@ func (k Keeper) CheckFees(ctx sdk.Context, fees sdk.Coins, msgs []sdk.Msg) error
 	}
 
 	if !requiredFees.IsZero() && (requiredFees.IsAnyGT(fees) || fees.IsZero()) {
+		k.Logger(ctx).Error("required fees not satisfied", "required", requiredFees, "given", fees)
```
RiccardoM, line 65: I think it's better to avoid logging these kind of errors since they will be included inside the transaction result anyway. By using the `Error` method the message will be displayed inside full nodes, which might alert node operators for nothing. We should use that only to log errors that might cause the node to crash

```suggestion
```

RiccardoM (commented):
Overall looks great. I would just edit some small things but then we're ready to go.
Also it's probably worth adding a CHANGELOG entry since this is something that is going to be visible by node operators

RiccardoM (commented):
@bragaz Can you please add a CHANGELOG entry since it will be something that node operators are going to notice?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Added the ability to paginate the registered reactions ([#471](https://github.com/desmos-labs/desmos/issues/471))
 - Removed all the legacy queriers ([#496](https://github.com/desmos-labs/desmos/issues/496))
 - Changed how chain links are stored ([#480](https://github.com/desmos-labs/desmos/issues/480))
+- Added logger and logs to modules' keepers ([#483](https://github.com/desmos-labs/desmos/issues/483))
```
RiccardoM, line 11:
```suggestion
- Added the logging of some interesting operations ([#483](https://github.com/desmos-labs/desmos/issues/483))
```

RiccardoM (approved):

Issue #492, opened 2021-06-16
Make DTag case insensitive


RiccardoM:
## Context
Currently `DTag -> Profile` associations are stored using the following key: 

```golang
func DTagStoreKey(dTag string) []byte {
	return append(DTagPrefix, []byte(dTag)...)
}
```

While this works, it has a very serious problem: it considers DTag as **case sensitive**. 

This means that currently it's possible to have multiple accounts with very similar DTags such as
- `RiccardoMontagnin`
- `riccardomontagnin`
- `riccardoMontagnin`
- `rIccardoMontagnin`
- ...

This might cause very serious security problems in the future due to the possible impersonification. 

## Solution
In order to solve this, we should make a DTag case-insensitive. This can be done by easily changing how the key is generated to be: 

```go
func DTagStoreKey(dTag string) []byte {
	return append(DTagPrefix, []byte(strings.ToLower(dTag))...)
}
```

This will convert the DTag to a lower case and avoid any duplication due to different casing. 

### Note
Tests should be added to make sure that two DTags with different casing cannot exist at the same time.

PR #494, opened 2021-06-16
Case insensitive DTag

bragaz:
## Description
This PR makes the `profiles` DTag case insensitive.
#492.
 
## Checklist
- [ ] Wrote unit tests.


RiccardoM:
@bragaz Sims tests are failing, better check out why

RiccardoM (commented):
Looks great. Can you just add a unit test to make sure that trying to save a profile with a specific DTag does not care about casing? What I mean is: 

1. Store an existing profile with DTag `test` and try storing a new profile from another user with DTag `Test`. This should error. 
2. Store an existing profile with DTag `test` and try updating it using DTag `Test` using the same account. This should not error and update the data correctly.

x/profiles/simulation/operations_dtag_transfers.go
```diff
@@ -120,7 +120,10 @@ func randomDTagRequestTransferFields(
 	if err != nil {
 		return simtypes.Account{}, types.DTagTransferRequest{}, true
 	}
-	_ = k.StoreProfile(ctx, receiverProfile)
+	err = k.StoreProfile(ctx, receiverProfile)
+	if err != nil {
+		return simtypes.Account{}, types.DTagTransferRequest{}, true
+	}
```
RiccardoM, line 126: I think this should not have been edited. I checked the failing sim test before editing this and the error was: 
```
failed to execute message; message index: 0: the request recipient does not have a profile yet: invalid request [/home/riccardo/Coding/Forbole/Desmos/x/profiles/keeper/msg_server_dtag_transfers.go:27]
        Comment: MsgRequestDTagTransfer
```

This probably means that something about storing or getting a profile using a DTag is not working correctly with casing. We should add more unit tests to see why this is not working.

bragaz:
@RiccardoM I've added two more tests as you asked to ensure the fact that DTag casing is working properly and check also the error of the sim tests debugging it:
Schermata 2021-06-18 alle 10 03 20

Considering the error I think it's worth to add that check on `StoreProfile` right?

RiccardoM:
> Considering the error I think it's worth to add that check on `StoreProfile` right?

I think this error started showing up now because case insensitive DTags actually work. If we take a look at the test code, what we do is: 
1. We get the user address and see if it has a profile 
2. If it does not have a profile, we store a new one with a random DTag

Previously, if the random DTag did not match only because of casing, nothing would happen. But now, since we have case insensitive DTags, this means that if the random DTag is equals to the one of another profile that error will show up. 

For this reason, I think we should review how simulation tests deal with profiles. Instead of getting a random recipient, and then creating a profile for such recipient, what we can do instead is getting a random profile. This way we do not have to generate a new one cause we will always use an already existing one. The only thing we need to check is to make sure that the recipient and message signer are not the same user.

RiccardoM:
@bragaz Can you please add a CHANGELOG entry since this change is going to break the consensus state please? Then we're ready to merge this

bragaz:
@RiccardoM done

RiccardoM (approved):

PR #520, opened 2021-06-25
Allow to paginate DTag transfer requests

### LINKED ISSUE #519: Add paginated `DTagTransferRequests` query method
In the context of #448, we should implement the pagination for incoming DTag transfer requests for a user

RiccardoM:
## Description
This PR changes how DTag transfer requests are stored, allowing to paginate them when querying. 
Closes #519

## Checklist
- [x] Wrote unit tests.


proto/desmos/profiles/v1beta1/query_dtag_requests.proto
```diff
@@ -10,20 +10,27 @@ import "cosmos/base/query/v1beta1/pagination.proto";
 
 option go_package = "github.com/desmos-labs/desmos/x/profiles/types";
 
-// QueryDTagTransfersRequest is the request type for the Query/DTagTransfers RPC
-// endpoint
-message QueryDTagTransfersRequest {
+// QueryIncomingDTagTransferRequests is the request type for the
+// Query/IncomingDTagTransferRequests RPC endpoint
+message QueryIncomingDTagTransferRequestsRequest {
```
dadamu, line 15:
```suggestion
// QueryIncomingDTagTransferRequestsRequest is the request type for the
// Query/IncomingDTagTransferRequests RPC endpoint
message QueryIncomingDTagTransferRequestsRequest {
```

x/profiles/keeper/alias_functions_test.go
```diff
@@ -97,6 +97,39 @@ func (suite *KeeperTestSuite) TestKeeper_GetProfiles() {
...
+
+	for _, request := range requests {
+		err := suite.k.SaveDTagTransferRequest(suite.ctx, request)
+		suite.Require().NoError(err)
+	}
+
+	iterations := 0
+	suite.k.IterateUserIncomingDTagTransferRequests(suite.ctx, address, func(index int64, request types.DTagTransferRequest) (stop bool) {
+		iterations += 1
+		return index == 1
```
dadamu, line 128: Could it simply return false, instead?

x/profiles/keeper/grpc_query.go
```diff
@@ -58,17 +58,37 @@ func (k Keeper) Profile(ctx context.Context, request *types.QueryProfileRequest)
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid user address: %s", request.Receiver)
+	}
+
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	var requests []types.DTagTransferRequest
+
+	// Get user requests prefix store
+	store := sdkCtx.KVStore(k.storeKey)
+	relsStore := prefix.NewStore(store, types.IncomingDTagTransferRequestsPrefix(request.Receiver))
```
dadamu, line 72: I think it should be requestsStore here.

x/profiles/keeper/alias_functions_test.go
RiccardoM: Fixed the test to make sure it stops earlier

dadamu: It seems that the test would be passed if the actual result is over 2 cause it would be stopped earlier on 2.

RiccardoM: @dadamu Yeah I've made it so to make sure that both cases of the `if !stop` are tested: when `stop == true` and when `stop == false`. The purpose of the test is to make sure that it iterates over only 2 out of the 3 transfer requests accordingly

dadamu: @RiccardoM I got it, it is used for testing stop quickly. Thanks for the answer!

dadamu (approved):
The whole implementation is looks good to me!

Issue #493, opened 2021-06-16
Implement the pagination to `Report`


dadamu:
## Feature description 

This issue is for the `Report` part's implementation of #448.

In the current version, all the reports of the post are store in one key, `ReportPrefix + '<post-id>'`. In addition, A user can report a post repeatedly unless the report content are absolutely same as existing report. 
It could be split to a report to be associated to only one key associated to user so that it can prevent multiple times reports from a user and also can be paginated when querying.

## Implementation proposal

The key can be changed to `ReportPrefix + <post-id> + <reporter-address>`.
This allows:
1. Prevent a user report a post repeatedly
2. Query all the reports for the specific post by iterating over `ReportsPrefix + '<post-id>'`

What do you think? @RiccardoM @bragaz

bragaz:
I agree with you @dadamu. I'm also thinking about some improvements for the `Reports` object, I will open another issue soon

RiccardoM:
@dadamu I agree we should allow only one report per (post - user) tuple. However, as @bragaz said, we should also review the `Report` object implementation. I'm thinking for example to allow multiple reasons to be selected for the report (eg. nudity + other things). I will come more on the issue @bragaz is going to open thought.

RiccardoM:
Closing this as it will be implemented when we re-implement the `x/reports` module

PR #522, opened 2021-06-25
Report improvements

### LINKED ISSUE #507: Improve `Report` type
## Feature description 
Some days ago I was looking how CSN handle reports and I've noticed that they all let users select a category (or more then one) from a list of existent categories. I think it should be good to standardise this thing on Desmos.
Currently we only have a generic `type` field (which name is IMO a little bit confusing) that can be any string.
In the long, when there will be a variety of apps using Desmos this will result in a lot of confusions with possibly tons of different categories. 

## Implementation proposal
To improve it, I'm actually thinking about something like this:
```proto
enum Reason {
  option (gogoproto.goproto_enum_prefix) = false;

  NUDITY = 0;
  VIOLENCE = 1;
  INTIMIDATION = 2;
  HARASSMENT = 3;
  SUICIDE = 4;
  SELF_HARM = 5;
  FAKE_INFORMATION = 6;
  SPAM = 7;
  UNAUTHORIZED_SALES = 8;
  TERRORISM = 9;
  OTHERS = 10;
}

// Report is the struct of a post's reports
message Report {
  option (gogoproto.goproto_stringer) = true;

  // ID of the post for which the report has been created
  string post_id = 1 [
    (gogoproto.customname) = "PostID",
    (gogoproto.jsontag) = "post_id",
    (gogoproto.moretags) = "yaml:\"post_id\""
  ];

  // Identifies the type of the reports
  repeated Reason reasons = 2
      [ (gogoproto.jsontag) = "reasons", (gogoproto.moretags) = "yaml:\"reasons\"" ];

  // User message
  string message = 3 [
    (gogoproto.jsontag) = "message",
    (gogoproto.moretags) = "yaml:\"message\""
  ];

  // Identifies the reporting user
  string user = 4
      [ (gogoproto.jsontag) = "user", (gogoproto.moretags) = "yaml:\"user\"" ];
}
```

I'm just giving an example here and maybe there are better ways to handle this, just pointing this out.
Thoughts? @RiccardoM @dadamu

bragaz:
## Description
This PR improve the `Report` type.
Close #507.
## Checklist
- [ ] Wrote unit tests.


x/staging/posts/client/cli/tx.go
```diff
@@ -351,13 +351,13 @@ E.g.
...
 	cmd := &cobra.Command{
-		Use:   "report [post-id] [reports-type] [reports-message]",
+		Use:   "report [post-id] [reports-message] [reports-types...]",
 		Short: "reports a post",
 		Long: fmt.Sprintf(`
 Report an existent post specifying its ID, the reports's type and message.
 
 E.g.
-%s tx posts report a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc scam "this post is a scam" 
+%s tx posts report a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc "this post is a scam"  "scam,nudity"
```
dadamu, line 360:
```suggestion
%s tx posts report a4469741bb0c0622627810082a5f2e4e54fbbb888f25a4771a5eebc697d30cfc "this post is a scam" "scam,nudity" 
```

x/staging/posts/types/params_test.go
```diff
@@ -183,3 +188,35 @@ func TestValidateMaxAdditionalAttributesFieldKeyLengthParam(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateReportTypesParam(t *testing.T) {
+	tests := []struct {
+		name        string
+		reportTypes interface{}
```
dadamu, line 195: Could it be changed to []string, instead?

bragaz: @dadamu Nope. We use interface to ensure the test fails if we have a different type then the expected one

bragaz:
I will also add #493 to this PR. @dadamu I will re-request the review once completed it

bragaz:
added pagination

x/staging/posts/keeper/grpc_query_test.go
```diff
@@ -457,8 +457,8 @@ func (suite *KeeperTestSuite) Test_UserAnswers() {
 		},
 	}
 
-	suite.SetupTest()
 	for _, uc := range usecases {
+		suite.SetupTest()
```
dadamu, line 461: The cache context is used in the test. I think `SetupTest()` could be outside to make sure initialization.

x/staging/posts/types/keys.go
```diff
@@ -113,7 +113,12 @@ func UserAnswersStoreKey(id, user string) []byte {
 	return append(UserAnswersByPostPrefix(id), []byte(user)...)
 }
 
-// ReportStoreKey turns an id into a key used to store a report inside the reports store
-func ReportStoreKey(id string) []byte {
+// ReportsByPostIDPrefix turns an id into a key used to store a report inside the reports store
```
dadamu, line 116:
```suggestion
// ReportsByPostIDPrefix returns a prefix used to store all the reports for the post having given id```

File: x/staging/posts/types/report.go [hunk 5]
```diff
@@ -8,23 +8,40 @@ import (
...
 func (r Report) Validate() error {
 	if !IsValidPostID(r.PostID) {
 		return fmt.Errorf("invalid post id: %s", r.PostID)
 	}
 
-	if len(strings.TrimSpace(r.Type)) == 0 {
-		return fmt.Errorf("report type cannot be empty")
+	for _, reason := range r.Reasons {
+		if strings.TrimSpace(reason) == "" {
+			return fmt.Errorf("report reason cannot be empty")
```
dadamu, line 43:
```suggestion
			return fmt.Errorf("report reason cannot be empty or blank")
```

x/staging/posts/types/report.go
```diff
@@ -8,23 +8,40 @@ import (
...
+		}
+	}
+	return true
+}
+
 // Validate implements validator
 func (r Report) Validate() error {
 	if !IsValidPostID(r.PostID) {
 		return fmt.Errorf("invalid post id: %s", r.PostID)
 	}
```
dadamu, line 40: I think here should check if the reasons is empty or not.

```golang
if len(r.Reasons) == 0 {
    return fmt.Errorof("report reasons cannot be empty")
}
```

x/staging/posts/types/report_test.go
```diff
@@ -2,14 +2,53 @@ package types_test
...
 
+func TestReport_AreReasonsValid(t *testing.T) {
+	useCases := []struct {
+		name          string
+		report        types.Report
+		paramsReports []string
+		expBool       bool
+	}{
+		{
+			name: "not contained reports reasons return false",
```
dadamu, line 21:
```suggestion
			name: "not contained reports reasons returns false",
```

x/staging/posts/keeper/keeper_reports_test.go
```diff
@@ -15,38 +66,43 @@ func (suite *KeeperTestSuite) TestKeeper_SaveReport() {
...
 			storedReports: []types.Report{
-				types.NewReport("post_id", "type", "message", "user"),
+				types.NewReport("9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", []string{"scam"}, "message", "user"),
 			},
-			report: types.NewReport("post_id", "type", "message", "user"),
+			report: types.NewReport("9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08", []string{"scam"}, "message", "user"),
+			expErr: true,
+		},
+		{
+			name:   "trying  to store invalid report returns error",
```
dadamu, line 96:
```suggestion
			name:   "trying to store invalid report returns error",
```

x/staging/posts/keeper/grpc_query_test.go
```diff
@@ -691,3 +691,98 @@ func (suite *KeeperTestSuite) Test_PostComments() {
...
+		shouldErr bool
+		expLen    int
+	}{
+		{
+			name:      "invalid post id returns error",
+			req:       &types.QueryReportsRequest{},
+			shouldErr: true,
+		},
+		{
+			name:      "non existent post return error",
```
dadamu, line 737:
```suggestion
			name:      "non existent post returns error",
```

x/staging/posts/keeper/grpc_query.go
```diff
@@ -125,11 +125,37 @@ func (k Keeper) RegisteredReactions(goCtx context.Context, req *types.QueryRegis
...
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "post with id %s not found", request.PostId)
+	}
+
+	store := ctx.KVStore(k.storeKey)
+
+	reportsStore := prefix.NewStore(store, types.ReportsByPostIDPrefix(request.PostId))
+
+	var reports []types.Report
+
+	pageRes, err := query.FilteredPaginate(reportsStore, request.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
```
dadamu, line 146:
```suggestion
	store := ctx.KVStore(k.storeKey)
	reportsStore := prefix.NewStore(store, types.ReportsByPostIDPrefix(request.PostId))

	var reports []types.Report
	pageRes, err := query.FilteredPaginate(reportsStore, request.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
```

dadamu (changes requested):
It misses adding the pagination flag to the cli query method.

x/staging/posts/keeper/grpc_query_test.go
bragaz: Oh this is unwanted, reverted it now

x/staging/posts/client/cli/query.go
```diff
@@ -219,7 +219,9 @@ func GetCmdQueryReports() *cobra.Command {
 			return clientCtx.PrintProto(res)
```
dadamu, line 219: The request of query cmd should add pagination as well. such like:
```go
pageReq, err := client.ReadPageRequest(cmd.Flags())
	if err != nil {
		return err
	}

	res, err := queryClient.Reports(
		context.Background(),
		&types.QueryReportsRequest{PostId: args[0], Pagination: pageReq},
	)
```

x/staging/posts/types/keys.go
```diff
@@ -108,12 +108,17 @@ func UserAnswersByPostPrefix(id string) []byte {
...
 
-// UserAnswersStoreKey returns the store key used to store the user answer containing the given data
+// UserAnswersStoreKey returns the prefix used to store all the reports for the post having the given id
 func UserAnswersStoreKey(id, user string) []byte {
 	return append(UserAnswersByPostPrefix(id), []byte(user)...)
 }
 
-// ReportStoreKey turns an id into a key used to store a report inside the reports store
-func ReportStoreKey(id string) []byte {
+// ReportsByPostIDPrefix turns an id into a key used to store a report inside the reports store
```
dadamu, line 116:
```suggestion
// ReportsByPostIDPrefix returns a prefix used to store all the reports for the post having given id
```
I think the function returns a prefix, not a key.

dadamu (changes requested):
The overall looks good to me, just pageRequest in query cmd should be solved.

x/staging/posts/client/cli/query.go
bragaz: whoops! Forgot it, Thanks! @dadamu

dadamu (approved):
Great, LGTM!

proto/desmos/posts/v1beta1/msgs.proto
```diff
@@ -220,7 +220,7 @@ message MsgReportPost {
     (gogoproto.moretags) = "yaml:\"post_id\""
   ];
 
-  string report_type = 2 [ (gogoproto.moretags) = "yaml:\"type\"" ];
+  repeated string report_reasons = 2 [ (gogoproto.moretags) = "yaml:\"reasons\"" ];
```
RiccardoM, line 223: Can we rename this to simply be `reasons`?

proto/desmos/posts/v1beta1/params.proto
```diff
@@ -28,4 +28,8 @@ message Params {
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"max_additional_attributes_field_key_length\""
   ];
+
+  repeated string report_reasons = 5 [
```
RiccardoM, line 32: I think this can be renamed to be `allowed_reasons` instead

x/staging/posts/client/cli/tx.go
```diff
@@ -351,13 +351,13 @@ E.g.
 // GetCmdReportPost returns the command allowing to report a post
 func GetCmdReportPost() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "report [post-id] [reports-type] [reports-message]",
+		Use:   "report [post-id] [reports-message] [reports-types...]",
```
RiccardoM, line 354: I don't link this because it makes it mandatory to set a report message, which should not. I think we can use commands flags to set the report reasons making the second parameter (the message) optional instead

x/staging/posts/keeper/common_functions.go
```diff
@@ -181,3 +181,37 @@ func (k Keeper) IteratePostReactionsByPost(ctx sdk.Context, postID string, fn fu
 		i++
 	}
 }
+
+// IteratePostReportsByPostID iterates through the post's reports with the given id and performs the provided function
+func (k Keeper) IteratePostReportsByPostID(ctx sdk.Context, postID string, fn func(index int64, report types.Report) (stop bool)) {
```
RiccardoM, line 186: I think this can be renamed to simply be `IteratePostReports`

x/staging/posts/keeper/common_test.go
```diff
@@ -230,6 +230,7 @@ func (suite *KeeperTestSuite) SetupTest() {
 		"4e188d9c17150037d5199bbdb91ae1eb2a78a15aca04cb35530cccb81494b36e",
 	)
 
+	suite.k.SetParams(suite.ctx, types.DefaultParams())
```
RiccardoM, line 233: If needed, I think this should be set inside each test requiring it

docs/developers/types/staging/posts/report.md
```diff
@@ -2,10 +2,10 @@
 Inside Desmos, users can flag posts with reports.
 
 ## Contained data
-Reports contains data that allows users to create the most complete report they need.
+Report contains data that allows users to create the most complete report they need.
 
-### `Type`
-The type of the report which correspond to one of the reason showed in the list above.
+### `Reasons`
+The reasons of the report which correspond to one or more reasons saved in chain params.
```
RiccardoM, line 8:
```suggestion
The reasons of the report which correspond to one or more reasons saved as the module params.
```

RiccardoM (commented):
There are some conflicts. Once those are fixed we can merge it

RiccardoM:
@bragaz Looks like there are lint and test errors. Would you mind fixing those?

bragaz:
@RiccardoM fixed. Docs error are for desmos rpc link

x/staging/posts/client/cli/cli_test.go
```diff
@@ -838,20 +838,37 @@ func (s *IntegrationTestSuite) TestCmdReportPost() {
...
 				"message",
+				fmt.Sprintf("--%s=%s", cli.FlagReason, "scam"),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, val.Address.String()),
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(10))).String()),
+			},
+			respType: &sdk.TxResponse{},
+		},
+		{
```
RiccardoM, line 866: Can we add a test to make sure that using multiple times the `--reason` flag works properly?

x/staging/posts/client/cli/tx.go
```diff
@@ -351,22 +351,32 @@ E.g.
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
 				return err
 			}
 
-			msg := types.NewMsgReportPost(args[0], args[1], args[2], clientCtx.FromAddress.String())
+			reasons, _ := cmd.Flags().GetStringSlice(FlagReason)
+			if len(reasons) == 0 {
+				return fmt.Errorf("report reason should be provided and not be empty")
```
RiccardoM, line 371:
```suggestion
				return fmt.Errorf("at least one reason should be provided and not be empty")
```

x/staging/posts/client/cli/tx.go
```diff
@@ -375,6 +385,7 @@ E.g.
 		},
 	}
 
+	cmd.Flags().StringSlice(FlagReason, []string{""}, "Reasons for the report")
```
RiccardoM, line 388: Can we simply specify an empty array here? 

```suggestion
	cmd.Flags().StringSlice(FlagReason, nil, "Reasons for the report")
```

x/staging/posts/keeper/common_functions.go
```diff
@@ -181,3 +181,37 @@ func (k Keeper) IteratePostReactionsByPost(ctx sdk.Context, postID string, fn fu
 		i++
 	}
 }
+
+// IteratePostReportsByPost iterates through the post's reports with the given id and performs the provided function
+func (k Keeper) IteratePostReportsByPost(ctx sdk.Context, postID string, fn func(index int64, report types.Report) (stop bool)) {
```
RiccardoM, line 186: Maybe we should rename this to simply be `IteratePostReports`?

x/staging/posts/keeper/keeper_reports.go
```diff
@@ -32,22 +45,20 @@ func (k Keeper) SaveReport(ctx sdk.Context, report types.Report) error {
...
 func (k Keeper) GetPostReports(ctx sdk.Context, postID string) []types.Report {
-	store := ctx.KVStore(k.storeKey)
-	return types.MustUnmarshalReports(store.Get(types.ReportStoreKey(postID)), k.cdc)
+	var reports []types.Report
+	k.IteratePostReportsByPost(ctx, postID, func(_ int64, report types.Report) bool {
+		reports = append(reports, report)
+		return false
+	})
+	return reports
 }
```
RiccardoM, line 54: I think this should be moved inside the `alias_functions` file together with similar methods

x/staging/posts/keeper/keeper_reports.go
```diff
@@ -32,22 +45,20 @@ func (k Keeper) SaveReport(ctx sdk.Context, report types.Report) error {
...
-		postReports := types.MustUnmarshalReports(iterator.Value(), k.cdc)
-		reports = append(reports, postReports...)
-	}
-
+	k.IterateReports(ctx, func(_ int64, report types.Report) bool {
+		reports = append(reports, report)
+		return false
+	})
 	return reports
 }
```
RiccardoM, line 64: Same as above

x/staging/posts/keeper/keeper_reports_test.go
```diff
@@ -4,6 +4,58 @@ import (
 	"github.com/desmos-labs/desmos/x/staging/posts/types"
 )
 
+func (suite *KeeperTestSuite) TestKeeper_CheckReportValidity() {
+	tests := []struct {
+		name      string
+		report    types.Report
+		shouldErr bool
+	}{
+		{
```
RiccardoM, line 13: Can we add a test to make sure that if no reasons are specified, an error is returned as well?

x/staging/posts/types/report.go
```diff
@@ -8,27 +8,48 @@ import (
...
+		exists[reason] = true
+	}
+
+	for _, rr := range r.Reasons {
+		if !exists[rr] {
+			return false
+		}
+	}
+	return true
+}
```
RiccardoM, line 33: Probably this can be moved as a keeper method instead? Something like a `k.ValidateReport` method similar to `k.ValidateProfile`?

RiccardoM (commented):
I noticed that currently if a user tries reporting two times the same port, it returns an error. 
This makes it impossible for the user to edit the existing report. We should further investigate this and see how we should handle a report edit and also deletion.

x/staging/posts/keeper/keeper_reports.go
bragaz: I think we should leave it here since it's a reports' related method

x/staging/posts/types/report.go
bragaz: There's already the `CheckReportValidity` keeper method that pack this method and the `Validate()` one.
I think that should be enough, otherwise we should move both in keeper.

bragaz:
>I noticed that currently if a user tries reporting two times the same port, it returns an error.
This makes it impossible for the user to edit the existing report. We should further investigate this and see how we should handle a report edit and also deletion.

@RiccardoM What about handling this in a separate PR? Otherwise this one will become bigger and harder to review

x/staging/posts/types/report.go
RiccardoM: The I think it's better to move this logic inside the `CheckReportValidity` method itself instead of having it here. This would make sure that we do not need to remember to call two different methods to validate a report.

RiccardoM:
> > I noticed that currently if a user tries reporting two times the same port, it returns an error.
> > This makes it impossible for the user to edit the existing report. We should further investigate this and see how we should handle a report edit and also deletion.
> 
> @RiccardoM What about handling this in a separate PR? Otherwise this one will become bigger and harder to review

This is ok to me, but I would avoid merging this before that one is merged into this as well. Having multiple PRs working on very similar changes will make it harder if we have to cherry pick those changes in the future. So I think it's better to open a new branch and a PR from that branch into this one. Then merging this one into the `master` once everything is completed.

RiccardoM (approved):

PR #523, opened 2021-06-28
Improvements

RiccardoM:
## Description
This PR tries to fix the simulation tests that are currently failing inside the `master` branch. While fixing those, I realized there were a lot of things missing, so here is what I did: 

- Added missing validity checks when storing DTag transfer requests, user blocks, relationships, chain links and application links
- Added missing invariants for DTag transfer requests, user blocks, relationships, chain links and application links

## Checklist
- [x] Wrote unit tests.


x/profiles/keeper/grpc_query_test.go
```diff
@@ -514,11 +532,20 @@ func (suite *KeeperTestSuite) Test_UserBlocks() {
...
+			profile := suite.CreateProfileFromAddress(suite.testData.user)
+			otherProfile := suite.CreateProfileFromAddress(suite.testData.otherUser)
+
+			err := suite.k.StoreProfile(suite.ctx, profile)
+			suite.Require().NoError(err)
+
+			err = suite.k.StoreProfile(suite.ctx, otherProfile)
+			suite.Require().NoError(err)
 
 			for _, UserBlock := range uc.storedUserBlocks {
```
dadamu, line 546: I think here should be lowercase naming.

RiccardoM: Fixed

x/profiles/keeper/msgs_server_profile.go
```diff
@@ -45,7 +45,7 @@ func (k msgServer) SaveProfile(goCtx context.Context, msg *types.MsgSaveProfile)
 
 	// If the DTag changes, delete all the previous DTag transfer requests
 	if profile.DTag != msg.DTag {
-		k.DeleteAllDTagTransferRequests(ctx, msg.Creator)
+		k.DeleteAllUserIncomingDTagTransferRequests(ctx, msg.Creator)
 	}
```
dadamu, line 49: I think this part should be moved the last part of the function to ensure that no error happens after this part.

bragaz (approved):

x/profiles/keeper/alias_functions.go
```diff
@@ -33,8 +33,36 @@ func (k Keeper) GetProfiles(ctx sdk.Context) []*types.Profile {
 	return profiles
 }
 
+// HasProfile returns true iff the given user has a profile, or an error if something is wrong.
```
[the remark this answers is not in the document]
RiccardoM: "Iff" is actually an abbreviation used in logic for _If and only if_ ([Wikipedia](https://en.wikipedia.org/wiki/If_and_only_if)). I think it's better to leave it as it is

x/profiles/simulation/genesis.go
```diff
@@ -90,23 +173,44 @@ func containsRelationship(slice []types.Relationship, relationship types.Relatio
...
+		if !containsUserBlock(usersBlocks, block) {
+			usersBlocks[index] = block
+			index++
 		}
 	}
 
 	return usersBlocks
 }
+
+// containsUserBlock returns true iff the given slice contains the given block
```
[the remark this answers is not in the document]
RiccardoM: As above

x/profiles/keeper/msgs_server_profile.go
RiccardoM: Thanks for the tip. I actually ended up removing this entirely and moving it inside the `StoreProfile` method instead. This should make sure that if such method is called from anywhere else, the DTag requests will be deleted properly.

PR #529, opened 2021-07-01
Enable snapshot by default

### LINKED ISSUE #469: Enable snapshot in app.toml by default
## Feature description 

From my experience in many different chains developed with `Stargate`, most of the validators or fullnodes would not have snapshot enabled. This make state sync a new node very difficult. I suggest we make snapshot being enabled in `app.toml` by default when the file is being created.

dadamu:
## Description

This PR is the implementation of #469.
Closes #469.

## Checklist
- [ ] Wrote unit tests.


RiccardoM:
@dadamu Is there a way that allow us to call the default `init` command provided by Cosmos and only after, edit the configuration file instead of copy/pasting that command run function? If so I think this is what we should so

dadamu:
@RiccardoM The command function returns `*cobra.Command` to perform the `RunE` function to generate all the initial configs. 
There may be a way to extend the `RunE`. Let me explore it a while.

dadamu:
@RiccardoM done

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/mr-tron/base58 v1.2.0
+	github.com/pkg/errors v0.9.1
```
RiccardoM, line 16: Is this really necessary?

app/desmos/cmd/init.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// initCmd returns a command that initializes all files needed for Tendermint
+// and the respective application.
+func initCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
+	cmd := cosmosgenutilcli.InitCmd(mbm, defaultNodeHome)
+	defaultRun := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		clientCtx := client.GetClientContextFromCmd(cmd)
+		err := defaultRun(cmd, args)
```
RiccardoM, line 23: We should check for errors here
```suggestion
		err := defaultRun(cmd, args)
		if err != nil {
			return err
		}
```

app/desmos/cmd/init.go
```diff
@@ -0,0 +1,32 @@
...
+
+		// Set app.toml file
+		appConfig := srvconfig.DefaultConfig()
+		appConfig.StateSync.SnapshotInterval = 500
+		srvconfig.WriteConfigFile(filepath.Join(clientCtx.HomeDir, "config", "app.toml"), appConfig)
+		return err
```
RiccardoM, line 29: If the error is checked above, we can return nil here: 
```suggestion
		return nil
```

app/desmos/cmd/init.go
```diff
@@ -0,0 +1,32 @@
...
+func initCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
+	cmd := cosmosgenutilcli.InitCmd(mbm, defaultNodeHome)
+	defaultRun := cmd.RunE
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		clientCtx := client.GetClientContextFromCmd(cmd)
+		err := defaultRun(cmd, args)
+
+		// Set app.toml file
+		appConfig := srvconfig.DefaultConfig()
+		appConfig.StateSync.SnapshotInterval = 500
```
RiccardoM, line 27: Probably we need to also edit the sync infos to make sure they are compatible with state sync. From the `app.toml` file we can read

```toml
# snapshot-interval specifies the block interval at which local state sync snapshots are
# taken (0 to disable). Must be a multiple of pruning-keep-every.
snapshot-interval = 0
```

So we should change these values as well: 

```toml
# default: the last 100 states are kept in addition to every 500th state; pruning at 10 block intervals
# nothing: all historic states will be saved, nothing will be deleted (i.e. archiving node)
# everything: all saved states will be deleted, storing only the current state; pruning at 10 block intervals
# custom: allow pruning options to be manually specified through 'pruning-keep-recent', 'pruning-keep-every', and 'pruning-interval'
pruning = "default"

# These are applied if and only if the pruning strategy is custom.
pruning-keep-recent = "0"
pruning-keep-every = "0"
pruning-interval = "0"
```

Maybe something like this works: 

```toml
pruning = "custom"

# These are applied if and only if the pruning strategy is custom.
pruning-keep-recent = "100"
pruning-keep-every = "500"
pruning-interval = "10"
```

This way we can make sure that even if the `default` values change, state sync will work properly

bragaz (commented):
The overall changes looks fine to me, do we keep the `SnapshotKeepRecent` to 2?
Or it should/could be custom depending on validators?

dadamu:
@bragaz I think we can keep it to 2 as default, the validator can set custom value if they want to change it after file is generated.

RiccardoM (commented):
I think we should update the node setup guide accordingly telling about this feature being enabled by default, and how to tweak it

docs/fullnode/setup.md
```diff
@@ -170,6 +170,31 @@ trust_hash = "E8ED7A890A64986246EEB02D7D8C4A6D497E3B60C0CAFDDE30F2EE385204C314"
 trust_period = "168h0m0s"
 ```
 
+## (Optional) Edit snapshot config
+
+Currently, the `snapshot` feature is open by the default. It allows other nodes to rapidly join the network with downloading the application state at given block height.
```
bragaz, line 175:
```suggestion
Currently, the `snapshot` feature is open by the default. It allows other nodes to quickly join the network by syncing the application state at given block height.
```

docs/fullnode/setup.md
```diff
@@ -170,6 +170,31 @@ trust_hash = "E8ED7A890A64986246EEB02D7D8C4A6D497E3B60C0CAFDDE30F2EE385204C314"
...
+
+**Note: Make sure that snapshot-interval must be a multiple of pruning-keep-every in the `base` section**
+
+```toml
+pruning-keep-recent = "100"
+pruning-keep-every = "500"
+pruning-interval = "10"
+```
+
+The more information about pruning is [here](./overview.md#understanding-pruning).
```
bragaz, line 196:
```suggestion 
You can find out more about pruning [here](./overview.md#understanding-pruning).
```

bragaz (commented):
just some small changes

bragaz (approved):

docs/fullnode/setup.md
```diff
@@ -170,6 +170,31 @@ trust_hash = "E8ED7A890A64986246EEB02D7D8C4A6D497E3B60C0CAFDDE30F2EE385204C314"
...
+
+**Note: Make sure that snapshot-interval must be a multiple of pruning-keep-every in the `base` section**
+
+```toml
+pruning-keep-recent = "100"
+pruning-keep-every = "500"
+pruning-interval = "10"
+```
+
+You can find out more about pruning [here](./overview.md#understanding-pruning).
```
RiccardoM, line 196:
```suggestion
Currently, the `snapshot` feature is enabled by the default. This means that your node will periodically create snapshots of the chain state and make them public, allowing other nodes to quickly join the network by syncing the application state at a given height.

By default, we have set Desmos to take snapshots every 500 blocks, and persist the last 2 snapshots, deleting older ones. If you want to provide other nodes with more (or less) frequent snapshots, you can do this by editing a couple of things inside your `~/.desmos/config/app.toml` file, under the `state-sync` section:

```toml
# snapshot-interval specifies the block interval at which local state sync snapshots are
# taken (0 to disable). Must be a multiple of pruning-keep-every.
snapshot-interval = 500

# snapshot-keep-recent specifies the number of recent snapshots to keep and serve (0 to keep all).
snapshot-keep-recent = 2
```

**Note: Make sure that snapshot-interval is a multiple of `pruning-keep-every` in the `base` section**

```toml
pruning-keep-recent = "100"
pruning-keep-every = "500"
pruning-interval = "10"
```

You can find out more about pruning [here](./overview.md#understanding-pruning).
```

PR #533, opened 2021-07-05
Case insensitive chain name

### LINKED ISSUE #526: Make the name of ChainConfig case insensitive
## Feature description 

Currently, the name of the chain config is not In `x/profile` is not specified to use lower case. 
This means that currently it's possible to have many chain configs with very similar chain name, say:

* Cosmos
* cosmos
* cosMos
* ...  

This might be a problem in the future due to the possible impersonification.

## Implementation proposal

To solve it, we can convert the name to lower case in the constructor and make sure it is all in lower case by Validate().
```golang
// NewChainConfig allows to build a new ChainConfig instance
func NewChainConfig(name string) ChainConfig {
	return ChainConfig{
		Name: strings.ToLower(name),
	}
}

// Validate checks the validity of the ChainConfig
func (c ChainConfig) Validate() error {
	if strings.TrimSpace(c.Name) == "" {
		return fmt.Errorf("chain name cannot be empty or blank")
	}
	if strings.ToLower(c.Name) != c.Name {
	        return fmt.Errorof("chain name should be lower case")
	}
	return nil
}
```

What do you think? @RiccardoM @bragaz

dadamu:
## Description

This PR makes the name of `ChainConfig` case insensitive.
Closes #526

## Checklist
- [ ] Wrote unit tests.


bragaz (approved):

x/profiles/types/models_chain_links.go
```diff
@@ -22,7 +22,7 @@ import (
 // NewChainConfig allows to build a new ChainConfig instance
 func NewChainConfig(name string) ChainConfig {
 	return ChainConfig{
-		Name: name,
+		Name: strings.ToLower(name),
```
RiccardoM, line 25: I think this should be removed if we want the `Validate` method to fail

RiccardoM (commented):
The implementation looks good, but I'm having a hard time deciding whether this is worth or it might have some backslash in the future

bragaz:
@RiccardoM why backslash? I don't think we will see chains where name differs only for some case letter.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# Unreleased
+- Made the name of ChainConfig case-insensitive ([#533](https://github.com/desmos-labs/desmos/pull/533))
```
RiccardoM, line 2:
```suggestion
- Require chain name to be lowercase ([#533](https://github.com/desmos-labs/desmos/pull/533))
```

RiccardoM (approved):

PR #556, opened 2021-07-12
Endblocker subspace checks

### LINKED ISSUE #488: `EndBlocker` `subspaces` checks
With the addition of the `subspaces` module, `profiles` (and maybe `posts`) could need some specific checks that runs automatically in their `EndBlocker`s to ensure if users involved in operations like `relationships` or `blocks` 
are still registered in those `subspaces` where the operations are performed and, in case they're not, manage it.
The implementation shouldn't be difficult and I will take care of post here a proposal in the following days.
Thoughts on this? @RiccardoM @dadamu

dadamu:
## Description

This PR is the implementation of #488. It removes all the relationships and blocks of the subspaces-unregistered users during end block.
Closes #488.

## Checklist
- [x] Wrote unit tests.


x/profiles/keeper/keeper_blocks.go
```diff
@@ -108,3 +108,17 @@ func (k Keeper) DeleteAllUserBlocks(ctx sdk.Context, user string) {
 		store.Delete(types.UserBlockStoreKey(block.Blocker, block.Subspace, block.Blocked))
 	}
 }
+
+// DeleteSubspaceUserBlocks deletes the user blocks created by the specified user from the subspace having the given id
+func (k Keeper) DeleteSubspaceUserBlocks(ctx sdk.Context, subspaceID, user string) {
```
bragaz, line 113: I think that this method can be simply written like this:
```go
// DeleteSubspaceUserBlocks deletes the user blocks created by the specified user from the subspace having the given id
func (k Keeper) DeleteSubspaceUserBlocks(ctx sdk.Context, subspaceID, user string) {

	k.IterateSubspaceUserBlocks(ctx, user, subspaceID, func(index int64, block types.UserBlock) (stop bool) {
		store.Delete(types.UserBlockStoreKey(block.Blocker, block.Subspace, block.Blocked))
		return false
	})
}
```
Also we should add some unit tests for it

x/profiles/keeper/keeper_relationships.go
```diff
@@ -89,3 +89,17 @@ func (k Keeper) DeleteAllUserRelationships(ctx sdk.Context, user string) {
 		store.Delete(types.RelationshipsStoreKey(relationship.Creator, relationship.Subspace, relationship.Recipient))
 	}
 }
+
+// DeleteSubspaceUserRelationships deletes the relationships from the subspace having the given id
+func (k Keeper) DeleteSubspaceUserRelationships(ctx sdk.Context, subspaceID, user string) {
```
bragaz, line 94: Same as above for `blocks`

x/profiles/types/expected_keeper.go
```diff
@@ -6,6 +6,8 @@ import (
 	connectiontypes "github.com/cosmos/cosmos-sdk/x/ibc/core/03-connection/types"
 	channeltypes "github.com/cosmos/cosmos-sdk/x/ibc/core/04-channel/types"
 	ibcexported "github.com/cosmos/cosmos-sdk/x/ibc/core/exported"
+
+	subspacestypes "github.com/desmos-labs/desmos/x/staging/subspaces/types"
 )
```
bragaz, line 12: better move this file inside `keeper` folder as it's inside `x/posts`

bragaz (commented):
The overall looks good, some small stuff to adjust

x/profiles/keeper/keeper.go
```diff
@@ -205,3 +210,14 @@ func (k Keeper) ValidateProfile(ctx sdk.Context, profile *types.Profile) error {
 
 	return profile.Validate()
 }
+
+// DeleteUnregisteredRelationshipsAndBlocks deletes the relationships and blocks of subspaces-unregistered users
+func (k Keeper) DeleteUnregisteredRelationshipsAndBlocks(ctx sdk.Context) {
```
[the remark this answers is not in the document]
dadamu: @bragaz This function actually does 3 deletion things, deleting unregistered user relationships, unregistered user blocks and unregistered pairs. I think it can be simplified to `DeleteUnregisteredUserService` and add more comments to explain what the function does. What do you think?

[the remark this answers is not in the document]
bragaz: In this case I think `DeleteUnregisteredUserReferences` can better summarise what the methods do, please add also a further comment telling it's deleting also the pair

bragaz (approved):
Everything's cool for me! 👍

RiccardoM:
Closing this as per #690

PR #557, opened 2021-07-13
Remove profile duplicated validation

### LINKED ISSUE #548: Double validation when saving profile
Inside the `SaveProfile` function of `msg_server_profile.go` we validate the profile twice:
1. Inside the `Update` function;
2. Another time inside keeper's `ValidateProfile()`;

I think the validation could be removed from within the `Update()` function and left only inside the keeper's `Validate()`
function since `Update()` is only used once.

bragaz:
## Description
This PR remove the double validation made inside `Profile` `Update()` function leaving only that
inside the `keeper` `Validate()`.
Closes #548

## Checklist
- [ ] Wrote unit tests.


dadamu (approved):
LGTM

RiccardoM:
Re-checking this, to be honest I'm not sure whether we should have this removed inside the `Update` function or not. Ideally, that function can be used by third party developers as well, and those checks are very quick to perform. For this reason, I don't think there's the necessity to remove it at all. We could simply leave it here. Before jumping to the conclusions, though, is it possible to have some data? Like, could we measure the impact on performances that this double check has? I think this should be farily easy to do and it could give us a nice idea whether it is worth or not removing the check.

bragaz:
@RiccardoM I've performed a bench test on `Update()` and these are the results:

with validation:
```
goos: darwin
goarch: amd64
pkg: github.com/desmos-labs/desmos/x/profiles/types
BenchmarkUpdate

BenchmarkUpdate-1   	  296853	      3937 ns/op
BenchmarkUpdate-2   	  310422	              3490 ns/op
BenchmarkUpdate-3   	  308560	      3546 ns/op
BenchmarkUpdate-4   	  311792	              3612 ns/op
BenchmarkUpdate-5   	  312789	              3666 ns/op
BenchmarkUpdate-6   	  286815	              3789 ns/op
BenchmarkUpdate-7   	  293988	      4159 ns/op
BenchmarkUpdate-8   	  311395	              3671 ns/op
BenchmarkUpdate-9   	  313987	              3518 ns/op
BenchmarkUpdate-10     320106	              3681 ns/op
```

without validation:
```
goos: darwin
goarch: amd64
pkg: github.com/desmos-labs/desmos/x/profiles/types
BenchmarkUpdate

BenchmarkUpdate-1   	 3203461	       380 ns/op
BenchmarkUpdate-2   	 3068451	       378 ns/op
BenchmarkUpdate-3   	 3199720	       367 ns/op
BenchmarkUpdate-4   	 3229094	       381 ns/op
BenchmarkUpdate-5   	 3226158	       372 ns/op
BenchmarkUpdate-6   	 3105884	       386 ns/op
BenchmarkUpdate-7   	 3091993	       370 ns/op
BenchmarkUpdate-8   	 3018584	       386 ns/op
BenchmarkUpdate-9   	 3113236	       374 ns/op
BenchmarkUpdate-10    3172558	       376 ns/op
```
Thoughts?

RiccardoM:
@bragaz Can you please push the bechmark test code? It's hard to understand what those results mean without having a look at the code

bragaz:
@RiccardoM yeah sure, I forgot to push it

x/profiles/types/account_test.go
```diff
@@ -373,3 +352,31 @@ func TestProfileSerialization(t *testing.T) {
...
+	update := types.NewProfileUpdate(
+		"dtag-2",
+		"nickname-2",
+		"bio-2",
+		types.NewPictures(
+			"https://example.com/2",
+			"https://example.com/2",
+		),
+	)
+
```
RiccardoM, line 378: Let's reset the timer here

```suggestion

b.ResetTimer()
```

RiccardoM: @bragaz Can you run the tests again after this? Only one time per each test (with/without validation)

RiccardoM:
@bragaz After watching at the benchmarks and the code, I think it's safe to delete he `Validate` from within the `Update` method. Could you please update the `Update` documentation telling that it does not validate the updated profile, and this should be done before storing the profile itself? Just as a precaution for external developers

bragaz:
@RiccardoM done

x/profiles/types/account.go
```diff
@@ -261,8 +261,9 @@ func NewProfileUpdate(dTag, nickname, bio string, pictures Pictures) *ProfileUpd
 	}
 }
 
-// Update updates the fields of a given profile. An error is
-// returned if the resulting profile contains invalid values.
+// Update updates the fields of a given profile without validating it.
+// Before store the updated profile, a validation with Validate() should
```
RiccardoM, line 265:
```suggestion
// Before storing the updated profile, a validation with Validate() should
```

RiccardoM (approved):

PR #562, opened 2021-07-14
Implement expiration for application links

### LINKED ISSUE #516: Implement expiration data for application links
## Feature description 
Currently when a user links their centralized application with the Desmos profile, the created link contains a timestamp of when it has been created. Since centralized applications username can be switched and sold, we should implement a sort of "expiration date" on links. This means that after a certain amount of time passes, the link will be automatically marked as invalid, and the user has to connect it again in order to keep it valid. 

## Implementation proposal
The implementation should be very simple. What we can do is: 
1. Add a parameter inside the `x/profiles` parameters set, identifying how long a link should be considered valid from when it has been created. 
2. Inside the `EndBlocker`, check for links that are expired and update their state to set them as such. 

Please @bragaz @dadamu tell me if you see any problem with this implementation or better alternatives.

dadamu:
## Description

This PR is the implementation of #516. It allows to update the expiration state for application links in blocks.
Close #516

## Checklist
- [x] Wrote unit tests.


x/profiles/types/keys.go
```diff
@@ -133,3 +137,18 @@ func UserApplicationLinkKey(user, application, username string) []byte {
 func ApplicationLinkClientIDKey(clientID string) []byte {
 	return append(ApplicationLinkClientIDPrefix, []byte(clientID)...)
 }
+
+// ExpiringApplicationLinkPrefix returns the prefix used to store client id keys of the expiring application links
+// at the given block height
+func ExpiringApplicationLinkPrefix(blockHeight int64) []byte {
```
bragaz, line 143: I think this can be simply written like this:
```golang
buf := make([]byte, binary.MaxVarintLen64)
binary.BigEndian.PutUint64(buf, uint64(blockHeight))
return append(ApplicationLinkExpirationPrefix, buf...)
```
as stated [here](https://stackoverflow.com/a/35371760) it will not affect the sign of the number

x/profiles/types/models_params.go
```diff
@@ -260,3 +274,33 @@ func ValidateOracleParams(i interface{}) error {
...
+}
+
+// ValidateApplicationLinkParams returns an error if interface does not represent a valid ApplicationLinkParams instance
+func ValidateApplicationLinkParams(i interface{}) error {
+	params, isApplicationLinkParams := i.(ApplicationLinkParams)
+	if !isApplicationLinkParams {
+		return fmt.Errorf("invalid parameters type: %s", i)
+	}
+
+	if params.ExpiryInterval <= 0 {
```
bragaz, line 301: since we're checking this here, wouldn't be better to use directly `uint64`?

proto/desmos/profiles/v1beta1/models_params.proto
```diff
@@ -119,3 +124,14 @@ message OracleParams {
...
   ];
 }
+
+// ApplicationLinkParams defines the parameters related to application link
+message ApplicationLinkParams {
+  option (gogoproto.goproto_getters) = false;
+
+  // ExpiryInterval represents the expiration interval in blocks for an
+  // application link at creation
+  int64 expiry_interval = 1
```
bragaz, line 134: reviewing the whole PR made me think that maybe this should be `uint64` instead

bragaz (approved):
👍

.changeset/entries/2021-07-16T07:52:37Z.yaml
```diff
@@ -0,0 +1,5 @@
+type: added
+module: x/profiles
+pull_request: 562
+description: Added expiration for application links
+backward_compatible: true
```
RiccardoM, line 5: This is definitely not backward compatible as it adds a new feature that requires parameters changes etc

RiccardoM:
Superseeded by #723

PR #583, opened 2021-08-06
Add command generating chain link file

### LINKED ISSUE #572: Add command to generate chain link file
## Feature description 

Currently, the `sign` cmd returns the hex-ecoded public key. However, The public key in the proof object is required as base64-encoded string when creating a chain link. From my side, the `sign` cmd should return the public key as base64-encoded.

The chain link required file:
```json
{
   "address":{
      "@type":"/desmos.profiles.v1beta1.Bech32Address",
      "value":"cosmos15uc89vnzufu5kuhhsxdkltt38zfx8vcyggzwfm",
      "prefix":"cosmos"
   },
   "proof":{
      "pub_key":{
         "@type":"/cosmos.crypto.secp256k1.PubKey",
         "key":"A58DXR/lXKVkIjLofXgST/OHi+pkOQbVIiOjnTy7Zoqo"   /*<- It's base64-encoded*/
      },
      "signature":"ecc6175e730917fb289d3a9f4e49a5630a44b42d972f481342f540e09def2ec5169780d85c4e060d52cc3ffb3d677745a4d56cd385760735bc6db0f1816713be",
      "plain_text":"cosmos15uc89vnzufu5kuhhsxdkltt38zfx8vcyggzwfm"
   },
   "chain_config":{
      "name":"cosmos"
   }
}
```

## Implementation proposal

Let the public key of signature data in `sign` cmd be base64-encoded.

dadamu:
## Description

It is the implementation of #572, it add the command generating chain link file for creating chain link.
Closes #572.

## Checklist
- [x] Wrote unit tests.


app/desmos/cmd/generate-chain-link-json.go
```diff
@@ -0,0 +1,79 @@
...
+	addr, _ := sdk.Bech32ifyAddressBytes(app.Bech32MainPrefix, clientCtx.GetFromAddress())
+	sig, pubkey, err := clientCtx.Keyring.Sign(clientCtx.GetFromName(), []byte(addr))
+	if err != nil {
+		return profilescliutils.ChainLinkJSON{}, err
+	}
+
+	// create chain link json
+	cdc, _ := app.MakeCodecs()
+	chainLinkJSON := profilescliutils.NewChainLinkJSON(
+		types.NewBech32Address(addr, app.Bech32MainPrefix),
```
bragaz, line 71: Looks like `prefix` is never used, do you meant to use it here maybe?

app/desmos/cmd/generate-chain-link-json.go
```diff
@@ -0,0 +1,79 @@
...
+		types.NewProof(pubkey, hex.EncodeToString(sig), addr),
+		types.NewChainConfig(app.Bech32MainPrefix),
```
bragaz, line 73: as above

bragaz (commented):
As we discussed it should be useful to have the possibility to create json also for other chains.

RiccardoM:
To be honest I'm not a very big fun of this approach cause it requires the account to be already present inside the keystore in order to be used. What I think it's better is to have a UX that acts like the following: 

1. The user runs a command like `desmos create-chain-link` 
2. An interactive prompt is displayed, where the user is required to do the following: 
   1. Enter the mnemonic of the account they want to link
   2. Select the chain they want to link from a list of the supported ones (we can initially include Cosmos, Osmosis, Akash, etc); **or** select the _Other option_.  
      If _Other_ is selected, the user is required to input:  
      1. The Bech32 prefix
      2. The derivation path of the account

  3. Once everything is selected, we use this data to generate the private key, public key, address and from these to sign the data to create the link. 

Here is the complete flow:
Desmos Chain link

For the interactive prompt you can use something like [promptui](https://github.com/manifoldco/promptui)

dadamu:
Updated to apply interactive prompt to create chain link file.

app/desmos/cmd/create-chain-link.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// generateChainLink returns ChainLinkJSON for creating chain link
+func generateChainLink(mnemonic string, chain types.ChainType) (profilescliutils.ChainLinkJSON, error) {
+	// generate keybase for signing
+	keyBase := keyring.NewInMemory()
+	algo := hd.Secp256k1
+	hdPath := chain.DerivationPath
+	keyName := "chainlink"
+	_, err := keyBase.NewAccount("chainlink", mnemonic, "", hdPath, algo)
```
RiccardoM, line 70: I think this can be simplified 

```suggestion
	keyBase := keyring.NewInMemory()
	keyName := "chainlink"
	_, err := keyBase.NewAccount("chainlink", mnemonic, "", chain.DerivationPath, hd.Secp256k1)
```

app/desmos/cmd/create-chain-link.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// create chain link json
+	cdc, _ := app.MakeCodecs()
+	chainLinkJSON := profilescliutils.NewChainLinkJSON(
+		profilestypes.NewBech32Address(addr, chain.Prefix),
+		profilestypes.NewProof(pubkey, hex.EncodeToString(sig), addr),
+		profilestypes.NewChainConfig(chain.Name),
+	)
+	if err := chainLinkJSON.UnpackInterfaces(cdc); err != nil {
```
RiccardoM, line 90: What is this used for?

RiccardoM (commented):
I think that it would be a better UX to not use the `--filename` flag and, instead, print the result to the command output. This would allow the user to simply print it to any file they want using the `>` redirection.

app/desmos/cmd/create-chain-link.go
dadamu: It was used for json.Marshal. But, It has been replaced with cdc.Marshal already, I will remove it.

dadamu:
@RiccardoM The '>' interrupt the interactive prompt. With it, the user can't input the info.

RiccardoM:
> @RiccardoM The '>' interrupt the interactive prompt. With it, the user can't input the info. I guess we can add filename into prompt as well.

Yeah I think it's better to ask the user where to store the file using the prompt in this case, specifying it must be a fully qualified path. Just to avoid having them remember to also use a flag, it's better to keep the UX consistent for all the things

RiccardoM (approved):

PR #628, opened 2021-09-22
docs(adr): ADR-005: Enabled snapshot by default

dadamu:
## Description

This PR introduces the draft of the 5th ADR related to Enabled snapshot by default.
Discussed before here: #469.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/docs/architecture/adr-005-enabled-snap-shot-by-default.md
+++ b/docs/architecture/adr-005-enabled-snap-shot-by-default.md
@@ -1,0 +1,60 @@
+# ADR 005: Enabled snapshot by default
+
+## Changelog
+
+- {date}: {changelog}
+
+## Status
+
+{DRAFT | PROPOSED} Not Implemented
+
+> Please have a look at the [PROCESS](./PROCESS.md#adr-status) page.
+> Use DRAFT if the ADR is in a draft stage (draft PR) or PROPOSED if it's in review.
+
+## Abstract
+
+> "If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the ADR.
+> A short (~200 word) description of the issue being addressed.
+
+## Context
+
+> This section describes the forces at play, including technological, political, social, and project local. These forces are probably in tension, and should be called out as such. The language in this section is value-neutral. It is simply describing facts. It should clearly explain the problem and motivation that the proposal aims to resolve.
+> {context body}
+
+## Decision
+
+> This section describes our response to these forces. It is stated in full sentences, with active voice. "We will ..."
+> {decision body}
+
+## Consequences
+
+> This section describes the resulting context, after applying the decision. All consequences should be listed here, not just the "positive" ones. A particular decision may have positive, negative, and neutral consequences, but all of them affect the team and project in the future.
+
+### Backwards Compatibility
+
+> All ADRs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The ADR must explain how the author proposes to deal with these incompatibilities. ADR submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+### Positive
+
+{positive consequences}
+
+### Negative
+
+{negative consequences}
+
+### Neutral
+
+{neutral consequences}
+
+## Further Discussions
+
+While an ADR is in the DRAFT or PROPOSED stage, this section should contain a summary of issues to be solved in future iterations (usually referencing comments from a pull-request discussion).
+Later, this section can optionally list ideas or improvements the author or reviewers found during the analysis of this ADR.
+
+## Test Cases [optional]
+
+Test cases for an implementation are mandatory for ADRs that are affecting consensus changes. Other ADRs can choose to include links to test cases if applicable.
+
+## References
+
+- {reference link}


docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,98 @@
...
+## Status
+
+PROPOSED
+
+## Abstract
+
+Currently, most of the validators or full nodes does not have snapshot enabled because they use the default 
+setting whose snapshot is not enabled. This makes state sync a new node very difficult. 
+In oder to increase the number of snapshots in the network so that making creating a new node easily through 
+`StateSync`, we SHOULD enable snapshot by default.
```
RiccardoM, line 17:
```suggestion
Currently, most of the validators or full nodes do not have snapshot enabled because they use the default 
setting which have the snapshot feature not enabled. This makes it very difficult to use state sync on a new node. 
In order to increase the number of snapshots in the network, and make it easier to create a new node easily through `StateSync`, we SHOULD enable snapshot by default.  
```

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Context
+
+Tendermint provides an alternative mechanism called `StateSync` for bootstrapping a new node without downloading 
+blocks or going through consensus. It fetches a snapshot of the state machine at a given height and restores it.
+`StateSync` will greatly improve the experience of joining a network, reducing the time required to sync a node
+by several orders of magnitude.
+However, state sync a node is almost impossible now since there is too few snapshot in the network now. In order to ensure 
+the number of snapshots is enough, we SHOULD encourage validators and full nodes to set snapshot enabled. 
+As the result, making it enabled by default can be more convenient to who wants to provide the snapshot to the network.
```
RiccardoM, line 27:
```suggestion
However, state syncing a node is almost impossible now since there are too few snapshot in the network. In order to ensure the number of snapshots is enough, we SHOULD encourage validators and full nodes to enable the snapshot creation process. 
As a result, enabling it by default can be more convenient to who wants to provide the snapshot to the network. 
```

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,98 @@
...
+)
+```
+
+
+## Consequences
+
+### Backwards Compatibility
+
+The change only includes replacing the `initCmd` to custom one to enable snapshot so the backwards compatibility is 
+not relevant as there won't be any issue related to it.
```
RiccardoM, line 77: Actually, there are some consequences here. 

We use the `init` command also when creating a testing network. This change would make it impossible to run the simulation tests and the on-chain upgrade tests properly since the node would be created with the state sync enabled.

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Positive
+
+* Increase the number of snapshots in the network
+* Make creating a new node easily by StateSync
+
+### Negative
+
+(none known)
```
RiccardoM, line 86: A negative side of this change is that enabling snapshot creation by default would require validators a higher amount of disk space which might cause them to spend more on their setup. This needs to be taken into consideration.

RiccardoM, line 86: Another negative side is that in order to get the snapshots from a node, the node MUST have the proper TCP/IP port opened accordingly, otherwise any snapshost download will fail.

docs/architecture/adr-005-enabled-snap-shot-by-default.md
dadamu: The state sync feature by default is closed. This ADR only enable snapshot, so sim-tests and on-chain upgrade tests would work properly.

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,97 @@
...
+blocks or going through consensus. It fetches a snapshot of the state machine at a given height and restores it.
+`StateSync` will greatly improve the experience of joining a network, reducing the time required to sync a node
+by several orders of magnitude.
+However, state syncing a node is almost impossible now since there are too few snapshot in the network. In order to ensure the number of snapshots is enough, we SHOULD encourage validators and full nodes to enable the snapshot creation process. 
+As a result, enabling it by default can be more convenient to who wants to provide the snapshot to the network. 
+In addition, it would increase the willing of validators or full nodes to enable snapshot as well.
+
+## Decision
+
+The implementation idea is to build a custom `initCmd` including `StatSync` setting to init the `config/app.toml`
```
bragaz, line 30:
```suggestion
The implementation idea is to build a custom `initCmd` including `StateSync` setting to init the `config/app.toml` 
```

bragaz (commented):
To mitigate the downsides of enabling the snapshot by default and give validators a chance to think about whether they want to enable it or not, maybe we can add a flag to the init command for this purpose. What do you think?
@RiccardoM @dadamu

dadamu:
@bragaz I agree with it. It makes enabling snapshot feature or not by default easier to validators or full nodes.

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Backwards Compatibility
+
+The change only includes replacing the `initCmd` to custom one to enable snapshot so the backwards compatibility is 
+not relevant as there won't be any issue related to it.
+
+### Positive
+
+* Increase the number of snapshots in the network
+* Make creating a new node easily by StateSync
```
RiccardoM, line 80:
```suggestion
* Simplify the initial sync of any full node
```

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Negative
+
+* Need higher amount of disk space
```
RiccardoM, line 84:
```suggestion
* Full nodes will need a higher amount of disk space
```

docs/architecture/adr-005-enabled-snap-shot-by-default.md
```diff
@@ -0,0 +1,97 @@
...
+* Require that the node MUST have the proper TCP/IP port opened, or any snapshot download will fail
```
RiccardoM, line 85:
```suggestion
* To work properly, all full nodes providing snapshots MUST have the proper TCP/IP port opened, or any snapshot download will fail
```

RiccardoM (commented):
I agree with @bragaz too. It would be better to not make this an option enabled by default, but instead have a set of flags that allow to enable it. We should first search if this is already possible with the default `init` command though.

Issue #685, opened 2021-12-06
Add support for hex address in chain link


dadamu:
## Feature description 

Currently, Desmos doesn't support linking the hex address to the profile. Therefore, the user can't connect their address of Etheruem or other chains, which apply hex address, to the profile.
We should provide the possibility to create the chain link with a hex address.

## Implementation proposal

Implement the Hex address structure for `AddressData` interface.

bragaz:
I agree with this. We should extend the support to ETH as well. Maybe this is worth for an ADR? @RiccardoM

RiccardoM:
Agree this is something we should be supporting

> I agree with this. We should extend the support to ETH as well. Maybe this is worth for an ADR? @RiccardoM

No ADR is required for such a small change. We can just implement it.

dadamu:
Cool, will work on it soon.

PR #688, opened 2021-12-07
feat: add support for hex address in chain link

dadamu:
## Description

Closes: #685 

This PR is the implementation of #685. It extends the support to the hex address like ETH.

Note: Rename PR to add support for eth address since eth address is generated in the other way.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/profiles/types/models_chain_links.go
```diff
@@ -202,6 +204,51 @@ func (b Base58Address) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
...
+	bz, err := hex.DecodeString(addr)
+	if err != nil {
+		return false, err
+	}
+	pub, err := btcec.ParsePubKey(key.Bytes(), btcec.S256())
+	if err != nil {
+		return false, err
+	}
+	uncompressedPub := pub.SerializeUncompressed()
+	return bytes.Equal(crypto.Keccak256(uncompressedPub[1:])[12:], bz), err
```
dadamu, line 247: Ethereum address is generated by uncompressed public key which is 65 bytes.

Reference: https://goethereumbook.org/wallet-generate/

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,18 @@ message Base58Address {
 
   // Value contains the Base58-encoded address
   string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+}
+
+// EthAddress represents a Eth address
```
bragaz, line 98:
```suggestion
// EthAddress represents an Eth address
```

x/profiles/types/models_chain_links.go
```diff
@@ -202,6 +204,51 @@ func (b Base58Address) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
...
 
+var _ AddressData = &EthAddress{}
+
+// NewEthAddress returns a new EthAddress instance
+func NewEthAddress(value, prefix string) *EthAddress {
+	return &EthAddress{Value: value, Prefix: prefix}
+}
+
+func (e EthAddress) Validate() error {
+	if len(strings.TrimSpace(e.Value)) <= len(e.Prefix) {
```
bragaz, line 215:
```suggestion
	if len(strings.TrimSpace(e.Value)) <= len(strings.TrimSpace(e.Prefix)) {
```
maybe better to trim the spaces from prefix too

x/profiles/types/models_chain_links.go
bragaz, line 247:
```suggestion
	return bytes.Equal(crypto.Keccak256(uncompressedPubKey[1:])[12:], bz), err
```

x/profiles/types/models_chain_links.go
dadamu: I think `strings.TrimSpace` should be remove here since it is for avoiding panic in the next codeline 
`prefix, addrWithoutPrefix := e.Value[:len(e.Prefix)], e.Value[len(e.Prefix):]`.

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,18 @@ message Base58Address {
 
   // Value contains the Base58-encoded address
   string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+}
+
+// EthAddress represents an Eth address
+message EthAddress {
```
RiccardoM, line 99: I think we should name this something like `HexAddress` since the other ones have names that are related to the way they are encoded (`Bech32Address`, `Base58Address`) and not the chains they are used on (`CosmosAddress`, `SolanaAddress`).

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,18 @@ message Base58Address {
...
+  option (gogoproto.goproto_getters) = false;
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+  option (cosmos_proto.implements_interface) = "AddressData";
+
+  // Value represents the Eth address value
+  string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+
+  // Prefix represents the prefix of the Eth address
+  string prefix = 2 [ (gogoproto.moretags) = "yaml:\"prefix\"" ];
```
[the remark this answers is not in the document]
dadamu: Eth only use `0x` as prefix, will remove it.

proto/desmos/profiles/v1beta1/models_chain_links.proto
dadamu: Maybe we should call it `EthHexAddress`.

bragaz: I would go for a generically `HexAddress` too in order to be consistent with the names we used for the other chains.

dadamu: I noticed that the eth address applies `keccak256` to hash uncompressed public key into address while solana and cosmos utilizes `ripemd160` hasher and compressed public key. 
Any other chains might use different algorithm to generate `base58/bech32` address as well, like `Tron` uses `keccak256` with uncompressed public key then encode the address byte into base58, the current `Base58Address` will not be valid to `Tron`.

RiccardoM: @dadamu I think it's better to open an issue on this specific finding so that we can see how we should edit the `AddressData` interface to support different encoding as well

dadamu: @RiccardoM 
Okay, I will rename it into hex address then note that current `HexAddress` mainly supports eth address.

RiccardoM: @dadamu No need to note that it supports ETH addresses specifically. We can just specify that it supports keccak256-compressed addresses

dadamu: @RiccardoM @bragaz Updated.

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,15 @@ message Base58Address {
 
   // Value contains the Base58-encoded address
   string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+}
+
+// HexAddress represents an Eth Hex-encoded address
```
bragaz, line 98:
```suggestion
// HexAddress represents an Hex-encoded address
```

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,15 @@ message Base58Address {
...
+message HexAddress {
+  option (gogoproto.goproto_getters) = false;
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+  option (cosmos_proto.implements_interface) = "AddressData";
+
+  // Value represents the Eth hex address value
```
bragaz, line 105:
```suggestion
  // Value represents the hex address value
```

x/profiles/types/models_chain_links.go
```diff
@@ -202,6 +204,57 @@ func (b Base58Address) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
...
+		return fmt.Errorf("value cannot be empty or blank")
+	}
+
+	if len(h.Value) <= 2 {
+		return fmt.Errorf("address cannot be smaller than prefix")
+	}
+
+	prefix, addrWithoutPrefix := h.Value[:2], h.Value[2:]
+	if prefix != "0x" {
+		return fmt.Errorf("prefix does not match")
```
bragaz, line 227: I think we should be more clear with this error, extending this with "address prefix does not match the 0x prefix"

bragaz (approved):
Looks ready now 👍

x/profiles/types/models_chain_links.go
```diff
@@ -202,6 +204,57 @@ func (b Base58Address) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
 
 // --------------------------------------------------------------------------------------------------------------------
 
+var _ AddressData = &HexAddress{}
+
+// NewHexAddress returns a new HexAddress instance
+// NOTE: Currently it only supports keccak256-uncompressed addresses
```
RiccardoM, line 210: I would move this note directly inside the Protobuf `message` description, so that it's visible to anyone using this type
```suggestion
```

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,15 @@ message Base58Address {
 
   // Value contains the Base58-encoded address
   string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+}
+
+// HexAddress represents an Hex-encoded address
```
RiccardoM, line 98: Here
```suggestion
// HexAddress represents an Hex-encoded address
// NOTE: Currently it only supports keccak256-uncompressed addresses
```

x/profiles/types/models_chain_links.go
```diff
@@ -202,6 +204,57 @@ func (b Base58Address) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
...
+	}
+
+	if len(h.Value) <= 2 {
+		return fmt.Errorf("address cannot be smaller than prefix")
+	}
+
+	prefix, addrWithoutPrefix := h.Value[:2], h.Value[2:]
+	if prefix != "0x" {
+		return fmt.Errorf("address prefix does not match the 0x prefix")
+	}
```
RiccardoM, line 228: Honestly I'm not very convinced of these checks. There might be some chains that do not have any prefix before the real address. What we should do instead is handling the `0x` prefix as something that is completely optional IMO

dadamu: Since there might be different prefix from `0x`, should we add prefix into `HexAddress`? Then, we can skip prefix checking like:
```
_, addrWithoutPrefix := h.Value[:len(h.Prefix)], h.Value[len(h.Prefix):]
```

RiccardoM: @dadamu Before deciding to implement this back, do you have a handful of examples of chains that use hex-encoded addresses, and the prefixes they use? It would be better to analyze that before deciding how to do things

dadamu: @RiccardoM Nope, currently only eth applies hex address. Other chains almost use `bech32` or `base58`.

RiccardoM: @dadamu Ok, thanks. Honestly, since the hex encoding can only have one prefix (which is `0x`), I think we can assume that the `Value` of a `HexAddress` can have or not such prefix. For this reason, I think we can implement the `Prefix` field as you did previously, allowing it to also be empty in the case that a chain does not use it. The only check we will have to do then is to make sure that the `Value` begins with the prefix.

proto/desmos/profiles/v1beta1/models_chain_links.proto
```diff
@@ -93,4 +93,19 @@ message Base58Address {
...
+message HexAddress {
+  option (gogoproto.goproto_getters) = false;
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+  option (cosmos_proto.implements_interface) = "AddressData";
+
+  // Value represents the hex address value
+  string value = 1 [ (gogoproto.moretags) = "yaml:\"value\"" ];
+
+  // Prefix represents the prefix of the Eth address
```
RiccardoM, line 109:
```suggestion
  // Prefix represents the optional prefix used during address encoding (e.g. 0x)
```

x/profiles/types/models_chain_links_test.go
```diff
@@ -306,6 +329,65 @@ func TestBase58Address_GetValue(t *testing.T) {
...
+		address   *types.HexAddress
+		shouldErr bool
+	}{
+		{
+			name:      "empty address returns error",
+			address:   types.NewHexAddress("", ""),
+			shouldErr: true,
+		},
+		{
+			name:      "address smaller than prefix length returns error",
```
RiccardoM, line 344:
```suggestion
			name:      "address value shorter than prefix returns error",
```

x/profiles/types/models_chain_links_test.go
```diff
@@ -306,6 +329,65 @@ func TestBase58Address_GetValue(t *testing.T) {
...
+			address:   types.NewHexAddress("0", "0x"),
+			shouldErr: true,
+		},
+		{
+			name:      "prefix does not match returns error",
```
RiccardoM, line 349:
```suggestion
			name:      "not matching prefix returns error",
```

x/profiles/types/models_chain_links_test.go
```diff
@@ -306,6 +329,65 @@ func TestBase58Address_GetValue(t *testing.T) {
...
 
+func TestHexAddress_Validate(t *testing.T) {
+	testCases := []struct {
+		name      string
+		address   *types.HexAddress
+		shouldErr bool
+	}{
+		{
+			name:      "empty address returns error",
+			address:   types.NewHexAddress("", ""),
```
RiccardoM, line 340: We can cover two cases in one: empty and blank
```suggestion
			address:   types.NewHexAddress("", "   "),
```

RiccardoM (approved):
Just some small changes but otherwise LGTM

PR #694, opened 2021-12-15
docs(adr): ADR-006: Subspaces module

RiccardoM:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


docs/architecture/adr-006-subspace-module.md
```diff
@@ -0,0 +1,157 @@
...
+canWrite := (userPermissions & PermissionWrite) == PermissionWrite  // True
+canModerateContent := (userPermissions & PermissionModerateContent) == PermissionModerateContent // False
+```
+
+### `Msg` Service
+We will allow the following operations to be performed:
+
+* delete contents that do not respect the ToS;
+* ban users that do not respect the ToS.
+
```
bragaz, line 99:
```suggestion
* create a subspace
* edit a subspace
* set user permissions
```

bragaz: And I think we should allow the possibility to transfer the subspace ownership too

bragaz (commented):
Well, we discussed this months ago and at this stage this is what we envisioned for the "final" version of subspaces (with ACL helping building an easy way to manage access). I've left a some comments up here but the overall ADR to me is good. I've only one concern about the treasury account. Since I've proposed the ownership transfer, how should we manage the treasury transfer? Make the treasury transfer all in one with the owernship transfer?

docs/architecture/adr-006-subspace-module.md
RiccardoM: >  I think we should allow the possibility to transfer the subspace ownership too
@bragaz This should be already handled by the `MsgEditSubspace` that allows to change the treasury address

RiccardoM:
> Well, we discussed this months ago and at this stage this is what we envisioned for the "final" version of subspaces (with ACL helping building an easy way to manage access). 

@bragaz Yes, confronting with the `staging` implementation, the only change I made were the following: 
1. Changed the subspace id from `string` to `int`.
  Honestly I think it makes little sense to force users to select their subspace id using complicated SHA-256 hashed that also take up a lot of disk space. Instead, I think we should adopt an approach more similar to the on-chain proposals that the Cosmos SDK uses: generate the ID in an incrementing way. So users will not be able to specify the subspace ID when they create it, but it will be returned to them upon creation. This should make it easier to later reference the subspace (an integer is easier to handle than a string) and should also take up less disk space 

2. Thought about a different ACL implementation.  
   This way instead of having multiple messages per different operation we have only a single one and it should also take up less disk space.


> I've only one concern about the treasury account. Since I've proposed the ownership transfer, how should we manage the treasury transfer? Make the treasury transfer all in one with the owernship transfer?

Since the `treasury` and the `owner` address are two separate things, the subspace manger can decide whether to edit them together or separately. In any case, both of these fields can be edited using a `MsgEditSubspace` which allows to specify a new `treasury` and/or `owner` address. 

Allowing independent changes to such values is to allow ownership transfers that are independent from treasury transfers. This allows to represent subspaces with a single owner but a group of people representing the treasury. This might be the case for existing social networks such as Facebook where the owner is represented by `Mark Zuckerberg` but the `treasury` is represented by a multisig wallet (i.e. the board of Meta). If we forced the change of the `owner` and `treasury` together we would end up not allowing to change the CEO without the board, which might not always be the case.

dadamu:
Just have an idea about treasury. Could the subspace have a specific unique address but the the pubkey of it can be updated by something like `MsgEditSubspaceTreasury`?

```
treasury address
pubkeyA

Update
|
v

treasury address
pubkeyB
```

bragaz:
@dadamu This is interesting and we I think we can include it directly within the `MsgEditSubspace` since it will handle all the related edits. Do you have any reference/docs of Solana so we can take a look on how it works there?

RiccardoM:
@dadamu Unfortunately, that is not yet possible within Cosmos. There's [ADR-034](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-034-account-rekeying.md) which aims at implementing it. Thus it will be possible in the future once that's done. 

However, I think Cosmos will also provide better ways of handling a treasury such as **group accounts** ([ADR-032](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-042-group-module.md)). 

I think that leaving the `treasury` as a simple account will allow us to give the opportunity to subspace admins to use all those new things once they are implemented without having us to implement them by hand in advance.

bragaz:
@RiccardoM Oh pity. So for the moment anyone who's transferring the treasury will need to tokens from the treasury account to another.

docs/architecture/adr-006-subspace-module.md
```diff
@@ -0,0 +1,159 @@
...
+  // the creation time of the subspace
+  CreationTime time.Time
+}
+```
+
+#### ACL
+In order to easily implement an ACL, we will use a simple set of keys made as follows: 
+```
+ACLPrefix + User Address -> ACL Value
```
dadamu, line 59: I guess it should be `ACLPrefix + SubspaceID + User Address` since the ACL is based on the subspace.

bragaz: @dadamu Yes you are right. Otherwise without it once we cannot properly iterate over the mods/admin of a particular subspace.
Maybe @RiccardoM you implicit thought that the `subspaceID` is included in the ACLPrefix?
I suggest to have something like:
* `ACLSubspacePrefix` = `ACLPrefix + Subspace` which will make it easy to iterate over all the users of a subspace.
* `ACLUserKey` = `ACLSubspacePrefix + User Address` to check on a specific user ACL value

docs/architecture/adr-006-subspace-module.md
```diff
@@ -0,0 +1,161 @@
...
+  // the creation time of the subspace
+  CreationTime time.Time
+}
+```
+
+#### ACL
+In order to easily implement an ACL, we will use a simple set of keys made as follows: 
+```
+ACLPrefix + Subspace ID + User Address -> ACL Value
+```
```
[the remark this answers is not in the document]
RiccardoM: @dadamu I like this idea, can you give a practical example of how that would work?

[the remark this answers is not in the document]
RiccardoM: Thank you for the explanation and the references @dadamu. I just edited the ADR with a new system that could simplify things a bit allowing to have user groups with custom permissions as well as individual permissions for a single user. Let me know what you think about it

docs/architecture/adr-006-subspace-module.md
```diff
@@ -0,0 +1,221 @@
...
+* Edit a subspace
+
+**Content management**
+* Delete contents that do not respect the ToS
+
+**User management**
+* Create a new group
+* Delete a group
+* Set group permissions
+* Set user permissions
```
dadamu, line 125: Is that `Set user permissions` including `Add user to a group`? If not, I think we should add it.

RiccardoM: Thanks for pointing this out. I should have added all the methods necessary. Let me know if something else is missing

dadamu (approved):
Everything is good to me.

RiccardoM:
I have just updated the permissions section as I've noticed something was a little bit of with the keys used. I've also specified how permissions should be checked.

bragaz:
Personally I think this is a very good way to implement subspaces.
Shootout to @dadamu to have bring out the Linux permission model. I've read the reference and now the whole implementation proposal looks very clear. I don't think there's anything missing atm so, we can proceed with the implementation whenever you guys want.

bragaz (approved):
Personally I think this is a very good way to implement subspaces.
Shootout to @dadamu to have bring out the Linux permission model. I've read the reference and now the whole implementation proposal looks very clear. I don't think there's anything missing atm so, we can proceed with the implementation whenever you guys want.

Issue #633, opened 2021-09-29
Support multisig address inside MsgCreateChainLink


ryuash:
## Feature description 

I would like to link my multisig addresses to my profile

dadamu:
@RiccardoM @bragaz 
The chain-link in Desmos also does not serve multisig account as well, the proto of chain-link should be modified to handle it.
What do you think?

RiccardoM:
@dadamu Can we have a specification of how it should be edited?

dadamu:
@RiccardoM Okay, let me open an ADR and research more about multisig.

dadamu:
@RiccardoM @bragaz Before completing the adr, I would like to discuss with what I researched and the idea how to implement the feature.

In general, there are two problem which should be solved:

### 1. Multisig address regeneration
it is impossible to confirm if the multisig address is right by pubkeys of signers. 
For instance, there is a multisig address generated by 3 keypairs with threshold which is 2, then the create-chain-link which the user prepared is signed with only 2 pubkey. There is a problem that Desmos can't regenerate multisig address since it contains only 2 pubkey. The way of multisig generation is [here]((https://github.com/cosmos/cosmos-sdk/blob/be4a965599dfae6992451e4b5135c487ed45053b/crypto/keys/multisig/multisig.go#L21)).

To solve it, the chain-link-json must be signed by all keypairs involved the multisig generation.

### 
(The verification process)[https://github.com/cosmos/cosmos-sdk/blob/be4a965599dfae6992451e4b5135c487ed45053b/crypto/keys/multisig/multisig.go#L50] of multisig signature is also big different from the single keypair. The multisignature contains a list of pubkeys, and a list of signatures generated by them.
The process is like:
1. Check the number of signature and pubkeys are more than threshold
2. Iterate signatures and verify it with pubkeys

It means we need to implement the feature to verify and store the [multi signatures](https://github.com/cosmos/cosmos-sdk/blob/be4a965599dfae6992451e4b5135c487ed45053b/types/tx/signing/signature_data.go#L24).

Let me know what do you think, @RiccardoM @bragaz.

RiccardoM:
Thanks @dadamu for the details explanation of where the problems may lie with multisig accounts. I've taken some time to dig deeper inside the Cosmos SDK codebase, in order to understand how the multisig are handled for transaction and how we can leverage the same process for our case. Here is what I got from my journey: 

[...]

PR #708, opened 2021-12-22
feat: ADR-005: Support multisig chain link

dadamu:
## Description

Closes: #633 
This PR implements ADR-005.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


dadamu:
Hi, @RiccardoM @bragaz 
It seems that we can not use `SignatureDescriptor_Data` in the `Proof` directly, The error happens when `msg.GetSignBytes()` is called, which is 
```
--- FAIL: TestMsgLinkChainAccount_GetSignBytes (0.00s)
panic: Unregistered interface signing.isSignatureDescriptor_Data_Sum [recovered]
        panic: Unregistered interface signing.isSignatureDescriptor_Data_Sum
``` 
because `signing.isSignatureDescriptor_Data_Sum` doesn't register to the `AnimoCodec`. In addition, the interface is a private interface generated by proto automatically.

I propose that we can convert it to hex string from the proto, instead.

What do you think?

RiccardoM:
@dadamu What if we use the `SignatureData` interface instead, and encode it using an `Any` instance the same way we do with `MsgLinkChainAccount#AddressData`?

dadamu:
@RiccardoM Unfortunately, `SignatureData` doesn't implement `proto.Message`. Besides, the structs of it are not the proto as well. 
So it can not be used in the `Proof` proto object.

RiccardoM:
@dadamu I have just finished experimenting with this problem, and I think that we can do as follows: 

**1.** Create a new interface called `SignatureData` that extends `proto.Message` and has a `isSignatureDataMethod`
**2.** Create two new Proto definitions: `SingleSignatureData` and `MultiSignatureData` that contain the same fields of the Cosmos ones. The only difference is that our `MultiSignatureData` will have its `Signatures` field that is going to be an array of  `codectypes.Any` rather than an array of `signing.SignatureData`.
**3.** Add a method called `SignatureDataToCosmosSinatureData` that takes a `SignatureData` interface instance and returns a `signing.SignatureData` instance. 

Then, we only need to register the new `SignatureData` interface and its implementations inside our Amino instance and everything should work fine. 

This will allow us to use `proto.Any` as the `Signature` type inside the `Proof` object, and serialize it both using Amino and Protobuf.  

I have tried implementing all of this and pushed the changes to a new branch named `riccardo/adr-005-multisig-chainlink-impl-test`. You can take a look at it and pull the changes in your branch if you find them good enough

dadamu:
@RiccardoM Okay, got it. will extend the interface with the same way as `AddressData` to solve the issue.

x/profiles/types/models_chain_links.go
```diff
@@ -61,16 +69,11 @@ func (p Proof) Validate() error {
 		return fmt.Errorf("public key field cannot be nil")
 	}
 
-	_, err := hex.DecodeString(p.Signature)
-	if err != nil {
-		return fmt.Errorf("invalid hex-encoded signature")
-	}
```
RiccardoM, line 67: Probably we should add a `Validate` method to the `SignatureData` interface so that we can call it here and make sure the signature is valid

x/profiles/types/models_chain_links.go
```diff
@@ -109,7 +145,114 @@ func (p Proof) Verify(unpacker codectypes.AnyUnpacker, address AddressData) erro
...
+	switch data := data.(type) {
+	case *signing.SingleSignatureData:
+		return &SingleSignatureData{
+			Mode:      data.SignMode,
+			Signature: data.Signature,
+		}
+	case *signing.MultiSignatureData:
+		sigAnys := make([]*codectypes.Any, len(data.Signatures))
+		for i, data := range data.Signatures {
+			sigAny, _ := codectypes.NewAnyWithValue(SignatureDataFromCosmosSignatureData(data))
```
RiccardoM, line 201: We should check the error here to make sure it's `nil` and the signature is encoded properly. The `SignatureDataFromCosmosSignatureData` should return `(SignatureData, error)`

app/desmos/cmd/chainlink/types/getter.go
```diff
@@ -13,8 +13,13 @@ import (
 
 // ChainLinkReferenceGetter allows to get all the data needed to generate a ChainLinkJSON instance
 type ChainLinkReferenceGetter interface {
-	// GetMnemonic returns the mnemonic
-	GetMnemonic() (string, error)
+
+	// GetIsSingleSignatureAccount returns if the target account is single signature account
+	GetIsSingleSignatureAccount() (bool, error)
```
bragaz, line 18:
```suggestion
	IsSingleSignatureAccount() (bool, error)
```
I would rename this like this to make it clearer

testutil/profiles.go
```diff
@@ -60,3 +63,31 @@ func ProfileFromAddr(address string) *types.Profile {
 
 	return profile
 }
+
+func SingleSignatureProtoFromHex(s string) types.SignatureData {
```
bragaz, line 67: can you add a line of docs to this function?

testutil/profiles.go
```diff
@@ -60,3 +63,31 @@ func ProfileFromAddr(address string) *types.Profile {
...
+	if err != nil {
+		panic(err)
+	}
+	return &types.SingleSignatureData{
+		Mode:      signing.SignMode_SIGN_MODE_DIRECT,
+		Signature: sig,
+	}
+}
+
+func MultiSignatureProtoFromHex(unpacker codectypes.AnyUnpacker, s string) types.SignatureData {
```
bragaz, line 78: as above

x/profiles/legacy/v230/models_chain_links.go
```diff
@@ -0,0 +1,345 @@
...
+
+// AddressData is an interface representing a generic external chain address
+type AddressData interface {
+	proto.Message
+
+	// Validate checks the validity of the AddressData
+	Validate() error
+
+	// GetValue returns the address value
+	GetValue() string
```
bragaz, line 125:
```suggestion
	GetAddress() string
```
As it's more explicit

dadamu: This file is the copy paste of the current version for the v210 migration since the new updates will break the proto structure.

app/desmos/cmd/chainlink/types/getter.go
```diff
@@ -72,6 +108,52 @@ func (cp ChainLinkReferencePrompt) GetFilename() (string, error) {
...
+		return "", err
+	}
+	prompt := promptui.Prompt{
+		Label:   "Please insert the path of multisigned tx file (fully qualified path)",
+		Default: path.Join(wd, "tx.json"),
+	}
+	return prompt.Run()
+}
+
+// getMnemonic asks the user the mnemonic and then returns it
```
bragaz, line 142: wrong docs here

x/profiles/legacy/v230/models_chain_links.go
```diff
@@ -0,0 +1,345 @@
+package v230
```
dadamu, line 1: v230 is the copy paste of the latest chain link file for the v210 store_test.

x/profiles/module.go
```diff
@@ -87,6 +88,7 @@ func (AppModuleBasic) GetQueryCmd() *cobra.Command {
 
 // RegisterInterfaces registers interfaces and implementations of the profiles module.
 func (AppModuleBasic) RegisterInterfaces(registry codectypes.InterfaceRegistry) {
+	v230.RegisterInterfaces(registry)
```
dadamu, line 91: This is for v210 store_test as mentioned in v230 files.

app/desmos/cmd/chainlink/create_json.go
```diff
@@ -42,15 +47,37 @@ The mnemonic is only used temporarily and never stored anywhere.`,
...
+				mnemonic, err := getter.GetMnemonic()
+				if err != nil {
+					return err
+				}
 
-			// Build che chain link JSON
-			chainLinkJSON, err := generateChainLinkJSON(mnemonic, chain)
-			if err != nil {
-				return err
+				// Build che chain link JSON
```
bragaz, line 58:
```suggestion
				// Build the chain link JSON
```

RiccardoM:
@dadamu I've improved the code organization by splitting apart the various getters and creating a new builder interface. The multi-sig chain link test is however failing. Can you check why please? I didn't change pretty much anything except forcing the `Sequence` and `AccountNumber` to `0` when building the `SignatureData` instance

dadamu:
@RiccardoM Fixed.

app/desmos/cmd/chainlink/builder/multi/builder.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Parse the transaction
+	parsedTx, err := txCfg.TxJSONDecoder()(bytes)
+	if err != nil {
+		return utils.ChainLinkJSON{}, err
+	}
+
+	// Get the sign mode
+	signMode := signing.SignMode_SIGN_MODE_LEGACY_AMINO_JSON
```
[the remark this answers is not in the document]
dadamu: They are all signed in `SIGN_MODE_LEGACY_AMINO_JSON` if it is the multi-sig tx as well as the ledger tx.

[Reference](https://github.com/cosmos/cosmos-sdk/blob/2b02f066aae39cd2ece720e2616acf5b3308bf8e/x/auth/client/tx.go#L43)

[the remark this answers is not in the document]
RiccardoM: @dadamu Perfect, thanks for the reference link as well! :pray:

RiccardoM (approved):
I've just updated the `create-chain-link-json` documentation to include extensive explanation on what it's required when using it with multi-sig accounts. I've also added the missing docs to the `generatePubKeyAndMultiSignatureData` testing method. 

I think the implementation is correct and ready to go.

bragaz (approved):
Ready to ship

PR #723, opened 2022-01-17
feat: ADR-004: Application links expiration time

bragaz:
## Description

Closes: #562 

This PR introduce a new field `ExpirationTime` in the `AppLinks` object. 
It will allow to automatically delete app links when they expires. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/profiles/module.go
```diff
@@ -186,7 +190,9 @@ func (AppModule) ConsensusVersion() uint64 {
 }
 
 // BeginBlock returns the begin blocker for the profiles module.
-func (am AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {
+func (am AppModule) BeginBlock(ctx sdk.Context, _ abci.RequestBeginBlock) {
+	// TODO should we emit an event for this?
+	am.keeper.DeleteExpiredApplicationLinks(ctx)
```
bragaz, line 195: should we emit some events for this?

x/profiles/types/models_params.go
```diff
@@ -16,12 +17,13 @@ const (
...
-	DefaultMinDTagLength     = sdk.NewInt(3)
-	DefaultMaxDTagLength     = sdk.NewInt(30)
-	DefaultMaxBioLength      = sdk.NewInt(1000)
+	DefaultMinNicknameLength      = sdk.NewInt(2)
+	DefaultMaxNicknameLength      = sdk.NewInt(1000) // Longest name on earth count 954 chars
+	DefaultRegEx                  = `^[A-Za-z0-9_]+$`
+	DefaultMinDTagLength          = sdk.NewInt(3)
+	DefaultMaxDTagLength          = sdk.NewInt(30)
+	DefaultMaxBioLength           = sdk.NewInt(1000)
+	DefaultAppLinksExpirationTime = time.Hour * 24 * 7 * 4 * 6 // This duration is equal to roughly 5.5 months time in minutes 241920
```
bragaz, line 26: I take the `DefaultUnbondingTime` param as an example:
https://github.com/cosmos/cosmos-sdk/blob/6ea204994443edc7841df4f0dc86ee5b4edc8377/x/staking/types/params.go#L21

.github/workflows/bench.yml
```diff
@@ -0,0 +1,42 @@
+name: Benchmarks
```
bragaz, line 1: This workflow uses [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat#section-readme) a common go tool for comparing different benchmarks results.
In our case, the workflow will compare the `master` branch benchmark's result with the ones of a new PR which is IMO very useful to let us understand if we have introduced a performance regression.
In order to make this properly working, bench tests should be written when we consider it's the case.

.github/workflows/on-chain-upgrade.yml
```diff
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
-      GENESIS_DESMOS_VERSION: "v2.3.0-mainnet"
-      GENESIS_URL: "https://raw.githubusercontent.com/RiccardoM/desmos-states/master/desmos-mainnet-1160425.json"
-      UPGRADE_NAME: "v2.3.1"
+      GENESIS_DESMOS_VERSION: "v2.3.1"
```
bragaz, line 25: This has been edited in order to test an on-chain-upgrade with the migration of the `ApplicationLinks` store and `params`

bragaz:
In the meantime I figure out how to fix the failing checks problems, you can start reviewing this @RiccardoM @dadamu

proto/desmos/profiles/legacy/v200/models_app_links.proto
```diff
@@ -0,0 +1,155 @@
+syntax = "proto3";
+package desmos.profiles.legacy.v200;
+
+option go_package = "github.com/desmos-labs/desmos/v2/x/profiles/legacy/v200";
```
dadamu, line 4: Why v200 but not v230? I think the legacy version should be the last version.

bragaz: because it was since v2.0.0 that the `appLinks` had not been changed. So I thought they should belong to that particular version. Should we move this to v2.3.0?

dadamu: You are right, I thought it means the legacy `appLink` is available until v2.3.0.  Then v2.3.0 chain link should be v2.1.0...

x/profiles/keeper/alias_functions_test.go
```diff
@@ -265,6 +271,68 @@ func (suite *KeeperTestSuite) TestKeeper_GetApplicationLinks() {
...
+	)
+	ctx, _ := suite.ctx.CacheContext()
+
+	for _, link := range links {
+		suite.ak.SetAccount(ctx, testutil.ProfileFromAddr(link.User))
+		err := suite.k.SaveApplicationLink(ctx, link)
+		suite.Require().NoError(err)
+	}
+
+	// save an expiration key referring to an already deleted AppLink to test lines 225-227
```
dadamu, line 319: It should not include the lines in the reference since it might be changed in the future.

x/profiles/keeper/alias_functions.go
```diff
@@ -211,6 +210,32 @@ func (k Keeper) GetApplicationLinks(ctx sdk.Context) []types.ApplicationLink {
 	return links
 }
 
+// IterateExpiringApplicationLinks iterates through all the expiring application links references.
+// The key will be skipped and deleted if the application link has already been deleted.
+func (k Keeper) IterateExpiringApplicationLinks(ctx sdk.Context, fn func(index int64, link types.ApplicationLink) (stop bool)) {
+	store := ctx.KVStore(k.storeKey)
+
+	iterator := sdk.KVStorePrefixIterator(store, types.ApplicationLinkExpiringTimePrefix(ctx.BlockTime()))
```
dadamu, line 218: It is needed the block produced at the exactly time when applinks expired, right? 
If yes, it will not work properly because the block time is not always at time, says: a applink expired at `time.Date(2022, 1, 1, 00, 00, 00, 000, time.UTC)`, but actually the block might be produced at `time.Date(2022, 1, 1, 00, 00, 00, 001, time.UTC)`.

dadamu: I think iterating the `ExpiringAppLinkTimePrefix` in the asc way then deleting the applinks until the time is larger the given time would solve the issue since key in the db is sorted.

proto/desmos/profiles/legacy/v200/models_app_links.proto
bragaz: @dadamu actually also your thought makes sense. @RiccardoM How should we handle legacy code?

RiccardoM: I'm actually looking into this problem inside #730 by reviewing where different things are located for legacy stuff and updates

x/profiles/simulation/decoder_test.go
```diff
@@ -80,6 +90,7 @@ func TestDecodeStore(t *testing.T) {
 		{"DTag transfer request", fmt.Sprintf("RequestA: %s\nRequestB: %s\n", request, request)},
 		{"Relationship", fmt.Sprintf("Relationships A: %s\nRelationships B: %s\n", relationship, relationship)},
 		{"User block", fmt.Sprintf("User block A: %s\nUser block B: %s\n", userBlock, userBlock)},
+		{"ClientID", fmt.Sprintf("Client ID A: %s\nClient ID B: %s\n", clientID, clientID)},
```
RiccardoM, line 93: Should this be named "Expiring application link"?
```suggestion
		{"Expiring application link", fmt.Sprintf("Client ID A: %s\nClient ID B: %s\n", clientID, clientID)},
```

x/profiles/module.go
RiccardoM: Yes, we should emit one event for each deleted link. I don't know whether it would be better to do this here or inside the `DeleteExpiredApplicationLinks` method though

x/profiles/types/models_params.go
bragaz: Since we can't sum decimals in time params calculation I've added an `ExpirationTimeCorrectionFactor` that I calculated as follow:
-  4382.91 hours are 182 days;
- `DefaultExpirationTime` were 168 days;
- 182 - 168 = 14 days = 20160 minutes;
- `DefaultAppLinksExpirationTime` + 20160 minutes = 262080 = 5.9835 Months.

x/profiles/types/models_params.go
```diff
@@ -257,3 +267,28 @@ func ValidateOracleParams(i interface{}) error {
...
+	return NewAppLinksParams(DefaultAppLinksExpirationTime)
+}
+
+func ValidateAppLinksParams(i interface{}) error {
+	params, isAppLinksParams := i.(AppLinksParams)
+	if !isAppLinksParams {
+		return fmt.Errorf("invalid parameters type: %s", i)
+	}
+
+	if params.ExpirationTime <= 0 {
```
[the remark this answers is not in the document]
bragaz: I've set a `FourteenDays` const that I both use for this check and as a correction factor for the `DefaultAppLinksExpirationTime`

x/profiles/module.go
bragaz: I think we can follow how we did with other keeper's methods and put it inside the `DeleteExpiredApplicationLinks` method.

contrib/upgrade_testnet/start.sh
```diff
@@ -15,7 +15,17 @@ rm -r -f $BUILDDIR
 
 # Create the 4 nodes folders with the correct denom
 echo "===> Creating $NODES nodes localnet"
-make setup-localnet COIN_DENOM="udaric" NODES=$NODES > /dev/null > /dev/null
```
bragaz, line 18: @RiccardoM I tried editing this in order to make it use the docker compiled binary for the nodes, instead of using setup-localnet that use make build-linux (without linking the correct static library needed).
I'm now having some errors inside the workflow, in particular these:
https://github.com/desmos-labs/desmos/runs/5035433674?check_suite_focus=true#step:5:330

https://github.com/desmos-labs/desmos/runs/5035433674?check_suite_focus=true#step:5:385

https://github.com/desmos-labs/desmos/runs/5035433674?check_suite_focus=true#step:5:388

https://github.com/desmos-labs/desmos/runs/5035433674?check_suite_focus=true#step:5:391

https://github.com/desmos-labs/desmos/runs/5035433674?check_suite_focus=true#step:5:394

I'm not sure why these are happening 🤔 I've also tried to add sudo in front of the cmd but it din't work.
Can you take a look on this?

RiccardoM: First of all, I would suggest you removing the usage of `--gentx-coin-denom="udaric"` since it's not necessary. The `setup-genesis.py` script that is run will take care of replacing all the `stake` coin denoms with the one you provide with `COIN_DENOM` so that's sufficient without editing the previous `testnet` command. 

That being said, this error is due to file permissions. You need to run the Docker command passing to it the same user that owns all those files. You can do so by adding the `--user $UID:$GID` to the `docker run` command. 

Finally, please don't use `sudo` on Docker commands, and try using `make` commands (eg. `make setup-localnet`) instead of duplicating the code by copy/pasting the commands

bragaz: @RiccardoM I would like to use the `make` one, but it looks like it's not possible since we need to include the static library during the build process.

x/profiles/keeper/keeper_app_links.go
```diff
@@ -122,3 +130,20 @@ func (k Keeper) DeleteAllUserApplicationLinks(ctx sdk.Context, user string) {
 		store.Delete(types.UserApplicationLinkKey(link.User, link.Data.Application, link.Data.Username))
 	}
 }
+
+// DeleteExpiredApplicationLinks deletes all the expired application links in the given context
+func (k Keeper) DeleteExpiredApplicationLinks(ctx sdk.Context) {
+	k.IterateExpiringApplicationLinks(ctx, func(_ int64, link types.ApplicationLink) (stop bool) {
+		ctx.EventManager().EmitEvent(
```
dadamu, line 137: It would create a lot of index to a block, should we emit only event called `EventTypeApplicationLinksDeleted` outside of iterator instead?

bragaz: 🤔 This make sense to me. @RiccardoM any comment about it?

dadamu:
@bragaz I tried to print address which occurs the on-chain-upgrade error. The address is desmos1y6rzsgdh09em2mptk6juk5unsjsudaric9dmfwsa, but it should be `desmos1y6rzsgdh09em2mptk6juk5unsjsudsm9dmfwsa`.
I guess something change bech32 `udsm` string into `udaric` accidentally.

RiccardoM:
I've updated all the code and how the migrations are handled to work best. I would like to ask @dadamu to review the code and also fix the now failing test (it's always the one related to the JSON within the profiles module). Once that's done I think we can merge this

RiccardoM (approved):

x/profiles/keeper/alias_functions.go
```diff
@@ -131,6 +133,41 @@ func (k Keeper) GetApplicationLinks(ctx sdk.Context) []types.ApplicationLink {
...
+// The key will be skipped and deleted if the application link has already been deleted.
+func (k Keeper) IterateExpiringApplicationLinks(ctx sdk.Context, fn func(index int64, link types.ApplicationLink) (stop bool)) {
+	store := ctx.KVStore(k.storeKey)
+
+	iterator := sdk.KVStorePrefixIterator(store, types.ExpiringAppLinkTimePrefix)
+	defer iterator.Close()
+
+	i := int64(0)
+	for ; iterator.Valid(); iterator.Next() {
+		trimmedPrefixKey := bytes.TrimPrefix(iterator.Key(), types.ExpiringAppLinkTimePrefix)
```
dadamu, line 146: Can you add a comment to describe `Key` and `Value` of iterator representing here?
It would be helpful to understand this closure.

RiccardoM: Sure, added!

x/profiles/keeper/keeper_app_links_benchmarks_test.go
```diff
@@ -0,0 +1,128 @@
...
+		ak.SetAccount(ctx, profilestesting.ProfileFromAddr(link.User))
+		err := k.SaveApplicationLink(ctx, link)
+		require.NoError(b, err)
+	}
+
+	b.ResetTimer()
+	b.Run("iterate and delete expired links", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			b.ReportAllocs()
+			profiles.BeginBlocker(ctx, k)
```
dadamu, line 125: This method is out of keeper range, it would be better to have a specified function like `DeleteExpiredApplicationLinks`.

RiccardoM: I've moved it to be inside the `profiles` package as it should be (since it's testing the `BeginBlocker` method). I removed the `DeleteExpiredApplicationLinks` as in my opinion it was pointless to have that method only so that we could write benchmark tests when we could simply have done it this way

x/profiles/keeper/migrations.go
```diff
@@ -32,7 +30,12 @@ func (m Migrator) Migrate4to5(ctx sdk.Context) error {
...
 
-// Migrate5To6 migrates from version 5 to 6.
-func (m Migrator) Migrate5To6(ctx sdk.Context) error {
+// Migrate5to6 migrates from version 5 to 6.
+func (m Migrator) Migrate5to6(ctx sdk.Context) error {
 	return v5.MigrateStore(ctx, m.keeper.storeKey, m.keeper.cdc, m.keeper.legacyAmino)
 }
+
+// Migrate6to7 migrates from version 6 to 7.
+func (m Migrator) Migrate6to7(ctx sdk.Context) error {
```
dadamu, line 39: I am confused the version, `v6.MigrateStore` is actually migrating to v6 from v5 although the method is called `Migrate6to7`. Is that consensus version different from legacy store version?

RiccardoM: That was actually an error within comments themselves. I've now updated them to make it more clear. Generally, the `Migrate<X>to<X+1>` calls the `v<X>` package that handles the migration from `X` to `X+1` properly. So in this case `Migrate6to7` calls the `v6` package that migrates from `6` to `7` (latest version). 

The only thing wrong here was the comment on the `v6.MigrateStore` function. It migrates from `6` to `7` (and not `5` to `6` as it was written). I've now fixed it

dadamu: @RiccardoM Thanks for the explanation. But it would be another problem that `v6.MigrateStore` actually migrates `v5types` to the current type `v6` [like this](https://github.com/desmos-labs/desmos/blob/628932d7360e3552fc6ee0f6a37a909f0e4070e4/x/profiles/legacy/v6/store.go#L50).

x/profiles/types/codec.go
```diff
@@ -40,14 +40,14 @@ func RegisterInterfaces(registry types.InterfaceRegistry) {
 	registry.RegisterImplementations((*exported.VestingAccount)(nil), &Profile{})
 	registry.RegisterImplementations((*authtypes.GenesisAccount)(nil), &Profile{})
 	registry.RegisterInterface(
-		"desmos.profiles.v2.AddressData",
+		"desmos.profiles.v3.AddressData",
```
dadamu, line 43: Another question related to version. Should we set the version here to consensus version? e.g., `desmos.profiles.v7.AddressData`.

dadamu: As well as the [query.proto](https://github.com/desmos-labs/desmos/blob/leonardo/application-links-expiration-impl/proto/desmos/profiles/v3/query.proto#L23).

RiccardoM: No, these types should reflect the Protobuf versioning (which is different from the module consensus versioning). We should update them only when we change the Protobuf definitions, as they change less often than the module consensus version (see from v4 to v5 where we didn't change a single Protobuf version but we had to do a migration to add new keys)

x/profiles/legacy/v6/store_test.go
```diff
@@ -0,0 +1,252 @@
...
+					},
+					time.Date(2020, 1, 1, 0, 0, 0, 00, time.UTC),
+					profilestesting.AccountFromAddr("cosmos1nejmx335u222dj6lg7qjqrufchkpazu8e0semf"),
+				)
+				require.NoError(t, err)
+				authKeeper.SetAccount(ctx, profile)
+			},
+			check: func(ctx sdk.Context) {
+				// Check the profile to make sure it contains the same data
+				v6Profile, err := types.NewProfile(
```
dadamu, line 107:
```suggestion
				v7Profile, err := types.NewProfile(
```

x/profiles/legacy/v6/store_test.go
```diff
@@ -0,0 +1,252 @@
...
+				)
+				require.NoError(t, err)
+
+				sdkAddr, err := sdk.AccAddressFromBech32("cosmos1nejmx335u222dj6lg7qjqrufchkpazu8e0semf")
+				require.NoError(t, err)
+
+				account := authKeeper.GetAccount(ctx, sdkAddr)
+				profile, ok := account.(*types.Profile)
+				require.True(t, ok)
+				require.Equal(t, v6Profile, profile)
```
dadamu, line 123:
```suggestion
				require.Equal(t, v7Profile, profile)
```

Issue #724, opened 2022-01-18
Add more address hash method  for `AddressData`


dadamu:
## Feature description 

Currently, Desmos chain link supports bech32, base58 and hex address, but each of them utilize only one hash algorithm to generate address from the key, says:
Bech32 address is generated by cosmos default algorithm, which is simple single address use the function `ripemd160(sha256(compressed public key))[:20]` while legacy animo multisig address use the function `sha256(aminoCdc.Marshal(pk))[:20]`.
Base58 address is generated by `compressed public` key directly, without any hash function.
Hex address is generated by the function `keccak64(uncompressed public key)[12:]`

To support more chain with more, we should support more address generation algorithms on each address type.

## Implementation proposal

I proposed using `strategy` design pattern on the AddressData structure. 
For instance on the `HexAddress`, Desmos can have a `AddressStrategy` interface like:
```go
type AddressStrategy interface{
    GenerateAddress(cryptotypes.PubKey) []byte
}
```
Then add the strategy field into `HexAddress` which can be called when `VerifyPubKey(key cryptotypes.PubKey)`:
```go
type HexAddress struct {
    Value     string
    Prefix    string
    Strategy  AddressStrategy
}

func (h HexAddress) VerifyPubKey(key cryptotypes.PubKey) (bool, error) {
	addr := h.Value[len(h.Prefix):]
	bz, err := hex.DecodeString(addr)
	if err != nil {
		return false, err
	}
	bzFromPubKey := h.Strategy.GenerateAddress(key)
	return bytes.Equal(bzFromPubKey, bz), err
}
```

Please tell me your ideas. @RiccardoM @bragaz

RiccardoM:
It makes sense to me, @dadamu would you mind opening an ADR for these changes?

bragaz:
Straight forward to me too. 👍

1408eduardo:
Good

PR #727, opened 2022-01-24
docs(adr): ADR-007: Address generation of chain link proof

dadamu:
## Description

References: #724

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+Subsequently, the `AddressData` instance like `Bech32Address` can include the any value of the strategy instance which implements `AddressGenerationStrategy` as a property.
+```proto
+message Bech32Address{
+    ...
+    google.protobuf.Any strategy = 3 [(cosmos_proto.accepts_interface) = "AddressGenerationStrategy"];
+}
+```
```
RiccardoM, line 41: Honestly I'm not a big fan if this approach. We should use an `interface` (and the related serialization strategy - `Any`) only when there are different implementations that need different data to be created. This was true for `AddressData`: 
- `Bech32AddressData` requires a prefix and a value 
- `HexAddressData` requires only a value
- `Base58AddressData` requires only a value

In this case, each address generation strategy does not require anything else aside from the public key, which is passed to the `GenerateAddress` method itself. For this reason, I think the best alternative is to simply add a new method to the `AddressData` interface instead: 
```go
// AddressData is an interface representing a generic external chain address
type AddressData interface {
    proto.Message

    // Validate checks the validity of the AddressData
    Validate() error

    // GetValue returns the address value
    GetValue() string

    // GetAddressGenerationStrategy returns the address generation strategy to be used
    GetAddressGenerationStrategy() AddressGenerationStrategy

    // VerifyAddress verifies that the given address is associated with this address data
    VerifyAddress(address []byte) (bool, error)
}
```

As you can see, I've also changed `VerifyPubKey` to `VerifyAddress`. Now the flow should become: 
1. generate the proper address using the associated strategy;
2. verify that the address matches the address data value.

```go
address, err := GenerateAddress(b.GetAddressGenerationStrategy(), key)
if err != nil {
    return err
}

valid, err := addressData.VerifyAddress(bz)
if err != nil {
    return err
}
```

The `GenerateAddress` is going to be a public method that simply takes an `AddressGenerationStrategy`, a public key and generates the address properly: 
```go
func GenerateAddress(strategy AddressGenerationStrategy, pubKey cryptotypes.PubKey) (address []byte, err error) {
    switch strategy { 
        case StrategyRipemd160:
            return generateRipemd160Address(pubKey)
        // ...
    }
```

This implementation will also require to change all the `AddressData` instances constructors to allow passing an (optional) address generation strategy

bragaz: I'll wait for this to be included in the ADR and then review it. Notify me when ready @dadamu.

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,79 @@
...
+- January 20th, 2022: Initial draft;
+
+## Status
+
+PROPOSED
+
+## Abstract
+
+Currently, Desmos chain link supports `Bech32`, `Base58` and `Hex` address formats. Besides, each of them only couples one address generation algorithms. 
+Since Desmos will support more addresses from other chains, which might apply the supported address format but with the unsupported algorithm, we SHOULD split the address algorithm from `AddressData` instances.
```
RiccardoM, line 14: > which might apply the supported address format but with the unsupported algorithm

I was wondering: do we have such examples?

dadamu: [elrond address](https://docs.elrond.com/technology/glossary/) is a bech32 format, but it encode public key into bech32 directly instead of `ripemd160(sha256(pubKey))` as cosmos ecosystem.

RiccardoM:
@dadamu @bragaz I've completely reviewed this ADR as I think there might be an easier way to go to support different algorithm strategies and encodings.

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,157 @@
...
+## Decision
+
+In order to allow developers to integrate any kind of address, we will review how the current `Address` structure is made. Instead of using an interface with multiple implementations based on the encoding algorithm, we will use a single structure that allows to specify both the encoding algorithm and the hashing algorithm(s) that should be used to get the proper address: 
+
+```protobuf
+syntax = "proto3";
+
+// Address contains the data of an external address
+message Address {
+  // Hex-encoded value of the address
```
RiccardoM, line 51: Instead of requesting the developers to put the hex-encoded address bytes here, we can also allow them to put the already-encoded address using the specified encoding algorithm. Then, inside the `Validate` check we simply: 
1. generate the addresss bytes from the public key; 
2. encode the address using the provided encoding method; 
3. check if the resulting encoded address is equals to the provided value

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,157 @@
...
+  string value = 1;
```
dadamu, line 52: We can name it `hex_value` since it is hex encoded of address bytes.

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,157 @@
...
+  
+  // Algorithm that has been used in order to generate the address starting from the public key bytes
+  GenerationAlgorithm generation_algorithm = 2;
+  
+  // Algorithm that needs to be used to properly encode the address 
+  google.proto.Any encoding = 3 [ (cosmos_proto.accepts_interface) = "AddressEncoding" ];
```
bragaz, line 58: I would name this similar to the above, so like:
```suggestion
  google.proto.Any encoding_algorithm = 3 [ (cosmos_proto.accepts_interface) = "AddressEncoding" ];
```

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,157 @@
...
+message Bech32Encoding {
+  option (cosmos_proto.implements_interface) = "AddressEncoding";
+  
+  // Prefix to be used
+  string prefix = 1; 
+}
+
+// Base58Encoding represents the encoding algorithm based on the Base58 format
+message Base58Encoding {
+  option (cosmos_proto.implements_interface) = "AddressEncoding";
```
dadamu, line 86: I think we should provide the `prefix` field since Polkadot base58 address format has prefix.
References:
https://wiki.polkadot.network/docs/learn-accounts#address-format

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
dadamu, line 51:
```suggestion
  // Encoded value of the address
```

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
```diff
@@ -0,0 +1,160 @@
...
+
+  // Prefix to be used
+  string prefix = 1;
```
dadamu, line 89: I think we should provide the prefix field since Polkadot base58 address format has prefix.
References:
https://wiki.polkadot.network/docs/learn-accounts#address-format

bragaz: I've nothing against this, since it brings more support for other chains.
Thoughts? @RiccardoM

RiccardoM (approved):

docs/architecture/adr-007-address-generation-of-chain-link-proof.md
RiccardoM: Nothing against, just made it optional though

bragaz (approved):

PR #728, opened 2022-01-24
feat: ADR-006: Subspaces module

RiccardoM:
## Description
This PR implements _ADR-006: Subspaces module_ adding a new `x/subspaces` module.  

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,382 @@
...
+			if err != nil {
+				return err
+			}
+
+			subspaceID, err := types.ParseSubspaceID(args[0])
+			if err != nil {
+				return err
+			}
+
+			name := args[1]
```
bragaz, line 193: I think the name of the group should be saved as case insensitive so that we can ensure that in a subspace there won't be two groups named for example: "Admins" and "admins".
To do it, you can simply use:
```go
strings.ToLower(args[1])
```

proto/desmos/subspaces/v1/subspace.proto
```diff
@@ -0,0 +1,40 @@
...
+
+  // Address of the subspace creator
+  string creator = 6 [ (gogoproto.moretags) = "yaml:\"creator\"" ];
+
+  // the creation time of the subspace
+  google.protobuf.Timestamp creation_time = 7 [
+    (gogoproto.nullable) = false,
+    (gogoproto.stdtime) = true,
+    (gogoproto.moretags) = "yaml:\"creation_time\""
+  ];
```
bragaz, line 39: It should be useful to also have a `last_edited_time`. What do you think? @RiccardoM

x/subspaces/keeper/msg_server.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+var _ types.MsgServer = msgServer{}
+
+// CreateSubspace defines a rpc method for MsgCreateSubspace
+func (k msgServer) CreateSubspace(goCtx context.Context, msg *types.MsgCreateSubspace) (*types.MsgCreateSubspaceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Get the next subspace ID
+	subspaceID, err := k.GetSubspaceID(ctx)
```
bragaz, line 31: Are you sure of this approach for the ID? I remember we first used the same in our first `posts` implementation and then we switched to a SHA-256 id to avoid duplication if I'm not wrong.

proto/desmos/subspaces/v1/subspace.proto
dadamu: Agree with it. But it sounds costly if all the changes to the subspace including `Permissions` and `Group` related operation will update `last_update_date` field.

x/profiles/keeper/keeper_blocks.go
```diff
@@ -60,13 +60,13 @@ func (k Keeper) GetAllUsersBlocks(ctx sdk.Context) []types.UserBlock {
 
 // IsUserBlocked tells if the given blocker has blocked the given blocked user
 func (k Keeper) IsUserBlocked(ctx sdk.Context, blocker, blocked string) bool {
-	return k.HasUserBlocked(ctx, blocker, blocked, "")
+	return k.HasUserBlocked(ctx, blocker, blocked, 0)
```
dadamu, line 63: The subspace id should start from 1 if we do this.

proto/desmos/subspaces/v1/subspace.proto
RiccardoM: Honestly I'm not very supportive of this idea. I don't see how a `last_edited` field would be useful for a subspace. It's not something that will be edited very often, and even if is it will be edited with very small changes. The last edited time can always be retrieved by looking at the transaction history (which also gives you the editor that applied the changes)

x/subspaces/keeper/msg_server.go
RiccardoM: Yes, I think duplication will not be a problem. The subspaces are thought to be created by people who want to create a DApp. The idea is that you want to build something on Desmos and thus you create the subspace. Later you will have all your clients that rely on that subspace that are marketed as "official" for such subspace. Even if someone creates a copy subspace, they would have to have their clients relying on the new subspace instead of the official one, which will be quite a lot of work. 

Also, the `treasury` address is used to allow the creators to verify their subspace with external applications (eg. their website) so that other people cannot create copies of it that easily. It's the same principle of app links but applied to subspaces: the more links you add, the safer and easier for users is to verify the subspace is what it is claiming to be.

x/subspaces/client/cli/tx.go
RiccardoM: You're right. I think it might be even better to create a `UserGroup` type as follows: 

```proto 
message UserGroup {
  uint32 id = 1; 
  string name = 2; 
  string description = 3
}
```

By doing so we can refer to different groups using their IDs rather than the name. We can assume that the subspace admin won't create multiple groups with different names. And even if they do, they can simply delete them later. 
But we could have an extensible type for the future if we want to add more things to it. What do you think (also @dadamu)?

x/subspaces/keeper/msg_server.go
```diff
@@ -0,0 +1,316 @@
...
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Signer),
+		),
+		sdk.NewEvent(
+			types.EventTypeEditSubspace,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", updated.ID)),
```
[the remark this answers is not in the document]
RiccardoM: This wouldn't be easy to do: each message can edit different fields and so we would have to dynamically get the list of attributes to add, which is not ideal. I think it's sufficient to simply return the id of the edited subspace and let the clients get the new values after. They can always get the edited fields by looking ad the message anyway

x/profiles/legacy/v300/store.go
```diff
@@ -0,0 +1,112 @@
...
+	err = migrateRelationships(store, cdc)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// migrateUserBlocks migrates the user blocks stored to the new type, converting the subspace from string to uint64
+func migrateUserBlocks(store sdk.KVStore, cdc codec.BinaryCodec) error {
```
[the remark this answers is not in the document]
RiccardoM: This is done due to the fact that, looking the `Iterator` interface documentation we can see the following: 
> No writes can happen to a domain while there exists an iterator over it, some backends may take
> out database locks to ensure this will not happen.

So we first need to iterate over the various keys, get the values and then once the iterator is closed we can update the domain. I just noticed I need to close it before the second for loop, will fix it

x/profiles/types/msgs_relationships.go
```diff
@@ -41,10 +41,6 @@ func (msg MsgCreateRelationship) ValidateBasic() error {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "sender and receiver must be different")
 	}
 
-	if !IsValidSubspace(msg.Subspace) {
```
dadamu, line 44: I think the subspace should not be 0 since subspace 0 is the dummy subspace used for invalid error.

x/profiles/types/msgs_relationships.go
```diff
@@ -93,10 +89,6 @@ func (msg MsgDeleteRelationship) ValidateBasic() error {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "user and counterparty must be different")
 	}
 
-	if !IsValidSubspace(msg.Subspace) {
```
dadamu, line 96: Same as above.

x/profiles/types/msgs_relationships.go
```diff
@@ -146,10 +138,6 @@ func (msg MsgBlockUser) ValidateBasic() error {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "blocker and blocked must be different")
 	}
 
-	if !IsValidSubspace(msg.Subspace) {
```
dadamu, line 149: Same as above

x/profiles/types/msgs_relationships.go
```diff
@@ -198,10 +186,6 @@ func (msg MsgUnblockUser) ValidateBasic() error {
...
```
dadamu, line 201: Same as above

x/profiles/types/msgs_relationships.go
RiccardoM: Subspace with ID `0` is not actually used as a dummy subspace for invalid errors. It's the subspace that identifies the entire set of subspaces. So for example if someone blocks you on subspace `0` they are blocking you on **all** subspaces. We could call it the "Desmos" subspace.

proto/desmos/subspaces/v1/genesis.proto
```diff
@@ -0,0 +1,46 @@
...
+
+  repeated UserGroup user_groups = 4 [ (gogoproto.nullable) = false ];
+}
+
+// ACL represents a single Access Control List entry
+message ACLEntry {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
+  // Either group name or user address
```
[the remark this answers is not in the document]
RiccardoM: It should have been referred to `target`. I should have fixed it now

x/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,382 @@
...
+		Use:                        types.ModuleName,
+		Short:                      "Subspaces transaction subcommands",
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	subspacesTxCmd.AddCommand(
+		GetCmdCreateSubspace(),
+		GetCmdEditSubspace(),
```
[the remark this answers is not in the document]
RiccardoM: Added now :heavy_check_mark:

app/app.go
```diff
@@ -524,6 +535,7 @@ func NewDesmosApp(
 	app.mm.SetOrderBeginBlockers(
 		upgradetypes.ModuleName, capabilitytypes.ModuleName, minttypes.ModuleName, distrtypes.ModuleName, slashingtypes.ModuleName,
 		evidencetypes.ModuleName, stakingtypes.ModuleName, ibchost.ModuleName,
+		// TODO: See if we need a BeginBlocker for the subspaces module
```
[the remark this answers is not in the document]
RiccardoM: Ok, I'll just remove this and we can add the begin blocker later when there's the need to do so (as there isn't now)

x/subspaces/client/cli/tx.go
bragaz: Yes I think `UserGroup`s deserves their own struct. We will be able to handle them better later on.

proto/desmos/subspaces/v1/subspace.proto
bragaz: Ok I see.

proto/desmos/subspaces/v1/msgs.proto
```diff
@@ -0,0 +1,187 @@
...
+  string group_name = 2 [ (gogoproto.moretags) = "yaml:\"group_name\"" ];
+  string signer = 3 [ (gogoproto.moretags) = "yaml:\"signer\"" ];
+}
+
+// MsgDeleteUserGroupResponse defines the Msg/DeleteUserGroup response type
+message MsgDeleteUserGroupResponse {}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// MsgSetUserGroupPermissions represents the message used to add a user to a
```
dadamu, line 136:
```suggestion
// MsgAddUserToUserGroup represents the message used to add a user to a
```

x/subspaces/client/cli/tx.go
dadamu: Agree with we should have `UserGroup` struct. But if we use group_id instead, it breaks the `SetPermissions` message and permission store key, the type of target is string now.

proto/desmos/subspaces/v1/genesis.proto
```diff
@@ -0,0 +1,47 @@
...
+  repeated desmos.subspaces.v1.Subspace subspaces = 2
+      [ (gogoproto.nullable) = false ];
+
+  repeated ACLEntry acl = 3
+      [ (gogoproto.customname) = "ACL", (gogoproto.nullable) = false ];
+
+  repeated UserGroup user_groups = 4 [ (gogoproto.nullable) = false ];
+}
+
+// ACL represents a single Access Control List entry
```
dadamu, line 26:
```suggestion
// ACLEntry represents a single Access Control List entry
```

x/subspaces/client/cli/tx.go
RiccardoM: All right, will implement the `UserGroup` type. I think we can have the group permissions as a field of the `UserGroup`, and then different messages for users/groups separately (eg. `MsgSetUserPermission` and `MsgSetGroupPermission`). What do you think @dadamu?

x/profiles/legacy/v300/store.go
```diff
@@ -0,0 +1,112 @@
...
+
+		// Set the key inside the store
+		store.Set(keys[i], blockBz)
+	}
+
+	return nil
+}
+
+// migrateRelationships migrates the relationships stored to the new type, converting the subspace from string to uint64
+func migrateRelationships(store sdk.KVStore, cdc codec.BinaryCodec) error {
```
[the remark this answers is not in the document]
RiccardoM: Added the tests :heavy_check_mark:

x/subspaces/client/cli/tx.go
dadamu: @RiccardoM It is good to me.

proto/desmos/subspaces/v1/genesis.proto
```diff
@@ -0,0 +1,47 @@
...
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
+  uint64 initial_subspace_id = 1
+      [ (gogoproto.customname) = "InitialSubspaceID" ];
+
+  repeated desmos.subspaces.v1.Subspace subspaces = 2
+      [ (gogoproto.nullable) = false ];
+
+  repeated ACLEntry acl = 3
```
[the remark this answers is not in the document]
RiccardoM: I've used the ACL term since it's the acronym for Access Control List. I think adding _entries_ (and making it _Access Control List entries_) would be redundant

RiccardoM:
Groups have been implemented

x/subspaces/keeper/msg_server.go
```diff
@@ -0,0 +1,499 @@
...
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "group %d could not be found", msg.GroupID)
+	}
+
+	signer, err := sdk.AccAddressFromBech32(msg.Signer)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid signer address: %s", msg.Signer)
+	}
+
+	// Check the permissions
+	if !k.HasPermission(ctx, msg.SubspaceID, signer, types.PermissionSetPermissions) {
```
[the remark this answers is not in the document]
RiccardoM: Actually, the `PermissionManageGroups` is thought to allow users to edit the details of groups (the name, description, etc) and not the members or the associated permissions. Since adding/removing a user to a group is the same thing as setting their permissions individually, I think it's better to require the `PermissionManageGroups` here instead.

x/subspaces/client/cli/query.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	flags.AddQueryFlagsToCmd(cmd)
+	flags.AddPaginationFlagsToCmd(cmd, "user groups")
+
+	return cmd
+}
+
+// GetCmdQueryUserGroupMembers returns the command to query the members of a specific user group
+func GetCmdQueryUserGroupMembers() *cobra.Command {
```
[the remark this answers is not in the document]
RiccardoM: Done!

x/subspaces/keeper/msg_server.go
```diff
@@ -0,0 +1,499 @@
...
```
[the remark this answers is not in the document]
RiccardoM: Same as above

x/subspaces/client/cli/cli_test.go
```diff
@@ -0,0 +1,760 @@
...
+func (s *IntegrationTestSuite) TestCmdQueryUserGroups() {
+	val := s.network.Validators[0]
+	testCases := []struct {
+		name        string
+		args        []string
+		shouldErr   bool
+		expResponse types.QueryUserGroupsResponse
+	}{
+		{
+			name: "subspaces are returned correctly",
```
bragaz, line 243:
```suggestion
			name: "users groups are returned correctly",
```

x/subspaces/keeper/msg_server.go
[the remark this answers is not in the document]
bragaz: 🤔 Did you mean `PermissionSetPermissions`?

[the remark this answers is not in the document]
RiccardoM: Yes, thanks. Fixed

bragaz (approved):
Looks ready to me. 💯

x/subspaces/keeper/hooks.go
```diff
@@ -0,0 +1,68 @@
...
+
+// DONTCOVER
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/desmos-labs/desmos/v2/x/subspaces/types"
+)
+
+// Implements StakingHooks interface
```
dadamu, line 11:
```suggestion
// Implements SubspacesHooks interface
```

x/subspaces/keeper/grpc_query.go
```diff
@@ -0,0 +1,154 @@
...
+	if !k.HasSubspace(sdkCtx, request.SubspaceId) {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace with id %d not found", request.SubspaceId)
+	}
+
+	sdkAddr, err := sdk.AccAddressFromBech32(request.User)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid user address: %s", request.User)
+	}
+
+	permission := k.GetUserPermissions(sdkCtx, request.SubspaceId, sdkAddr)
```
dadamu, line 152: The user might be in a group, should we show group permissions of the user?

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // UserGroupMembers queries all the members of a given user group
+  rpc UserGroupMembers(QueryUserGroupMembersRequest)
+      returns (QueryUserGroupMembersResponse) {
+    option (google.api.http).get =
+        "/desmos/subspaces/v1/groups/{subspace_id}/{group_id}/members";
+  }
+
+  // UserPermissions queries the permissions for the given user
```
dadamu, line 45: `UserPermissions` query is for the specific user. I think Having `GroupPermissions` query would be helpful.

dadamu: Never mind I found it have been included in the `UserGroup` method already.

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,150 @@
...
+  // Subspace queries all the information about the subspace with the given id
+  rpc Subspace(QuerySubspaceRequest) returns (QuerySubspaceResponse) {
+    option (google.api.http).get =
+        "/desmos/subspaces/v1/subspaces/{subspace_id}";
+  }
+
+  // UserGroups queries all the groups that are present inside the subspace with
+  // the given id
+  rpc UserGroups(QueryUserGroupsRequest) returns (QueryUserGroupsResponse) {
+    option (google.api.http).get = "/desmos/subspaces/v1/groups/{subspace_id}";
```
dadamu, line 28: As a rest api, should it follow the naming guide to be "/desmos/subspaces/v1/subspaces/{subspace_id}/groups"?

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // UserGroup queries the user group having the given id inside the specific
+  // subspace
+  rpc UserGroup(QueryUserGroupRequest) returns (QueryUserGroupResponse) {
+    option (google.api.http).get =
+        "/desmos/subspaces/v1/groups/{subspace_id}/{group_id}";
```
dadamu, line 35: As above

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // UserGroupMembers queries all the members of a given user group
+  rpc UserGroupMembers(QueryUserGroupMembersRequest)
+      returns (QueryUserGroupMembersResponse) {
+    option (google.api.http).get =
+        "/desmos/subspaces/v1/groups/{subspace_id}/{group_id}/members";
```
dadamu, line 42: As above

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // UserPermissions queries the permissions for the given user
+  rpc UserPermissions(QueryUserPermissionsRequest)
+      returns (QueryUserPermissionsResponse) {
+    option (google.api.http).get =
+        "/desmos/subspaces/v1/permissions/{subspace_id}/{user}";
```
dadamu, line 49: As above

x/subspaces/client/cli/query.go
```diff
@@ -0,0 +1,214 @@
...
+	flags.AddPaginationFlagsToCmd(cmd, "subspaces")
+
+	return cmd
+}
+
+// -------------------------------------------------------------------------------------------------------------------
+
+// GetGroupsQueryCmd returns a new command to perform queries for user groups
+func GetGroupsQueryCmd() *cobra.Command {
+	groupsTxCmd := &cobra.Command{
```
dadamu, line 107: I would name it groupsQueryCmd.

x/subspaces/keeper/grpc_query.go
RiccardoM: You're right, I've updated the response type to also include those permissions as well as the details of each one

x/subspaces/keeper/alias_functions.go
```diff
@@ -0,0 +1,145 @@
...
+		stop := fn(i, member)
+		if stop {
+			break
+		}
+		i++
+	}
+}
+
+// GetGroupMembers iterates returns all the members of a group inside a specific subspace
+func (k Keeper) GetGroupMembers(ctx sdk.Context, subspaceID uint64, groupID uint32) []sdk.AccAddress {
```
dadamu, line 113: It is only used in the situation test. Should we move it to the test code?

x/subspaces/keeper/groups.go
```diff
@@ -0,0 +1,113 @@
...
+
+// IsMemberOfGroup returns whether the given user is part of the group with
+// the specified id inside the provided subspace
+func (k Keeper) IsMemberOfGroup(ctx sdk.Context, subspaceID uint64, groupID uint32, user sdk.AccAddress) bool {
+	store := ctx.KVStore(k.storeKey)
+	return store.Has(types.GroupMemberStoreKey(subspaceID, groupID, user))
+}
+
+// RemoveUserFromGroup removes the specified user from the subspace group having the given id.
+// If the group does not exist inside the subspace, it returns an error.
```
dadamu, line 107: It does not return anything.

x/subspaces/keeper/groups.go
```diff
@@ -0,0 +1,113 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/desmos-labs/desmos/v2/x/subspaces/types"
+)
+
+// SetGroupID sets the new group id for the specific subspace to the store
```
dadamu, line 10: I would name it to `SetNextGroupID` since `SetGroupID` makes me think it the current group id.
And should we make it private method to prevent other modules change it?

x/subspaces/keeper/groups.go
RiccardoM: The documentation is outdated, it's better to not return anything. Let me fix this

```suggestion
```

x/subspaces/keeper/groups.go
RiccardoM: `SetGroupID` is used to match `GetGroupID`, which returns the group id to be used when creating a group id. I think the documentation is clear enough, and it's better to match the name of the getter instead (like it is right now). 

I think it's fine keeping it public since we need it inside tests as well.

x/profiles/simulation/genesis.go
```diff
@@ -38,10 +41,16 @@ func RandomizedGenState(simsState *module.SimulationState) {
 	simsState.GenState[authtypes.ModuleName] = bz
 
 	// Create and set profiles state
```
dadamu, line 43:
```suggestion
```

x/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,570 @@
...
+
+	cmd.Flags().String(FlagDescription, "", "Description of the group")
+	cmd.Flags().StringSlice(FlagPermissions, []string{types.SerializePermission(types.PermissionNothing)}, "Permissions of the group")
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// GetCmdEditUserGroup returns the command to edit a user group
+func GetCmdEditUserGroup() *cobra.Command {
```
dadamu, line 292: Can you add the test for it?

x/subspaces/client/cli/tx.go
```diff
@@ -0,0 +1,570 @@
...
+	cmd.Flags().String(FlagName, types.DoNotModify, "New human readable name of the group")
+	cmd.Flags().String(FlagDescription, types.DoNotModify, "Description of the group")
+
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// GetCmdSetUserGroupPermissions returns the command to set the permissions for a user group
+func GetCmdSetUserGroupPermissions() *cobra.Command {
```
dadamu, line 347: As above, there is no test for it.

proto/desmos/subspaces/v1/query.proto
```diff
@@ -0,0 +1,153 @@
...
+
+  uint64 subspace_id = 1 [ (gogoproto.moretags) = "yaml:\"subspace_id\"" ];
+  string user = 2 [ (gogoproto.moretags) = "yaml:\"user\"" ];
+}
+
+// QueryUserPermissionsRequest is the response type for the
+// Query/UserPermissions method
+message QueryUserPermissionsResponse {
+  uint32 permissions = 1 [ (gogoproto.moretags) = "yaml:\"permissions\"" ];
+  repeated PermissionDetail details = 2
```
dadamu, line 151: Should we add `nullable = null` to prevent it returns the pointer?

x/subspaces/types/models.go
```diff
@@ -0,0 +1,229 @@
...
+		update.Name,
+		update.Description,
+		group.Permissions,
+	)
+}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// NewPermissionDetailUser returns a new PermissionDetail for the user with the given address and permission value
+func NewPermissionDetailUser(user string, permission Permission) *PermissionDetail {
```
dadamu, line 208: I think it is better to return a structure instead of a pointer.

dadamu (changes requested):
Overrall is good to me. Just have minor questions as above, and the sim test is failed.

x/subspaces/client/cli/tx.go
RiccardoM: Added

x/subspaces/client/cli/tx.go
RiccardoM: Added

dadamu (approved):
Ready to go🚀

PR #750, opened 2022-02-22
feat: relationships module

RiccardoM:
## Description
This PR implements the new `x/relationships` module. This module contains all the user blocks and relationships actions that were previously present inside the `x/profiles` module. 

Some changes have been done to the logic of relationships and user blocks, particularly: 
- it is **not** required anymore to have a Desmos profile in order to block another user 
- it is **not** required anymore that both parties have a Desmos profile in order to establish a relationship 
- blocking a user inside the subspace `0` will **not** lead to the user being blocked on all subspaces anymore. The subspace with id `0` now only identifies the `x/profiles` subspace. Users blocked there won't be able to send DTag transfer requests to other users or perform other actions from within the `x/profiles` module that involve the blocker. 

Depends-On: #753

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/desmos/relationships/v1/genesis.proto
```diff
@@ -0,0 +1,22 @@
...
+// GenesisState defines the profiles module's genesis state.
+message GenesisState {
+  option (gogoproto.goproto_getters) = false;
+
+  repeated desmos.relationships.v1.Relationship relationships = 2 [
+    (gogoproto.moretags) = "yaml:\"relationships\"",
+    (gogoproto.nullable) = false
+  ];
+
+  repeated desmos.relationships.v1.UserBlock blocks = 3 [
```
bragaz, line 18:
```suggestion
  repeated desmos.relationships.v1.UserBlock blocks = 2 [
```

x/profiles/keeper/keeper.go
```diff
@@ -152,11 +158,7 @@ func (k Keeper) RemoveProfile(ctx sdk.Context, address string) error {
 	store := ctx.KVStore(k.storeKey)
 	store.Delete(types.DTagStoreKey(profile.DTag))
 
-	// Delete all the blocks
-	k.DeleteAllUserBlocks(ctx, address)
-
-	// Delete all the relationships
-	k.DeleteAllUserRelationships(ctx, address)
+	// TODO: Add here OnProfileDeleted to delete all the user blocks and relationships
```
bragaz, line 161: I think we should implement this as well

x/relationships/handler.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/gogo/protobuf/proto"
+
+	"github.com/desmos-labs/desmos/v2/x/relationships/keeper"
+	"github.com/desmos-labs/desmos/v2/x/relationships/types"
+)
+
+// NewHandler returns a handler for "profile" type messages.
```
bragaz, line 12:
```suggestion
// NewHandler returns a handler for "relationships" type messages.
```

x/relationships/keeper/msg_server.go
```diff
@@ -0,0 +1,165 @@
...
+	// Check if the subspace exists
+	if !k.DoesSubspaceExist(ctx, msg.SubspaceID) {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace with id %d not found", msg.SubspaceID)
+	}
+
+	// Check if the receiver has blocked the sender before
+	if k.HasUserBlocked(ctx, msg.Blocker, msg.Blocked, msg.SubspaceID) {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "%s has already blocked %s", msg.Blocker, msg.Blocked)
+	}
+
```
bragaz, line 112: Shouldn't we also check if the two users has a relationship and, in that case, delete that as well?

proto/desmos/profiles/v2/models_relationships.proto
```diff
@@ -16,7 +16,10 @@ message Relationship {
 
   string creator = 1 [ (gogoproto.moretags) = "yaml:\"creator\"" ];
   string recipient = 2 [ (gogoproto.moretags) = "yaml:\"recipient\"" ];
-  string subspace = 3 [ (gogoproto.moretags) = "yaml:\"subspace\"" ];
+  string subspace_id = 3 [
+    (gogoproto.customname) = "SubspaceID",
+    (gogoproto.moretags) = "yaml:\"subspace\""
```
dadamu, line 21:
```suggestion
    (gogoproto.moretags) = "yaml:\"subspace_id\""
```

proto/desmos/profiles/v2/models_relationships.proto
```diff
@@ -34,7 +37,10 @@ message UserBlock {
...
   string reason = 3 [ (gogoproto.moretags) = "yaml:\"reason\"" ];
 
-  // Subspace contains the ID of the subspace inside which the user should be
-  // blocked
-  string subspace = 4 [ (gogoproto.moretags) = "yaml:\"subspace\"" ];
+  // SubspaceID represents the ID of the subspace inside which the user should
+  // be blocked
+  string subspace_id = 4 [
+    (gogoproto.customname) = "SubspaceID",
+    (gogoproto.moretags) = "yaml:\"subspace\""
```
dadamu, line 44:
```suggestion
    (gogoproto.moretags) = "yaml:\"subspace_id\""
```

proto/desmos/relationships/v1/models.proto
```diff
@@ -1,22 +1,27 @@
...
 message Relationship {
   option (gogoproto.equal) = true;
   option (gogoproto.goproto_stringer) = true;
 
+  // Creator represents the creator of the relationship
   string creator = 1 [ (gogoproto.moretags) = "yaml:\"creator\"" ];
-  string recipient = 2 [ (gogoproto.moretags) = "yaml:\"recipient\"" ];
+
+  // Counterparty represents the other user involved in the relationship
+  string counterparty = 2 [ (gogoproto.moretags) = "yaml:\"recipient\"" ];
```
dadamu, line 21:
```suggestion
  string counterparty = 2 [ (gogoproto.moretags) = "yaml:\"counterparty \"" ];
```

proto/desmos/relationships/v1/models.proto
```diff
@@ -1,22 +1,27 @@
...
+
+  // SubspaceID represents the id of the subspace for which the relationship is
+  // valid
   uint64 subspace_id = 3 [
     (gogoproto.customname) = "SubspaceID",
     (gogoproto.moretags) = "yaml:\"subspace\""
```
dadamu, line 27:
```suggestion
    (gogoproto.moretags) = "yaml:\"subspace_id \""
```

x/relationships/keeper/msg_server.go
RiccardoM: I think it's best to leave DApp handle this case where a relationship and a block exist between two users. One DApp might want to preserve the relationships in the case the block gets removed after. Once other might want to deleted it entirely. I think it's best to leave them there

x/profiles/keeper/keeper.go
RiccardoM: Since now relationships and user blocks do **not** require users to have a profile, I think it can be safely removed. I've instead implemented another thing: when a subspace gets deleted, the relationships and user blocks inside that subspace will be deleted too to make sure we have no garbage inside our store. This should be a better implementation overall rather than relying on profiles (which can be quickly deleted and recreated).

proto/desmos/profiles/v2/models_relationships.proto
RiccardoM: This comment (and all the ones below) actually refer to the Proto definitions for either `v1` or `v2`, which I think should be kept as they are rather than changed (even thought the change to the YAML tag does not have any effect since data is stored as bytes).

bragaz (approved):

bragaz (commented):
The overall looks good but there are some tests failing because of the hooks.

app/sim_test.go
```diff
@@ -231,7 +229,6 @@ func TestAppImportExport(t *testing.T) {
 
 		{app.keys[subspacestypes.StoreKey], newApp.keys[subspacestypes.StoreKey], [][]byte{}},
 		{app.keys[profilestypes.StoreKey], newApp.keys[profilestypes.StoreKey], [][]byte{}},
-		{app.keys[relationshipstypes.StoreKey], newApp.keys[relationshipstypes.StoreKey], [][]byte{}},
```
bragaz: why we remove this?

dadamu (approved):
Overall is good to me. It is ready to go once the problem @bragaz proposed is solved.

PR #764, opened 2022-03-02
docs(adr): ADR-008 Fees module

bragaz:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/docs/architecture/adr-008-fee-module.md
+++ b/docs/architecture/adr-008-fee-module.md
@@ -1,44 +1,45 @@
-# ADR 008: Fee modules
+# ADR 008: Fees modules
 
 ## Changelog
 
-- March 2, 2022: Initial draft.
+- March 2, 2022: Initial draft;
+- March 8, 2022: First review.
 
 ## Status
 DRAFTED
 
 ## Abstract
-This ADR defines the `x/fees` module which enables the possibility to set custom fees for each desmos message
-through governance proposals.
+This ADR defines the `x/fees` module which allows setting custom min fees for each message type.
 
 ## Context
-In order to better prevent any kind of spam, from zero-gas attack, to smart contracts implementation, to subspaces creation and, 
-in the near future, through posts, it's useful to have a system with which the community can tweak the costs of desmos messages.
+In order to better prevent any kind of spam (e.g. zero-gas attacks, garbage smart contracts implementations, etc.) it's useful 
+to have a system that allows to set a minimum amount of fees that needs to be paid when sending specific messages that can be vectors of such spam attacks. 
+This system should allow changing dynamically such min fees amounts, so that the community can properly tweak them if necessary.
 
 ## Decision
-We will create a module named `fees` which give desmos community the possibility to add minimum fees to any Desmos custom
-message through governance.
+We will create a module named `fees` that allows setting such min fee amounts of any kind of messages that our chain supports. 
+This will then make sure that when such messages are broadcast inside a transaction, the transaction signer 
+is paying at least the minimum amount of fees for each message.
 
 ### Types
-Minimum (or custom) fees will contain the following elements:  
 
+Minimum (or custom) fees will contain the following elements:
 * a string identifying the message type delivered with the transaction;
 * the amount of fees associated with the give message.
 
 #### Min Fees
 ```go
 type MinFee struct {
-	// The message type identifying the Desmos message
+	// The message type for which this min fee amount is valid
 	messageType string
 	
-	// The amount of fees to be paid for the message
+	// The min amount of fees to be paid for each instance of this type of message
 	amount sdk.Coins
 }
 ```
 
 #### Params
-We will save the `MinFee`s for messages in the module's `Params` in order to be able to change them with governance 
-`ParameterChangeProposal`s. To pursue efficiency, the ideal implementation for them is the following:
+We will save each `MinFee` instance as the module on-chain params in order to be able to later change them with governance proposals as follows:
 
 ```go
 type Params struct {
@@ -59,35 +60,29 @@ type MinFeeDecorator struct {
 }
 ```
 
-This decorator will implement the following interface:
-```go
-// AnteDecorator wraps the next AnteHandler to perform custom pre- and post-processing.
-type AnteDecorator interface {
-	AnteHandle(ctx Context, tx Tx, simulate bool, next AnteHandler) (newCtx Context, err error)
-}
-```
+The custom `AnteHandler` will iterate over all the transaction's messages and perform the following operations:
+- Fetch all the min fees to be paid for such message from the `x/fees` params (if no min fees are set, then `minFee = 0`);
+- Sum all the min fees together;
 
-### `Query` Service
-The module will expose the following query:
-```protobuf
-// Query defines the gRPC querier service.
-service Query {
-  // Params queries the fees module params
-  rpc Params(QueryParamsRequest) returns (QueryParamsResponse);
-}
-```
+After having calculated the total required min fees, it will check that fees are greater or equals to the min fees, 
+and, if not, return an error.
 
 ## Consequences
 
+### Positive
+- Spam prevention of any kind of transaction that contains messages with custom fees;
+- Avoid uncontrolled smart contracts deployment.
+
 ### Negative
-Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`, 
+- Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`, 
 which will need to perform stateful checks that can eventually slow down the node a bit. 
 
 ## Test Cases
 
 We will need to add the following test cases :
-* The custom fees are applied correctly;
-* Benchmark on `AnteHandler`.
+- a transaction not having enough fees is rejected;
+- a transaction having enough fees is accepted;
+- `AnteHandler` benchmark tests to make sure it does not impact the transaction handling process too much.
 
 ## References
 
--- a/docs/architecture/adr-008-fee-module.md
+++ b/docs/architecture/adr-008-fee-module.md
@@ -66,7 +66,7 @@ and, if not, return an error.
 ## Consequences
 
 ### Positive
-- Spam prevention of any kind of transaction that contains messages with custom fees;
+- Spam prevention of transaction containing specific messages
 - Avoid uncontrolled smart contracts deployment.
 
 ### Negative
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+	messageType string
+	
+	// The amount of fees to be paid for the message
+	amount sdk.Coins
+}
+```
+
+#### Params
+We will save the `MinFee`s for messages in the module's `Params` in order to be able to change them with governance 
+`ParameterChangeProposal`s. To pursue efficiency, the ideal implementation for them is the following:
```
RiccardoM, line 41:
```suggestion
We will save each `MinFee` instance as the module on-chain params in order to be able to later change them with governance proposals as follows:
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+
+```go
+type Params struct {
+	MinFees map[string]MinFee
```
RiccardoM, line 45: Considering that
- there won't be too many different messages types supported
- maps are non deterministic inside Go
- the `MinFee` type already contains the message type

We can simply have an array here instead 
```suggestion
	MinFees []MinFee
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+This map will use the proto message type as key, allowing us to keep a constant complexity lookup for its elements.
```
RiccardoM, line 49: If the above comment is accepted, this line needs to be removed

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+This decorator will implement the following interface:
+```go
+// AnteDecorator wraps the next AnteHandler to perform custom pre- and post-processing.
+type AnteDecorator interface {
+	AnteHandle(ctx Context, tx Tx, simulate bool, next AnteHandler) (newCtx Context, err error)
+}
+```
```
RiccardoM, line 68: Instead of telling what interface it will implement (which is not needed since all `AnteHandler` need to implement it anyway), it's better to instead tell what kind of operations this ante handler need to perform: 

> iterate over all the messages inside a transaction and, for each message, do the following: 
> 1. retrieve the min fees to be paid for such message from the `x/fees` params (if no min fees are set, consider `minFee = 0 `). 
> 2. sum all the min fees together 
> 3. check if the transaction fees are greater or equals to the min fees, if not return an error

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+
+### `Query` Service
+The module will expose the following query:
+```protobuf
+// Query defines the gRPC querier service.
+service Query {
+  // Params queries the fees module params
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse);
+}
+```
```
RiccardoM, line 78: No need to put this since it's not that important

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Consequences
+
```
RiccardoM, line 81: The positives consequences are missing here (spam prevention, etc)

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+### Negative
+Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`,
```
RiccardoM, line 83:
```suggestion
- Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`, 
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+## Consequences
+
+### Negative
+Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`, 
+which will need to perform stateful checks that can eventually slow down the node a bit. 
+
+## Test Cases
+
+We will need to add the following test cases :
+* The custom fees are applied correctly;
```
RiccardoM, line 89:
```suggestion
- a transaction not having enough fees is rejected;
- a transaction having enough fees is accepted;
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,95 @@
...
+* Benchmark on `AnteHandler`.
```
RiccardoM, line 90:
```suggestion
- `AnteHandler` benchmark tests to make sure it does not impact the transaction handling process too much
```

docs/architecture/adr-008-fee-module.md
bragaz: I choose this implementation after reading this article:
https://boltandnuts.wordpress.com/2017/11/20/go-slice-vs-maps/
The test-2 shows how it's always better to use map for strings lookup. 
It would make the whole code faster, we will use only one loop I guess and, considering this is executing on the `AnteHandler`, it's better to keep it as fast as possible IMO.

dadamu (approved):
Everything looks good to me.

docs/architecture/adr-008-fee-module.md
RiccardoM: @bragaz Ok, that makes sense. But then I would not use `MinFee` as the value type and instead use `sdk.Coins` directly. Otherwise we would have the message URL that is being stored twice each block (as the state is duplicated on each block): both inside the map key and inside the `MinFee` field. So I suggest using something like this 
```go
MinFees map[string]sdk.Coins
```

Where the key represents the Proto URL of the messages, and the `sdk.Coins` represent the min fees for that specific message.

bragaz: @RiccardoM I agree, but I still need to use the `MinFee` type like this:
```protobuf
message MinFee {
  repeated cosmos.base.v1beta1.Coin amount = 2 [
    (gogoproto.nullable) = false,
    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
  ];
}
```
It's not possible to use repeated types as map's value in protobuf, so I need to wrap it before.

RiccardoM: Hmm that's a bummer. I really don't like this approach to be honest since it creates overhead just to store something only because we want to use maps rather than lists (there wouldn't be this overhead if we used a lists since the message type would be inside the `MinFee` object itself). 

Let's just step back for a second and consider the usage maps vs lists. 

As you said, this is going to run inside the `AnteHandler`. Currently Desmos supports a blocks size that allows to store up to 50 messages per blocks (this is the maximum amount we've seen during the spam attack). Now, let's consider a 20x worse case: 1.000 messages per block. 

Let's also consider a case in which we set a minimum fee for 100 different types of messages (even though this number will be most likely very smaller). 

Considering [the benchmarks you posted](https://boltandnuts.wordpress.com/2017/11/20/go-slice-vs-maps/), and assuming we are very unlucky and constantly looking for the last element, we would then obtain the following performances: 
- using an array:  
  `128 ns/op * 1.000 = 0.128 ms`
- using a map:  
   `25.7 ns/op * 1.000 = 0.0257 ms`

Even thought the difference is quite significant (using a map allows a 79.9% improvement), the absolute amount of time required to perform such check is extremely small. In both cases we are well under 1 ms. 

Honestly I don't think using a list would cause any problem. On the other hand, using an array would allow us to get the following benefits: 
- a cleaner code base (since we just define a list of objects and put everything inside each object) 
- a more client-friendly setup (since clients would not need to iterate over the key-value pairs to get the data but they would simply have to read the list contents)

Maps are more efficient, I recognize that, but I don't think in our case this would change a lot even in the worst case scenario.

bragaz: Ok that's enough for me! Let's use `MinFee` directly then!

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,90 @@
...
+    amount sdk.Coins
+}
+```
+
+#### Params
+We will save each `MinFee` instance as the module on-chain params in order to be able to later change them with governance proposals as follows:
+
+```go
+type Params struct {
+	MinFees MinFee
```
RiccardoM, line 48:
```suggestion
	MinFees []MinFee
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,90 @@
...
+## Decision
+We will create a module named `fees` that allows setting such min fee amounts of any kind of messages that our chain supports. 
+This will then make sure that when such messages are broadcast inside a transaction, the transaction signer 
+is paying at least the minimum amount of fees for each message.
+
+### Types
+
+Minimum (or custom) fees will contain the following elements:
+* a string identifying the message type delivered with the transaction;
+* the amount of fees associated with the give message.
```
RiccardoM, line 30:
```suggestion
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,86 @@
...
+- Fetch all the min fees to be paid for such message from the `x/fees` params (if no min fees are set, then `minFee = 0`);
+- Sum all the min fees together;
+
+After having calculated the total required min fees, it will check that fees are greater or equals to the min fees, 
+and, if not, return an error.
+
+## Consequences
+
+### Positive
+- Spam prevention of any kind of transaction that contains messages with custom fees;
```
RiccardoM, line 69:
```suggestion
- Spam prevention of transaction containing specific messages
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,86 @@
...
+- Avoid uncontrolled smart contracts deployment.
```
RiccardoM, line 70: Can be removed as it's implicit in the first one
```suggestion
```

docs/architecture/adr-008-fee-module.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Negative
+- Applying custom fees to messages requires to add an extra decorator to the `AnteHandler`, 
+which will need to perform stateful checks that can eventually slow down the node a bit. 
+
+## Test Cases
+
+We will need to add the following test cases :
```
RiccardoM, line 78:
```suggestion
We will need to add the following test cases:
```

RiccardoM (approved):

Issue #777, opened 2022-03-15
Reverse search through app links


manu0466:
## Feature description 
Implement the possibility to search a Desmos profile using an app-link.
For example let's assume that Alice have linked her Twitter account (AliceInWonderland) to her Desmos profile.
This feature should allow to find her Desmos profile using her Twitter handle AliceInWonderland.

PR #793, opened 2022-03-23
feat: reverse search links

RiccardoM:
## Description
This PR adds the ability to reverse search for app links and chain links owners. The new following queries have been added: 
- search for app link owners using an application name (optional), and username (optional, works only if the application is specified as well)
- search for chain link owners using a chain name (optional), and target address (optional, works only if the chain name is specified as well)  

Closes: #777
Closes: #794

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/profiles/legacy/v5/store.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/desmos-labs/desmos/v3/x/profiles/types"
+)
+
+// MigrateStore performs in-place store migrations from v5 to v6
+// The migration includes:
+//
+// - add missing application links owner keys
+// - add missing chain links owner keys
+//
+func MigrateStore(ctx sdk.Context, storeKey sdk.StoreKey, cdc codec.BinaryCodec) error {
```
dadamu, line 17: Can you add test for it?

x/profiles/keeper/grpc_query.go
```diff
@@ -122,6 +123,43 @@ func (k Keeper) ChainLinks(ctx context.Context, request *types.QueryChainLinksRe
 	return &types.QueryChainLinksResponse{Links: links, Pagination: pageRes}, nil
 }
 
+// ChainLinkOwners implements the Query/ChainLinkOwners gRPC method
+func (k Keeper) ChainLinkOwners(ctx context.Context, request *types.QueryChainLinkOwnersRequest) (*types.QueryChainLinkOwnersResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	store := sdkCtx.KVStore(k.storeKey)
+
+	ownersPrefix := types.ChainLinkChainPrefix
+	switch {
```
dadamu, line 132: If should have a default case to return error to avoid the case whose fields are empty.

x/profiles/keeper/grpc_query.go
```diff
@@ -174,6 +212,43 @@ func (k Keeper) ApplicationLinkByClientID(ctx context.Context, request *types.Qu
 	return &types.QueryApplicationLinkByClientIDResponse{Link: link}, nil
 }
 
+// ApplicationLinkOwners implements the Query/ApplicationLinkOwners gRPC method
+func (k Keeper) ApplicationLinkOwners(ctx context.Context, request *types.QueryApplicationLinkOwnersRequest) (*types.QueryApplicationLinkOwnersResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	store := sdkCtx.KVStore(k.storeKey)
+
+	ownersPrefix := types.ApplicationLinkAppPrefix
+	switch {
```
dadamu, line 221: As above

x/profiles/keeper/grpc_query.go
RiccardoM: The default case is the one before the switch, where the entire `ChainLinkChainPrefix` it's used. I've made it like that because this way it's possible to iterate over all the chain link prefixes. This shouldn't be a problem because there is the pagination anyway.

x/profiles/legacy/v5/store.go
RiccardoM: Added, thanks!

bragaz (approved):

x/profiles/keeper/grpc_query.go
```diff
@@ -122,6 +123,43 @@ func (k Keeper) ChainLinks(ctx context.Context, request *types.QueryChainLinksRe
...
+		ownersPrefix = types.ChainLinkChainKey(request.ChainName)
+	}
+
+	var owners []types.QueryChainLinkOwnersResponse_ChainLinkOwnerDetails
+	ownersStore := prefix.NewStore(store, ownersPrefix)
+	pageRes, err := query.Paginate(ownersStore, request.Pagination, func(key []byte, value []byte) error {
+		keyWithPrefix := append(ownersPrefix, key...)
+		cleanedKey := bytes.TrimSuffix(bytes.TrimPrefix(keyWithPrefix, types.ChainLinkChainPrefix), value)
+		values := bytes.Split(cleanedKey, types.Separator)
+		chainName, target := values[0], values[1]
```
dadamu, line 145: I think it can be all implemented by `bytes.Split` instead of `bytes.TrimSuffix`, say:
```go
values := bytes.Split(cleanedKey, types.Separator)
chainName, target, value := values[0], values[1], values[2]
```
Then, the value of the key can be simply `0x1` in order to reduce the storage on chain.

RiccardoM: Right now, links owners are stored using the following keys: 
```
ChainLinkChainPrefix + Chain name + Separator + External Address + User address -> User address
```

So, if we use the `bytes.Split` it would return the following values:
```
1. Chain name
2. External Address + User address
```

We would have to remove the `User address` suffix anyway to get the external address

dadamu: @RiccardoM I mean that we can store the key like:

```
ChainLinkChainPrefix + Chain name + Separator + External Address + Spearator + User address -> 0x01
```

So the storage doesn't need to store user address again.

RiccardoM: @dadamu You are completely right! Thanks for pointing that out, this way we can save up a lot of disk space. I've just implemented it this way (and also added some missing tests)

bragaz: Considering this answer, should we look after other places where we use `bytes.TrimSuffix` and `bytes.TrimPrefix` and see if we can improve them as well?

RiccardoM: @bragaz All other keys instances do not trim prefixes unless necessary (eg. inside user group members). All the keys have already been reviewed inside their own PRs before being merged, and I don't think it would be appropriate to do it here anyway. I've checked though and didn't find any use of `bytes.TrimPrefix` aside from this case and user groups members anyway

x/profiles/keeper/keeper_chain_links_test.go
```diff
@@ -355,6 +355,14 @@ func (suite *KeeperTestSuite) TestKeeper_DeleteChainLink() {
 					"cosmos",
 					ext.GetAddress().String(),
 				))
+
+				// Check the additional keys
+				store := suite.ctx.KVStore(suite.storeKey)
```
dadamu, line 360: We should have checking the additional keys in the store test function as well.

RiccardoM: Done

x/profiles/keeper/keeper_app_links_test.go
```diff
@@ -487,6 +487,15 @@ func (suite *KeeperTestSuite) Test_DeleteApplicationLink() {
 					"twitter",
 					"twitteruser",
 				))
+
+				// Check the additional keys
+				store := suite.ctx.KVStore(suite.storeKey)
```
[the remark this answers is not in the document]
RiccardoM: Done

dadamu (approved):
Looks good to me, ready to go.

PR #802, opened 2022-04-01
fix: do not allow setting permissions for self

RiccardoM:
## Description
This PR changes how `MsgSetUserPermissions` is validated and how `MsgSetUserGroupPermissions` is handled to make sure a user cannot set their own permissions or the permissions of a group they are part of. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/subspaces/keeper/msg_server.go
+++ b/x/subspaces/keeper/msg_server.go
@@ -256,7 +256,8 @@ func (k msgServer) SetUserGroupPermissions(goCtx context.Context, msg *types.Msg
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	// Check if the subspace exists
-	if !k.HasSubspace(ctx, msg.SubspaceID) {
+	subspace, found := k.GetSubspace(ctx, msg.SubspaceID)
+	if !found {
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace with id %d not found", msg.SubspaceID)
 	}
 
@@ -276,8 +277,8 @@ func (k msgServer) SetUserGroupPermissions(goCtx context.Context, msg *types.Msg
 		return nil, sdkerrors.Wrapf(types.ErrPermissionDenied, "you cannot manage permissions in this subspace")
 	}
 
-	// Make sure that the user is not part of the group they want to change the permissions for
-	if k.IsMemberOfGroup(ctx, msg.SubspaceID, msg.GroupID, signer) {
+	// Make sure that the user is not part of the group they want to change the permissions for, unless they are the owner
+	if subspace.Owner != msg.Signer && k.IsMemberOfGroup(ctx, msg.SubspaceID, msg.GroupID, signer) {
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "cannot set the permissions for a group you are part of")
 	}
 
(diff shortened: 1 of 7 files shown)


x/subspaces/keeper/msg_server.go
```diff
@@ -276,6 +276,11 @@ func (k msgServer) SetUserGroupPermissions(goCtx context.Context, msg *types.Msg
 		return nil, sdkerrors.Wrapf(types.ErrPermissionDenied, "you cannot manage permissions in this subspace")
 	}
 
+	// Make sure that the user is not part of the group they want to change the permissions for
```
manu0466, line 279: What if is the subspace owner that wants to change the permissions of the group and is also part of it?

manu0466 (changes requested):

x/subspaces/keeper/msg_server.go
RiccardoM: Thanks for pointing this out @manu0466. I've updated the code to always allow the owner to set any group permission

bragaz (approved):

manu0466:
@RiccardoM I think that there is another problem.
From spec a user can't give to other users the `PermissionSetPermissions`, this can be baypassed in this 2 cases: 

## Case 1
A user have the `PermissionSetPermissions` the he can just call `SetUserPermissions` and give to another user the `PermissionSetPermissions`

## Case 2
- User **UA** have the `PermissionManageGroups`, then he create a new group with the defaultPermissions = `PermissionSetPermissions`
- Then he can add himelf to the group inheriting the `PermissionSetPermissions`

## Case 3
- User **UA** sets default permissions of group **GA** to something that have  `PermissionSetPermissions`, **UA** is not meber of the group.
- Then let's assume that **UA** have also the permissions `PermissionManageGroups` then he can add other users to the group giving them the `PermissionSetPermission`  since all the members inherit the group permissions.

In all cases I don't see any checks that makes sure that only the subspace owner can give the `SetUserPermissions` in `SetUserPermissions`, `CreateUserGroup`, `EditUserGroup`.

RiccardoM:
@manu0466 Let me answer you case by case

> ## Case 1
> A user have the `PermissionSetPermissions` the he can just call `SetUserPermissions` and give to another user the `PermissionSetPermissions`

This is the indented behavior. A subspace owner, which is the only initial user that can set permissions, should be very careful about giving out `PermissionSetPermissions` because the following might happen: 
1. Alice gives `PermissionSetPermissions` to Bob
2. Bob created a new account, named Carl
3. Bob gives `PermissionSetPermissions` to another user 

Unfortunately there is nothing we can do here. We should respect the choice of the subspace owner in this case and just allow them to do whatever they want. They can always decide to be the only one to give out permissions if they prefer. We just give options: it's up to them to use this feature or not. 
 
> ## Case 2
> * User **UA** have the `PermissionManageGroups`, then he create a new group with the defaultPermissions = `PermissionSetPermissions`
> * Then he can add himelf to the group inheriting the `PermissionSetPermissions`

This is actually a bug. I think it's better to open an issue about this and make sure during the handling of `MsgCreateUserGroup` that if `DefaultPermissions` contains `PermissionSetPermissions`, then the group owner must also have `PermissionSetPermissions` to avoid any trouble. 

> ## Case 3
> * User **UA** sets default permissions of group **GA** to something that have  `PermissionSetPermissions`, **UA** is not meber of the group.
> * Then let's assume that **UA** have also the permissions `PermissionManageGroups` then he can add other users to the group giving them the `PermissionSetPermission`  since all the members inherit the group permissions.

Actually, in order to add a user to a group you are required to have `PermissionSetPermissions`, not `PermissionManageGroups`. The latter is only checked when creating a group or editing an existing group. Also, editing user groups does **not** allow to set a new permission for the group itself. Groups permissions are only handled though `MsgSetUserGroupPermissions` which require `PermissionSetPermissions`.

manu0466 (approved):

PR #821, opened 2022-04-12
docs(adr): ADR-011: Reports module

RiccardoM:
## Description

This PR includes the ADR of the `x/reports` module that allows users to report misbehaving users or bad content inside subspaces.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


docs/architecture/adr-011-reports-module.md
```diff
@@ -0,0 +1,302 @@
...
+  
+  // Message attached to this report
+  optional string message = 5;
+  
+  // Address of the reporter 
+  required string reporter = 6;
+}
+
+// UserData contains the data of a report about a user
+message UserData {
```
bragaz, line 54: What's the reason behind using a wrapper for the `user` and the `post_id`?
I can only thing about extending the data contained here purposes but honestly I don't know
what else can be added here.

RiccardoM: It's both for extensibility and readability. 

Another way of doing this would have been to have: 
- an enum that contains all the possible values 
- a generic `data` field that can be used to specify things

However, the `data` would have needed to be something like `[]byte` to support integers (post ids) and strings (user addresses). Then we would have needed to read those in different ways based on the type of the report. 

Honestly I find it a lot more easy to manage and read having wrapping classes: it makes it very easy to integrate without requiring much documentation and special cases.

RiccardoM: You're right, they didn't make sense. I have now reviewed the `UserData` and `PostData` to also contain the id of the subspace inside which the user should be reported or the reported post is

bragaz: @RiccardoM Great! This made me think about another kind of report that we might want to consider, the `SubspaceReport`. Does it make sense to you? I'm struggling cause from one side, it could be useful to let people and clients know which are the those dangerous subspaces (like those promoting hatread, violence, racism) and from another one it could lead to what in the cinema and video games industries is called "review-bombing" (in our case **"reports-bombing"**) having users reporting a subspace just because it doesn't match their ideas or beliefs.

RiccardoM: @bragaz Honestly I don't think it's worth implementing it. We should allow for max freedom when creating subspaces, just like how the internet is. Also, we won't have any way of acting on top of subspaces anyway: we can't delete them so even if users report them what can we do?

bragaz: Nothing, true.

bragaz (approved):

docs/architecture/adr-011-reports-module.md
```diff
@@ -0,0 +1,307 @@
...
+  required uint32 reason_id = 1;
+}
+
+```
+
+### `Query` Service
+```protobuf
+service Query {
+  // Reports allows to query the reports for a specific target
+  rpc Reports(QueryReportsRequest) returns (QueryReportsResponse) {
```
dadamu, line 226: It would be useful to have `ReportTallyResult` to show the tally result. Moderator group can use the result to govern their subspaces.

docs/architecture/adr-011-reports-module.md
```diff
@@ -0,0 +1,307 @@
...
+message MsgRemoveReasonResponse {
+  // Id of the newly supported reason
+  required uint32 reason_id = 1;
+}
+
+```
+
+### `Query` Service
+```protobuf
+service Query {
```
dadamu, line 224: It would be useful to have `ReportTallyResult` query to show the tally result. Moderator group can use the result to govern their subspaces.

RiccardoM: @dadamu Honestly I don't think that's necessary. Usually _tally results_ are related to something that is programmed to be closed in the future (eg. governance proposal voting period, poll voting period, etc). Talking about reports, there is not going to be a predefined time slot during which the reporting will be valid. For this reason, in my opinion is much better to simply allow to query the various reports for something rather than having a tally result endpoint which would return the same values anyway.

dadamu: Yeah, you are right. 

Thinking about how to propose a proposal to delete a post whose harassment report amount is greater than 100 to a subspace DAO. Since this action is done by a member of DAO with the report tally result, and it is not necessary to have a tally endpoint for it.

dadamu (approved):
Looks good to me.

PR #847, opened 2022-04-27
feat: ADR-010: Posts module 

RiccardoM:
## Description

This PR implements the `x/posts` module as per ADR-010.

Implementation checklist
- [x] Types
- [x] Keeper
- [x] CLI
- [x] Simulation tests
- [x] Invariants
- [ ] `spec` folder

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [x] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] confirmed all author checklist items have been addressed
- [x] reviewed state machine logic
- [x] reviewed API design and naming
- [x] reviewed tests and test coverage


x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+	}
+}
+
+// Validate implements fmt.Validator
+func (u Url) Validate() error {
+	if u.Start > u.End {
+		return fmt.Errorf("invalid start and end indexes: %d %d", u.Start, u.End)
+	}
+
+	if strings.TrimSpace(u.Url) == "" {
```
bragaz, line 349: shouldn't we check also that a URL starts with http/https to ensure its validity?

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+	}
+
+	attachmentID, err := strconv.ParseUint(value, 10, 32)
+	if err != nil {
+		return 0, fmt.Errorf("invalid subspace id: %s", err)
+	}
+	return uint32(attachmentID), nil
+}
+
+// Attachments represents a slice of Attachment ojects
```
bragaz, line 371:
```suggestion
// Attachments represents a slice of Attachment objects
```

x/posts/module.go
```diff
@@ -0,0 +1,211 @@
...
+
+// RegisterServices registers module services.
+func (am AppModule) RegisterServices(cfg module.Configurator) {
+	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
+	types.RegisterQueryServer(cfg.QueryServer(), am.keeper)
+}
+
+// NewAppModule creates a new AppModule Object
+func NewAppModule(
+	cdc codec.Codec, keeper keeper.Keeper, sk subspaceskeeper.Keeper, ak authkeeper.AccountKeeper, bk bankkeeper.Keeper, fk feeskeeper.Keeper,
```
bragaz, line 115: Are `auth`, `bank` and `fees` keepers necessaries here?

proto/desmos/posts/v1/models.proto
```diff
@@ -0,0 +1,258 @@
...
+  google.protobuf.Timestamp creation_date = 10
+      [ (gogoproto.nullable) = false, (gogoproto.stdtime) = true ];
+
+  // (optional) Last edited time of the post
+  google.protobuf.Timestamp last_edited_date = 11
+      [ (gogoproto.stdtime) = true ];
+}
+
+// PostReference contains the details of a post reference
+// TODO: Move this inside the Post object
```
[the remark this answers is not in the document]
RiccardoM: I thought of moving it to be a inner type because the only place where the `PostReference` type is used is within the `Post` type itself. There are no other ways, so it doesn't make much sense to leave it where it is right now. Also, the current naming contains the unwanted `Post` prefix which could then be deleted (the go type would become `Post_Reference`)

proto/desmos/posts/v1/models.proto
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Attachment contains the data of a single post attachment
+message Attachment {
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
+  // Id of the subspace inside which the post to which this attachment should be
+  // connected is is
+  uint64 subspace_id = 1 [ (gogoproto.customname) = "SubspaceID" ];
```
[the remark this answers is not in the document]
RiccardoM: Saving the subspace id inside the attachment itself results in a cleaner handling overall. A very nice example of how this could be useful to developers is if a DApp reads the data from two different subspaces (for any reason): it this case they might have multiple attachments with the same `post_id` and `id` values, unless we return the `subspace_id` inside the query response or somehow they manage this by themselves. This is way harder than simply storing a `uint64` value (just 8 bytes) inside the storage itself in my opinion

x/posts/types/models.go
RiccardoM: I didn't add this check because there might be other different protocols that are valid and would require too much work to support. An example is the `ftp` protocol which could be used if someone hosts a file on an FTP server. Or other custom schemes that might be used by different file storages. For this reason, I think it's easier to simply leave validity checks to the DApps and allow them to store whatever URL they prefer here

x/posts/module.go
RiccardoM: Yes, they are later used inside the `WeightedOperations` method to create the weighted simulation operations: 
```go
func (am AppModule) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
  return simulation.WeightedOperations(simState.AppParams, simState.Cdc, am.keeper, am.sk, am.ak, am.bk, am.fk)
}
```

x/posts/keeper/attachments.go
```diff
@@ -0,0 +1,87 @@
...
+	attachment, found := k.GetAttachment(ctx, subspaceID, postID, attachmentID)
+	if !found {
+		return
+	}
+
+	// Delete the attachment
+	store.Delete(types.AttachmentStoreKey(attachment.SubspaceID, attachment.PostID, attachment.ID))
+
+	// Delete the poll user answers
+	for _, answer := range k.GetPollUserAnswers(ctx, attachment.SubspaceID, attachment.PostID, attachment.ID) {
```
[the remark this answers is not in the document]
RiccardoM: You're right, thanks!

x/posts/keeper/invariants.go
```diff
@@ -0,0 +1,222 @@
...
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// ValidActivePollsInvariant checks that all the active polls are valid
+func ValidActivePollsInvariant(k Keeper) sdk.Invariant {
+	return func(ctx sdk.Context) (message string, broken bool) {
+		var invalidActivePolls []types.Attachment
+		k.IterateActivePolls(ctx, func(_ int64, attachment types.Attachment) (stop bool) {
+			poll := attachment.Content.GetCachedValue().(*types.Poll)
+
```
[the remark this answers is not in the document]
RiccardoM: This is not needed. The validation of the poll is already performed inside the `ValidAttachmentsInvariant`. This will iterate over polls anyway so no need to double check for validity here since the active poll queue only stores a reference to the polls and not different poll objects from the ones that are iterated over within the `ValidAttachmentsInvariant`

x/posts/keeper/invariants.go
```diff
@@ -0,0 +1,222 @@
...
+			return false
+		})
+
+		return sdk.FormatInvariant(types.ModuleName, "invalid posts",
+			fmt.Sprintf("the following posts are invalid:\n %s", formatOutputPosts(invalidPosts)),
+		), invalidPosts != nil
+	}
+}
+
+// formatOutputPosts concatenates the given posts information into a string
```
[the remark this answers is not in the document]
RiccardoM: I've just named these methods this way due to the fact that they accept a `[]Post` and not `[]PostID` due to the fact that they need to also print out the subspace id associated to the post (only using a post id is not sufficient due to the fact that post ids are not unique across subspaces). The naming does not refer to what it does internally rather then input and output

x/posts/keeper/posts.go
```diff
@@ -0,0 +1,174 @@
...
+	if bz == nil {
+		return 0, sdkerrors.Wrapf(types.ErrInvalidGenesis, "initial post ID hasn't been set for subspace %d", subspaceID)
+	}
+
+	postID = types.GetPostIDFromBytes(bz)
+	return postID, nil
+}
+
+// DeletePostID removes the post id key for the given subspace
+func (k Keeper) DeletePostID(ctx sdk.Context, subspaceID uint64) {
```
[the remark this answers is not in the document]
RiccardoM: I agree the naming is fancy, but it reflects the associated ones (`SetPostID` and `GetPostID`). I couldn't find any better alternative, but I'm open to suggestions

x/posts/types/events.go
```diff
@@ -0,0 +1,19 @@
+package types
+
+// Posts module event types
+const (
+	EventTypeCreatePost           = "create_post"
+	EventTypeEditPost             = "edit_post"
```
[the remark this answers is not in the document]
RiccardoM: Added, thanks!

proto/desmos/posts/v1/models.proto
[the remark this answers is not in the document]
bragaz: Oh. Right. I didn't think it this way. Yes it's definitely better to handle this with the additional `subspace_id` rather than leaving developers handle it in many different ways. 
It would only bring chaos and makes thing messy for no reason.

x/posts/keeper/posts.go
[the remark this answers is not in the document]
bragaz: I thin the `SetPostID` name is fine, but we can enhance the other two.
What about:
- `GetLastPostID` (Since the function is retrieving the last add post ID)
- `DeleteLastPostID` (Since the function is deleting the last added post ID)
- 
They will be more verbose but at least it will be clear by reading them what they're doing.
And I think this will ease devs job when looking at the code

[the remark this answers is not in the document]
RiccardoM: To make everything even more clear I've just renamed all those methods to be `[Set/Get/Delete]NextPostID` and the attachment ones to be `[Set/Has/Get/Delete]NextAttachmentID` as this should be even better to understand properly what they are related to

proto/desmos/posts/v1/models.proto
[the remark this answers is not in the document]
RiccardoM: Ended up simply deleting this TODO as I changed my mind. Going to stick to the ADR specification as the model is used elsewhere as well (eg. inside `MsgCreatePost`)

x/posts/types/keys.go
```diff
@@ -0,0 +1,157 @@
...
+		panic(fmt.Errorf("unexpected key length (%d ≠ %d)", len(key[1:]), lenTime+8))
+	}
+
+	endTime, err := sdk.ParseTimeBytes(key[1 : 1+lenTime])
+	if err != nil {
+		panic(err)
+	}
+
+	subspaceID = subspacetypes.GetSubspaceIDFromBytes(key[1+lenTime : 1+lenTime+8])
+	postID = GetPostIDFromBytes(key[1+8+lenTime : 1+16+lenTime])
```
dadamu, line 127: Should we replace all magical numbers to named variable like lenTime?

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid entities: %s", err)
+		}
+
+		// Make sure that no entity has a start or end index that is greater to the text length
+		maxIndexAllowed := uint64(0)
+		if len(strings.TrimSpace(p.Text)) > 0 {
+			maxIndexAllowed = uint64(len(strings.TrimSpace(p.Text)) - 1)
+		}
+
```
dadamu, line 79: It is missing to validate the content of entities. Currently the following post can pass the validation:
```json
{
    ...
    "text": "Hello #Desmos",
    "entities": [{
        "hash_tags": [
            {
                "start": 0,
                "end": 3,
                "tag": "desmos"
            }
        ]
    }]
}
```

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+func (p *Poll) Validate() error {
+	if strings.TrimSpace(p.Question) == "" {
+		return fmt.Errorf("invalid question: %s", p.Question)
+	}
+
+	if len(p.ProvidedAnswers) < 2 {
+		return fmt.Errorf("insufficient amount of provided answers: %d", len(p.ProvidedAnswers))
+	}
+
+	answers := map[string]int{}
```
dadamu, line 538: It can be simplified to be `map[string]bool{}`.

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+		Text:       text,
+		Entities:   entities,
+		UpdateTime: updateTime,
+	}
+}
+
+// Update updates the fields of a given post without validating it.
+// Before storing the updated post, a validation with keeper.ValidatePost should
+// be performed.
+func (p Post) Update(update *PostUpdate) Post {
```
dadamu, line 200: Can we use non-pointer parameter `PostUpdate` here?

x/posts/types/models.go
RiccardoM: This is wanted. I believe tags should not have any validation because they might be used for some particular cases. 

As an example, let's suppose I want to tag you into a post. I would then write: 
```
Hi @dadamu! Welcome!
```

At this post, the tag would be something like
```
{
  "start": 3,
  "end": 9,
  "tag": "desmos1....."
}
```

As you can see, there is no way we can validate this properly. I prefer to leave it to the applications to handle these cases freely without any constrictions

x/posts/types/events.go
```diff
@@ -0,0 +1,21 @@
...
+	EventTypeDeletePost           = "delete_post"
+	EventTypeAddPostAttachment    = "add_post_attachment"
+	EventTypeRemovePostAttachment = "add_remove_attachment"
```
dadamu, line 9:
```suggestion
	EventTypeRemovePostAttachment = "remove_attachment"
```

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+	}
+
+	if len(segments) < 2 {
+		// With less than 2 segments there cannot be any overlap
+		return nil
+	}
+
+	// Sort the segments
+	sort.Slice(segments, func(i, j int) bool {
+		return segments[i].start < segments[j].end
```
dadamu, line 271: What is the reason why sorts it by `segments[i].start < segments[j].end` instead of `segments[i].start < segments[j].start`.

RiccardoM: Because I'm stupid :laughing: Fixed it, thanks!

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+
+	if p.LastEditedDate != nil {
+		// Instead of zero, we should use nil
+		if p.LastEditedDate.IsZero() {
+			return fmt.Errorf("invalid post last edited date: %s", err)
+		}
+
+		// Make sure the creation date is always before the last edit date
+		if p.LastEditedDate.Before(p.CreationDate) {
+			return fmt.Errorf("creation date cannot be before last edit date")
```
bragaz, line 123:
```suggestion
			return fmt.Errorf("last edited date cannot be before the creation date")
```

x/posts/types/models.go
```diff
@@ -0,0 +1,669 @@
...
+		first, second := segments[index], segments[index+1]
+		if first.end >= second.start {
+			return fmt.Errorf("entities cannot overlap: start %d end %d", first.end, second.start)
+		}
+	}
+
+	return nil
+}
+
+type entitySegment struct {
```
bragaz, line 285: Can we add docs to this struct and generate it with protobuf instead of manually coding it?

RiccardoM: I've added the documentation, but I don't think this should be generated using Protobuf. From the documentation, you will see that it's only a type definition that should be private within the `types` package as it's only used temporarily during the validation of an `Entity` object. It's not used elsewhere, hence the decision of making it private and not accessible from anywhere else

bragaz (approved):
Looks ready to me 👍

RiccardoM:
Thanks @bragaz. I've added the automerge label. As soon as @dadamu confirms it's ready, it will be merged

proto/desmos/posts/v1/models.proto
```diff
@@ -0,0 +1,257 @@
...
+  uint64 post_id = 2 [ (gogoproto.customname) = "PostID" ];
+
+  // Id of the poll to which this answer is associated
+  uint32 poll_id = 3 [ (gogoproto.customname) = "PollID" ];
+
+  // Indexes of the answers inside the ProvidedAnswers array
+  repeated uint32 answers_indexes = 4;
+
+  // Address of the user answering the poll
+  bytes user = 5 [ (gogoproto.casttype) =
```
dadamu, line 226: Why is the type of user is `AccAddress` instead of bech32 prefixed string? Is it to prevent the double vote with the same address via ibc or something else?

x/posts/keeper/subspaces_hooks.go
```diff
@@ -0,0 +1,68 @@
...
+		h.k.DeletePost(ctx, post.SubspaceID, post.ID)
+
+		// Emit an event
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeDeletePost,
+				sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", post.SubspaceID)),
+				sdk.NewAttribute(types.AttributeKeyPostID, fmt.Sprintf("%d", post.ID)),
+			),
+		)
```
dadamu, line 45: It will create a tons of indexes for a block if there are millions of posts in a subspace. Can we simplify the event to be a general one like `EventTypeDeleteAllSubspacePost`?

x/posts/keeper/subspaces_hooks.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Hooks represents a wrapper struct
+type Hooks struct {
+	k Keeper
+}
+
+var _ subspacestypes.SubspacesHooks = Hooks{}
+
+// Hooks creates new relationships hooks
```
dadamu, line 20:
```suggestion
// Hooks creates new subspaces hooks
```

x/posts/types/models.go
```diff
@@ -0,0 +1,676 @@
...
+
+// Attachments represents a slice of Attachment objects
+type Attachments []Attachment
+
+// NewAttachments returns a new Attachments instance
+func NewAttachments(attachments ...Attachment) Attachments {
+	return Attachments(attachments)
+}
+
+// Validate implement
```
dadamu, line 386:
```suggestion
```

x/posts/types/models.go
```diff
@@ -0,0 +1,676 @@
...
+	}
+}
+
+// Validate implements fmt.Validator
+func (r *PollTallyResults) Validate() error {
+	if len(r.Results) == 0 {
+		return fmt.Errorf("empty answer results")
+	}
+
+	ids := map[uint32]int{}
```
dadamu, line 659: It could be simplified to be `map[uint31]bool{}` as well.

dadamu, line 659: As above

x/posts/types/models.go
```diff
@@ -0,0 +1,676 @@
...
+
+	if a.PollID == 0 {
+		return fmt.Errorf("invalid poll id: %d", a.PollID)
+	}
+
+	if len(a.AnswersIndexes) == 0 {
+		return fmt.Errorf("answer indexes cannot be empty")
+	}
+
+	indexes := map[uint32]int{}
```
dadamu, line 628: As above

x/posts/types/msgs.go
```diff
@@ -0,0 +1,413 @@
...
+	if msg.PollID == 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "invalid poll id: %d", msg.PollID)
+	}
+
+	if len(msg.AnswersIndexes) == 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "at least one answer is required")
+	}
+
+	// Check duplicated answers
+	answers := map[uint32]int{}
```
dadamu, line 388: As above

x/posts/keeper/posts.go
```diff
@@ -0,0 +1,183 @@
...
+// post's author has not blocked the user referencing the post
+func (k Keeper) ValidatePostReference(ctx sdk.Context, postAuthor string, subspaceID uint64, referenceID uint64) error {
+	// Make sure the referenced post exists
+	referencedPost, found := k.GetPost(ctx, subspaceID, referenceID)
+	if !found {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "post with id %d does not exist", referenceID)
+	}
+
+	// Make sure the original author has not blocked the post author
+	if k.HasUserBlocked(ctx, referencedPost.Author, postAuthor, subspaceID) {
```
dadamu, line 52: Should it check the block in genral subspace in general subspace 0?

x/posts/keeper/posts.go
```diff
@@ -0,0 +1,183 @@
...
+func (k Keeper) ValidatePostReply(ctx sdk.Context, postAuthor string, subspaceID uint64, referenceID uint64) error {
+	replyPost, found := k.GetPost(ctx, subspaceID, referenceID)
+	if !found {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "post with id %d does not exist", referenceID)
+	}
+
+	switch replyPost.ReplySettings {
+	case types.REPLY_SETTING_FOLLOWERS:
+		// We need to make sure that a relationship between post author -> original author exists
+		if !k.HasRelationship(ctx, postAuthor, replyPost.Author, subspaceID) {
```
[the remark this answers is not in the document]
dadamu: nevermind there is no relationship in subspace 0.

x/posts/keeper/msg_server.go
```diff
@@ -0,0 +1,478 @@
...
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, sdk.MsgTypeURL(msg)),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Signer),
+		),
+		sdk.NewEvent(
+			types.EventTypeAnswerPoll,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", msg.SubspaceID)),
+			sdk.NewAttribute(types.AttributeKeyPostID, fmt.Sprintf("%d", msg.PollID)),
```
dadamu, line 472:
```suggestion
			sdk.NewAttribute(types.AttributeKeyPostID, fmt.Sprintf("%d", msg.PostID)),
```

x/posts/keeper/posts.go
```diff
@@ -0,0 +1,183 @@
...
+	return nil
+}
+
+// ValidatePost validates the given post based on the current params, returning an error if anything is wrong
+func (k Keeper) ValidatePost(ctx sdk.Context, post types.Post) error {
+	params := k.GetParams(ctx)
+
+	// Validate the conversation reference
+	if post.ConversationID != 0 {
+		err := k.ValidatePostReference(ctx, post.Author, post.SubspaceID, post.ConversationID)
```
dadamu, line 96: The conversation post has the same function with referenced post. Should we merge it into referenced post with `conversation` type?

proto/desmos/posts/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+    option (google.api.http).get =
+        "/desmos/posts/v1/{subspace_id}/posts/{post_id}/attachments";
+  }
+
+  // PollAnswers queries the answers for the poll having the given id
+  rpc PollAnswers(QueryPollAnswersRequest) returns (QueryPollAnswersResponse) {
+    option (google.api.http).get = "/desmos/posts/v1/{subspace_id}/posts/"
+                                   "{post_id}/polls/{poll_id}/answers";
+  }
+
```
dadamu, line 37: `PollTallyResults` is missing.

x/posts/keeper/polls.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	k.IteratePollUserAnswers(ctx, subspaceID, postID, pollID, func(_ int64, answer types.UserAnswer) (stop bool) {
+		// Update the results
+		for _, answerIndex := range answer.AnswersIndexes {
+			results[answerIndex]++
+		}
+
+		// Delete the user answer
+		k.DeleteUserAnswer(ctx, answer.SubspaceID, answer.PostID, answer.PollID, answer.User)
```
dadamu, line 50: The `PollAnswers` query will always return empty array if we remove it after tally. Should we remain `UserAnswer`.

x/posts/keeper/msg_server.go
```diff
@@ -0,0 +1,478 @@
...
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, sdk.MsgTypeURL(msg)),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Author),
+		),
+		sdk.NewEvent(
+			types.EventTypeCreatePost,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", msg.SubspaceID)),
+			sdk.NewAttribute(types.AttributeKeyPostID, fmt.Sprintf("%d", post.ID)),
+			sdk.NewAttribute(types.AttributeKeyAuthor, msg.Author),
+			sdk.NewAttribute(types.AttributeKeyCreationTime, post.CreationDate.Format(time.RFC3339)),
```
dadamu, line 104: I have a dump question, why timestamp should be indexed?

x/posts/keeper/subspaces_hooks.go
RiccardoM: I've deleted it. The event `EventTypeSubspaceDeleted` should be sufficient to allow clients to react and delete the associated posts as well

x/posts/keeper/posts.go
RiccardoM: No, it shouldn't. We have designed blocks to be subspace-specific: if you block someone on the subspace with id `0`, then it means that user will not be able to send you DTag transfer requests. But they will be still be able to tag you/reply you inside other subspaces. You should block them there as well if you don't want them to interact with you at all

x/posts/keeper/posts.go
RiccardoM: The idea of having a conversation is to allow replicating something similar to Twitter conversations. The `ConversationID` represents the first post of the conversation (which can continue as a series of posts). A `PostReference`, on the other hand, should be used for reposts, quotes and other things. But the conversation id should remain separate to easily identify all the posts of the same conversation in my opinion

proto/desmos/posts/v1/query.proto
RiccardoM: This has been removed because poll tally results are now directly placed inside the `Poll` object, just like the Cosmos SDK handles a proposal's tally results

x/posts/keeper/polls.go
RiccardoM: This is done on purpose in order to avoid filling the store with useless data that can be queried elsewhere. Even after deleted, clients will be able to get the user answers by querying for the following event: 
```
answer_poll.subspace_id={subspace_id}&answer_poll.post_id={post_id}&answer_poll.poll_id={poll_id}
```

This is the same reason why the Cosmos SDK removes a proposal's votes after the results have been tallied

x/posts/keeper/msg_server.go
RiccardoM: This is done in order to allow clients to get the creation time from the returned events. Since it is based on the block time, and this is not know to the client, the only way we have to return it to clients (so that they can update the post information locally) is though block events that can later be parsed

proto/desmos/posts/v1/query.proto
dadamu: Can you update the adr for it?

proto/desmos/posts/v1/models.proto
RiccardoM: Reverted to use a `string` instead of a `bytes` representation

proto/desmos/posts/v1/query.proto
RiccardoM: Done :heavy_check_mark:

dadamu (approved):
Ready to go

PR #860, opened 2022-05-09
feat: ADR-011: Reports module

RiccardoM:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/desmos/reports/v1/models.proto
```diff
@@ -0,0 +1,99 @@
...
+  uint32 reason_id = 3 [ (gogoproto.customname) = "ReasonID" ];
+
+  // (optional) Message attached to this report
+  string message = 4;
+
+  // Address of the reporter
+  string reporter = 5;
+
+  // Target of the report
+  google.protobuf.Any data = 6;
```
dadamu, line 32: I would name it `target`.

proto/desmos/reports/v1/models.proto
```diff
@@ -0,0 +1,99 @@
...
+
+  // Time in which the report was created
+  google.protobuf.Timestamp creation_date = 7
+      [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
+}
+
+// UserData contains the data of a report about a user
+message UserData {
```
dadamu, line 40: I would name it `UserTarget`.

proto/desmos/reports/v1/models.proto
```diff
@@ -0,0 +1,99 @@
...
+  option (gogoproto.equal) = true;
+  option (gogoproto.goproto_stringer) = true;
+
+  // Address of the reported user
+  string user = 1;
+}
+
+// PostData contains the data of a report about a post
+message PostData {
```
dadamu, line 49: I would name it `PostTarget`.

proto/desmos/reports/v1/msgs.proto
```diff
@@ -0,0 +1,131 @@
...
+  // Id of the reason this report has been created for
+  uint32 reason_id = 2 [ (gogoproto.customname) = "ReasonID" ];
+
+  // (optional) Message attached to this report
+  string message = 3;
+
+  // Address of the reporter
+  string reporter = 4;
+
+  google.protobuf.Any data = 5;
```
dadamu, line 47: As above, I would name it `target`.

proto/desmos/reports/v1/query.proto
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+// QueryReportsResponse is the request type for Query/Reports RPC method
+message QueryReportsRequest {
+  // Id of the subspace to query the reports for
+  uint64 subspace_id = 1;
+
+  // (optional) Data to query the reports for
+  google.protobuf.Any data = 2;
```
dadamu, line 38: I would name it `target`.

x/reports/types/models.go
```diff
@@ -0,0 +1,187 @@
...
+// UnpackInterfaces implements codectypes.UnpackInterfacesMessage
+func (r *Report) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
+	var data ReportData
+	return unpacker.UnpackAny(r.Data, &data)
+}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// ReportData represents a generic report data
+type ReportData interface {
```
dadamu, line 93: I would name it `ReportTarget`.

x/reports/types/events.go
```diff
@@ -0,0 +1,19 @@
+package types
+
+// DONTCOVER
+
+const (
+	EventTypeCreateReport          = "create_report"
+	EventTypeDeleteReport          = "delete_report"
+	EventTypeSupportStandardReason = "support_standard_reason"
+	EventTypeAddReason             = "add_reason"
+	EventTypeRemoveReason          = "add_reason"
```
dadamu, line 10:
```suggestion
	EventTypeRemoveReason          = "remove_reason"
```

x/reports/types/params.go
```diff
@@ -0,0 +1,122 @@
...
+type StandardReasons []StandardReason
+
+// NewStandardReasons returns a new instance of StandardReasons contains the given reasons
+func NewStandardReasons(reasons ...StandardReason) StandardReasons {
+	return reasons
+}
+
+// Validate implements fmt.Validator
+func (s StandardReasons) Validate() error {
+	ids := map[uint32]int{}
```
dadamu, line 86: It can be simplified to be `map[uint32]bool{}`.

x/reports/types/msgs.go
```diff
@@ -0,0 +1,275 @@
...
+
+// Type should return the action
+func (msg MsgRemoveReason) Type() string {
+	return ActionRemoveReason
+}
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgRemoveReason) ValidateBasic() error {
+	if msg.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", msg.Size())
```
dadamu, line 251:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", msg.SubspaceID)
```

x/reports/types/msgs.go
```diff
@@ -0,0 +1,275 @@
...
+
+// Type should return the action
+func (msg MsgAddReason) Type() string {
+	return ActionAddReason
+}
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgAddReason) ValidateBasic() error {
+	if msg.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", msg.Size())
```
dadamu, line 203:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", msg.SubspaceID)
```

x/reports/types/msgs.go
```diff
@@ -0,0 +1,275 @@
...
+
+// Type should return the action
+func (msg MsgSupportStandardReason) Type() string {
+	return ActionSupportReason
+}
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgSupportStandardReason) ValidateBasic() error {
+	if msg.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", msg.Size())
```
dadamu, line 154:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", msg.SubspaceID)
```

x/reports/types/msgs.go
```diff
@@ -0,0 +1,275 @@
...
+
+// Type should return the action
+func (msg MsgDeleteReport) Type() string {
+	return ActionDeleteReport
+}
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgDeleteReport) ValidateBasic() error {
+	if msg.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", msg.Size())
```
dadamu, line 106:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", msg.SubspaceID)
```

x/reports/types/msgs.go
```diff
@@ -0,0 +1,275 @@
...
+
+// Type should return the action
+func (msg MsgCreateReport) Type() string {
+	return ActionCreateReport
+}
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgCreateReport) ValidateBasic() error {
+	if msg.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", msg.Size())
```
dadamu, line 58:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", msg.SubspaceID)
```

x/reports/types/keys.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	ModuleName   = "reports"
+	RouterKey    = ModuleName
+	StoreKey     = ModuleName
+	QuerierRoute = ModuleName
+
+	ActionCreateReport  = "create_report"
+	ActionDeleteReport  = "delete_report"
+	ActionSupportReason = "support_reason"
+	ActionAddReason     = "add_reason"
```
dadamu, line 22: Missing `ActionSupportStandardReason`.

x/reports/types/models.go
RiccardoM: Renamed everything from _report data_ to _report target_

x/reports/types/models.go
```diff
@@ -0,0 +1,187 @@
...
+		Target:       targetAny,
+		Reporter:     reporter,
+		CreationDate: creationDate,
+	}
+}
+
+// Validate implements fmt.Validator
+func (r Report) Validate() error {
+	if r.SubspaceID == 0 {
+		return fmt.Errorf("invalid subspace id: %d", r.Size())
```
dadamu, line 56:
```suggestion
		return fmt.Errorf("invalid subspace id: %d", r.SubspaceID)
```

x/reports/module.go
```diff
@@ -0,0 +1,224 @@
...
+
+// AppModule implements an application module for the reports module.
+type AppModule struct {
+	AppModuleBasic
+	keeper keeper.Keeper
+	sk     subspaceskeeper.Keeper
+	pk     postskeeper.Keeper
+	ak     authkeeper.AccountKeeper
+	bk     bankkeeper.Keeper
+	fk     feeskeeper.Keeper
```
dadamu, line 108: ak, bk and fk are not used in `x/reports`. They could be removed.

x/reports/keeper/alias_functions.go
```diff
@@ -0,0 +1,192 @@
...
+		reasons = append(reasons, reason)
+		return false
+	})
+	return reasons
+}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// IterateReports iterates over all reports and performs the provided function
+func (k Keeper) IterateReports(ctx sdk.Context, fn func(index int64, report types.Report) (stop bool)) {
```
dadamu, line 105: `index` parameter can be removed from required function.

x/reports/keeper/alias_functions.go
```diff
@@ -0,0 +1,192 @@
...
+		stop := fn(i, report)
+		if stop {
+			break
+		}
+		i++
+	}
+}
+
+// IterateSubspaceReports iterates over all the given subspace reports and performs the provided function
+func (k Keeper) IterateSubspaceReports(ctx sdk.Context, subspaceID uint64, fn func(index int64, report types.Report) (stop bool)) {
```
dadamu, line 123: as above

x/reports/keeper/alias_functions.go
```diff
@@ -0,0 +1,192 @@
...
+	var reports []types.Report
+	k.IterateSubspaceReports(ctx, subspaceID, func(index int64, report types.Report) (stop bool) {
+		reports = append(reports, report)
+		return false
+	})
+	return reports
+}
+
+// IteratePostReports iterates over all the reports for the given post and performs the provided function
+func (k Keeper) IteratePostReports(ctx sdk.Context, subspaceID uint64, postID uint64, fn func(index int64, report types.Report) (stop bool)) {
```
dadamu, line 151: as above

x/reports/keeper/alias_functions.go
```diff
@@ -0,0 +1,192 @@
...
+		stop := fn(i, report)
+		if stop {
+			break
+		}
+		i++
+	}
+}
+
+// IterateUserReports iterates over all the reports for the given user and performs the provided function
+func (k Keeper) IterateUserReports(ctx sdk.Context, subspaceID uint64, user string, fn func(index int64, report types.Report) (stop bool)) {
```
dadamu, line 173: as above

x/reports/keeper/msg_server.go
```diff
@@ -0,0 +1,284 @@
...
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, sdk.MsgTypeURL(msg)),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Reporter),
+		),
+		sdk.NewEvent(
+			types.EventTypeCreateReport,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", msg.SubspaceID)),
+			sdk.NewAttribute(types.AttributeKeyReportID, fmt.Sprintf("%d", report.ID)),
```
dadamu, line 82: Missing `ReasonID` here.

x/reports/keeper/msg_server.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// Check if the report exists
+	report, found := k.GetReport(ctx, msg.SubspaceID, msg.ReportID)
+	if !found {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "report with id %d not found inside subspace %d", msg.ReportID, msg.SubspaceID)
+	}
+
+	// Check the permission to delete reports
+	isModerator := k.HasPermission(ctx, msg.SubspaceID, msg.Signer, subspacestypes.PermissionManageReports)
```
dadamu, line 110: It is weird to me that moderator can delete the report. Can you elaborate the usecase?

x/reports/keeper/msg_server.go
```diff
@@ -0,0 +1,284 @@
...
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, sdk.MsgTypeURL(msg)),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Reporter),
+		),
+		sdk.NewEvent(
+			types.EventTypeCreateReport,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", msg.SubspaceID)),
+			sdk.NewAttribute(types.AttributeKeyReportID, fmt.Sprintf("%d", report.ID)),
+			sdk.NewAttribute(types.AttributeKeyReporter, msg.Reporter),
```
dadamu, line 83: Should we add target to the event as well?

RiccardoM: It would be hard to do so since the target is an interface and would be serialized as a JSON inside a string. Not the best thing in my opinion

RiccardoM: I don't think it's necessary. The target of a report can easily be retrieved by taking the subspace and report ids (both returned inside the event) and using the `Report` query to get all the details.

x/reports/keeper/msg_server.go
```diff
@@ -0,0 +1,285 @@
...
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeyAction, sdk.MsgTypeURL(msg)),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.Signer),
+		),
+		sdk.NewEvent(
+			types.EventTypeDeleteReport,
+			sdk.NewAttribute(types.AttributeKeySubspaceID, fmt.Sprintf("%d", msg.SubspaceID)),
+			sdk.NewAttribute(types.AttributeKeyReportID, fmt.Sprintf("%d", msg.ReportID)),
```
dadamu, line 130: Missing `ReasonID` here

x/reports/types/genesis_test.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+		{
+			name: "duplicated reason returns error",
+			data: types.NewGenesisState(nil, []types.Reason{
+				types.NewReason(1, 1, "Spam", ""),
+				types.NewReason(1, 1, "Spam", ""),
+			}, nil, types.DefaultParams()),
+			shouldErr: true,
+		},
+		{
```
dadamu, line 41: Missing invalid reason case

x/reports/types/models_test.go
```diff
@@ -0,0 +1,234 @@
...
+		data      *types.UserTarget
+		shouldErr bool
+	}{
+		{
+			name:      "invalid user address returns error",
+			data:      types.NewUserTarget(""),
+			shouldErr: true,
+		},
+		{
+			name:      "valid data returns no error",
```
dadamu, line 139:
```suggestion
			name:      "valid target returns no error",
```

x/reports/types/genesis_test.go
```diff
@@ -0,0 +1,210 @@
...
+			name: "invalid report returns error",
+			data: types.NewGenesisState(
+				[]types.SubspaceDataEntry{
+					types.NewSubspacesDataEntry(1, 1, 2),
+				},
+				nil,
+				[]types.Report{
+					types.NewReport(
+						0,
```
dadamu, line 110: The subspace 0 has no next report id so the invalid report test is uncovered.
The test data should be:
```
types.NewReport(
    1,
    0,
    1,
    ...
)
```

proto/desmos/reports/v1/query.proto
```diff
@@ -0,0 +1,71 @@
...
+import "cosmos/base/query/v1beta1/pagination.proto";
+
+import "desmos/reports/v1/models.proto";
+
+option go_package = "github.com/desmos-labs/desmos/v3/x/reports/types";
+
+// Query defines the gRPC querier service.
+service Query {
+  // Reports allows to query the reports for a specific target
+  rpc Reports(QueryReportsRequest) returns (QueryReportsResponse) {
```
dadamu, line 17: Currently, to know if the user reported to the a target post or target user is required to progress as follows:
1. Get all the reports from `Reports` query
2. Search all reports whose reportor is the user via pagination

If there are millions of reports on a post, then the check to the report id for a user would be costly.
Therefore, I think we should have a `UserReport` query to improve the performance.

x/reports/keeper/msg_server.go
RiccardoM: There is no reason to add the `reason_id` and `target` to the deleted report. All that's enough to know is simply the report it since it's being deleted and not created/edited. I don't see a usecase where knowing only the id of a deleted report is not enough

x/reports/types/genesis_test.go
RiccardoM: There's no reason to add this test here since it' already present inside the `TestReport_Validate` method. Here we just need to make sure that an invalid report (for whatever reason it might be) can be properly identified and fails the entire genesis validation

x/reports/types/genesis_test.go
RiccardoM: I've removed the reason id and report id checks from `ValidateGenesis` and added the reason id check inside the invariants. This should make it easier and safer to perform all the checks properly: during the genesis initialization all data will be stored but, as soon as invariants checks are performed, an error will be thrown halting the chain.

proto/desmos/reports/v1/query.proto
RiccardoM: Since implementing this method would require a more complex logic and additional store keys, I don't think we should have it. All these kinds of queries can easily be implemented using an external data parser like [DJuno](https://github.com/desmos-labs/djuno) that can expose a GraphQL endpoint anyway

x/reports/keeper/msg_server.go
RiccardoM: In order to solve this problem, as well as to implement the query described [here](https://github.com/desmos-labs/desmos/pull/860#discussion_r886616253), I've added a new `reportedEvent` that is emitted when a report is created. 

If a post is reported, the event will contain the id of the post along with the subspace id and the reporter address.

If a user is reported, the event will instead contain the address of the reported user, the subspace id and the reporter address. 

This should allow to: 

1. query all the reports that were made to a now deleted post; 
2. query all the reports to a post/user that have been made from a specific user.

proto/desmos/reports/v1/query.proto
dadamu: Okay, I found that the query is not useful since we don't limit the user to report a post or a user multiple times. 
Shouldn't we check the user has reported yet when creating a report?

x/reports/types/genesis_test.go
dadamu: The genesis height could be non-zero and the invariant checker is based on `InvCheckPeriod`, meaning that the genesis height might not be the invariant check point.  As a result, desmos cannot detect the failure within report_id or reason_id when importing genesis. Honestly, I don't think we should remove the id checks from `ValidateGenesis`.

RiccardoM: Actually invariants are always checked inside the `x/crisis` module `InitGenesis` method, unless you explicitly tell to *not* do so (see [here](https://github.com/cosmos/cosmos-sdk/blob/main/x/crisis/module.go#L143)). This is also why such module should always be the last one to be initialized

dadamu: Got it, thanks your explanation.

proto/desmos/reports/v1/query.proto
RiccardoM: Initially I didn't check for duplicated reports due to the fact that a single report could be associated to a single reason. So, if you wanted to report a post for two reasons (e.g. spamming and inadequate contents) you would have to create two different reports. I've now changed this to allow a single report to be associated to multiple reasons. This allows to check for duplicated reports but requires us to remove the reporting reason ids from the events inside `MsgCreateReport` due to the fact that arrays cannot be indexed properly (I think it's fine anyway)

x/reports/keeper/msg_server.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	// Check the permission to delete reports
+	isModerator := k.HasPermission(ctx, msg.SubspaceID, msg.Signer, subspacestypes.PermissionManageReports)
+	canDelete := report.Reporter == msg.Signer && k.HasPermission(ctx, msg.SubspaceID, msg.Signer, subspacestypes.PermissionDeleteOwnReports)
+	if !isModerator && !canDelete {
+		return nil, sdkerrors.Wrap(subspacestypes.ErrPermissionDenied, "you cannot delete reports inside this subspace")
+	}
+
+	// Delete the post
```
dadamu, line 150:
```sugesstion
	// Delete the report
```

x/reports/keeper/reports_test.go
```diff
@@ -0,0 +1,664 @@
...
+		check      func(ctx sdk.Context)
+	}{
+		{
+			name:       "non existing report is deleted properly",
+			subspaceID: 1,
+			reportID:   1,
+			check: func(ctx sdk.Context) {
+				suite.Require().False(suite.k.HasReport(ctx, 1, 1))
+
+				// Check the content key
```
dadamu, line 620: The check can be removed safely in this case.

proto/desmos/reports/v1/query.proto
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// QueryReportsResponse is the request type for Query/Reports RPC method
+message QueryReportsRequest {
+  // Id of the subspace to query the reports for
+  uint64 subspace_id = 1 [ (gogoproto.moretags) = "yaml:\"subspace_id\"" ];
+
+  // (optional) Target to query the reports for
+  google.protobuf.Any target = 2 [ (gogoproto.moretags) = "yaml:\"target\"" ];
+
```
dadamu, line 39: Since we have the content key showing the relationship between the reported content and reporter, adding a `reporter` field would be useful to get the user report on a content (a post or a user).

x/reports/client/cli/tx.go
```diff
@@ -0,0 +1,339 @@
...
+	return cmd
+}
+
+// GetCmdAddReason returns the command allowing to add a new reporting reason
+func GetCmdAddReason() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-reason [subspace-id] [title] [[description]]",
+		Args:  cobra.RangeArgs(2, 3),
+		Short: "Add a new reporting reason",
+		Long:  "Add a new reporting reason with the given title and optional desciption to a subspace",
```
dadamu, line 262:
```suggestion
		Long:  "Add a new reporting reason with the given title and optional description to a subspace",
```

x/reports/client/cli/tx.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	cmd.Flags().String(FlagMessage, "", "Optional message associated with the report")
+
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// GetCmdReportPost returns the command allowing to report a user
```
dadamu, line 111:
```suggestion
// GetCmdReportPost returns the command allowing to report a post
```

dadamu (changes requested):
Overall is good to me, just some minor typos and the last question about the `Reports` query.

proto/desmos/reports/v1/query.proto
RiccardoM: Makes sense. I've now added this ability to the `Reports` query: the `reporter` field will be used only if a `target` is specified.

x/reports/types/genesis.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return data.Params.Validate()
+}
+
+// containsDuplicatedSubspacesData tells whether the given subspaces data slice
+// contains a duplicated entry for the same subspace id as the one given
+func containsDuplicatedSubspacesData(subspaces []SubspaceDataEntry, data SubspaceDataEntry) bool {
+	var count = 0
+	for _, r := range subspaces {
+		if r.SubspaceID == data.SubspaceID {
```
bragaz, line 78:
```suggestion
	for _, entry := range subspaces {
		if entry.SubspaceID == data.SubspaceID {
```

client/docs/swagger-ui/swagger.yaml
```diff
@@ -7880,2523 +7865,2188 @@ paths:
...
           description: >-
-            divider_exponent is a factor used to power the divider used to
-            convert the
+            offset is a numeric offset that can be used when key is unavailable.
 
-            supply to the desired representation.
+            It is less efficient than using key. Only one of offset or key
+            should
+
+            be set.
```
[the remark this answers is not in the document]
RiccardoM: Unfortunately this file is generated by the `make proto-swagger-gen` command and I don't think we should update it by hand. We can just leave it as it is since it shouldn't matter the formatting

dadamu (approved):
It is ready to me

x/reports/types/keys.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	ModuleName   = "reports"
+	RouterKey    = ModuleName
+	StoreKey     = ModuleName
+	QuerierRoute = ModuleName
+
+	ActionCreateReport          = "create_report"
+	ActionDeleteReport          = "delete_report"
+	ActionSupportStandardReason = "support_reason"
```
RiccardoM, line 21:
```suggestion
	ActionSupportStandardReason = "support_standard_reason"
```

bragaz (approved):

Issue #855, opened 2022-05-05
Improve the management of permissions


RiccardoM:
## Feature description 
Currently when handling permissions inside a subspace, we use the same method that is used inside Linux-based systems: permissions are managed using a single `uint` value which bytes are read individually. Although this system works properly and allows us to store a single `uint` value per user/group, it has some downsides:
1. permissions need to be sanitized before being stored, to make sure there are no collisions with values that might be added in the future;
2. each time there is a new value to be added, the `PermissionEverything` needs to be updated accordingly. An example of this can be seen [here](https://github.com/desmos-labs/desmos/pull/847/files#diff-7310cece7d1e145c902f5c6403e7cefb55bee56781283312b3355cbdd45169f4R43): two permissions where added (`PermissionInteractWithContent` and `PermissionEditOwnContent`) and thus the `PermissionEverything` had to be updated as well;
3. every time a module needs a permission that's specific for it, we need to update the `subspaces` module to include that permission as well.

Instead of doing this, I think we can have a better way of managing permissions that allows each module to register their own permissions without having to always modify the `subspaces` module each time there is a new one.

## Implementation proposal
I suggest to create the following code

```go
// Permission contains the details of a permission
type Permission struct {
  // Identifies the module which the permission is related to
  Module string 
  
  // Identifies the unique code of this permission within the module 
  Code uint16
}


// RegisterPermission allows to register the permission having the given code within the 
// provided module inside the list of supported permissions
func RegisterPermission(code uint16, module string) Permission {
  // ...
}
```

This way we could store a user permissions as follows:
```
PermissionPrefix | SubspaceID | UserAddress | -> ProtocolBuffer([]Permission)
```

Then, modules external to the `subspaces` one can simply register their custom permissions just like this:
```go
var (
  PermissionCreateContent = subspacestypes.RegisterPermission(1, types.ModuleName)
  PermissionEditContent   = subspacestypes.RegisterPermission(2, types.ModuleName)
)
```

And use them as follows:
```go
k.HasPermission(ctx, userAddress, 1, PermissionCreateContent)
```

Please let me know what you think about this. I will write an ADR if you think this change should be implemented @bragaz @dadamu

bragaz:
This is a smart approach as it will allows to upgrade the permissions easily and require an on-chain upgrade each time the `subspaces` module is edited.

PR #869, opened 2022-05-19
docs(adr): ADR-014: Improve the permission system

RiccardoM:
## Description

This ADR contains the details of how the current permission system can be improved, based on #855.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


docs/architecture/adr-014-improve-permissions.md
```diff
@@ -0,0 +1,202 @@
...
+
+```go
+var (
+	// registeredPermissions represents the list of permissions that are registered and should be considered valid
+	registeredPermissions []Permission
+)
+
+// RegisterPermission allows to register the permission with the given name and returns its value
+func RegisterPermission(permissionName string) Permission {
+	permission := Permission(strings.ToUpper(strings.ReplaceAll(permissionName, " ", "_")))
```
bragaz, line 47: Since users will be one day able to create their own, we should take under consideration the fact that `permissionName`s could arrive here in different formats, e.g. : `Permission.Manage.Subspace` or `Permission12!_Manage_Subspace`. 
To avoid this, I think we can rely on two possible regular expressions to make sure inserted permissions:
1. Only contains letters and spaces: `/^[a-zA-Z\s]*$/` no number or special symbols).
2. Are in `camelCase` format: `(?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])`

RiccardoM: I don't think we should have anything like this. Ideally, in the future, the only users that will be able to specify custom permissions will be subspaces owners or admins. It's going to be their responsibilities to create permissions with a readable name. Otherwise it's going to be something that makes their lives harder. It won't anyway impact any other user or subspace owner as permissions will be stored within each subspace anyway.

bragaz: Ok fine. Then I guess we're fine like this!

bragaz (approved):

docs/architecture/adr-014-improve-permissions.md
```diff
@@ -0,0 +1,202 @@
...
+### Negative
+
+- Storing `string` instead of `uint` takes up more storage space  
+
+### Neutral
+
+
+## Further Discussions
+
+In the future, we might even allow developers to register custom permissions within their own subspaces, and then request those permissions during the execution of different messages. This could be done by allowing them to specify a `MsgTypeURL -> []Permission` entry for each kind of message type, so that when a message with `MsgTypeURL` gets executed, the user needs to have the specified permission in order to successfully perform the request.
```
dadamu, line 193: Does it assume the subspace is controlled by contract, and the feature is about registering the permission for contract?

RiccardoM: No, the idea is to allow any subspace owner to set a subspace-specific set of permissions to be checked. This might be useful if they want some kind of custom permission to be set on a user before they can do something. An example might be allowing only specific users to interact with smart contracts inside their subspace (e.g. paying users) or anything else.

Anyway, this is there only as a reminder for future iterations that we might want to implement.

dadamu (approved):
Looks good to me.

PR #875, opened 2022-05-20
build(deps): bump ibc-go to v3.0.0 and wasmd to v0.27.0

bragaz:
## Description

Superseeds #874
Closes: #872
Closes: #873

Bumps [wasmd](https://github.com/CosmWasm/wasmd) to [v0.27.0](https://github.com/CosmWasm/wasmd/releases/tag/v0.27.0)

Includes also the changes of #874 since the two are strictly dependent.

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/desmos-labs/desmos/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://docs.cosmos.network/v0.44/building-modules/intro.html)
- [ ] included the necessary unit and integration [tests](https://github.com/desmos-labs/desmos/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [x] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [x] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #877, opened 2022-05-25
Example to fix reflect.wasm path

alpe:
For inspiration.


app/sim_test.go
```diff
@@ -153,7 +170,7 @@ func TestAppImportExport(t *testing.T) {
 	}()
 
 	app := NewDesmosApp(
-		logger, db, nil, true, map[int64]bool{}, DefaultNodeHome,
+		logger, db, nil, true, map[int64]bool{}, t.TempDir(),
```
alpe, line 173: Each simulation thread must run with heir own (wasm) folder

app/sim_test.go
```diff
@@ -113,7 +130,7 @@ func TestFullAppSimulation(t *testing.T) {
 
 	app := NewDesmosApp(
 		logger, db, nil, true, map[int64]bool{},
-		DefaultNodeHome, simapp.FlagPeriodValue, MakeTestEncodingConfig(), simapp.EmptyAppOptions{}, fauxMerkleModeOpt,
+		t.TempDir(), simapp.FlagPeriodValue, MakeTestEncodingConfig(), simapp.EmptyAppOptions{}, fauxMerkleModeOpt,
```
alpe, line 133: Each simulation thread must run with heir own (wasm) folder

app/sim_test.go
```diff
@@ -95,6 +99,19 @@ func SetupSimulation(dirPrefix, dbName string) (simtypes.Config, dbm.DB, string,
 			}
 		}
 	}
+	// set sim params to overwrite wasmd default reflect.wasm path
+	simParams := simtypes.AppParams{
+		wasmsim.OpReflectContractPath: []byte(`"testdata/reflect.wasm"`),
```
alpe, line 104: this path is relative to the `app/` folder.

app/sim_test.go
alpe:
```suggestion
		logger, db, nil, true, map[int64]bool{}, dir,
```

bragaz:
Thanks for the precious help @alpe!
I've fixed the #875 with your suggestion and now everything works!

Issue #853, opened 2022-05-04
Per-chain default address


RiccardoM:
## Feature description 
Currently when a user connects multiple external chain addresses to their Desmos profile, it is impossible for clients to understand which one of those should be used. This leads to client using different algorithms to determine the one that should be considered the "default" one. For this reason, we should allow users to specify, for each chain, which address is the default one to be used. If only an address is provided for a specific chain, that should be the one to be considered the default. 

## Implementation proposal
The best implementation that I could think of on the spot is to use additional storage keys to identify a default address. Such keys will be formed as 
```
DefaultAddressPrefix | UserAddress | ChainName -> External address
```
This will make sure that only a single default address can exist per each (user, chain) tuple. 

Please let me know what you think @dadamu @bragaz @kwunyeung

dadamu:
It is useful for the user to get their profile from chain link source address, it can be implemeted in `QueryProfile`.
Besides, providing `MsgDefaultChainLinkAddress` should be required to allow user to change the default chain link default address.
I agree with the changes.

PR #868, opened 2022-05-18
docs(adr): ADR-013: Per-chain default external address

RiccardoM:
## Description

This PR introduces the ADR for per-chain default external addresses (ref: #853). 

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [x] targeted the correct branch (see [PR Targeting](https://github.com/desmos-labs/desmos/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] followed the guidelines for [building modules](https://docs.cosmos.network/v0.44/building-modules/intro.html)
- [ ] included the necessary unit and integration [tests](https://github.com/desmos-labs/desmos/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [x] reviewed "Files changed" and left comments if necessary
- [x] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #887, opened 2022-06-01
feat: ADR-013: Per-chain default external address

dadamu:
## Description
This PR is the implementation of chain default external address described inside #868 .

Closes: #853

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/profiles/keeper/keeper_chain_links.go
```diff
@@ -95,3 +94,64 @@ func (k Keeper) DeleteAllUserChainLinks(ctx sdk.Context, user string) {
 		k.DeleteChainLink(ctx, link)
 	}
 }
+
+// getOldestUserChainByChain returns the oldest chain link of the given owner associated to the given chain name
+// If the such the chain link is not exists, return false instead.
```
RiccardoM, line 99:
```suggestion
// getOldestUserChainByChain returns the oldest chain link of the given owner associated to the given chain name.
// If such chain link does not exist, returns false instead.
```

I would also rename this to `getOldestUserChainLink`

x/profiles/keeper/keeper_chain_links.go
```diff
@@ -95,3 +94,64 @@ func (k Keeper) DeleteAllUserChainLinks(ctx sdk.Context, user string) {
...
+			oldestLink = link
+		}
+		return false
+	})
+
+	return oldestLink, found
+}
+
+// updateOwnerDefaultExternalAddress updates the default external address of the given chain name for the given owner
+// It must be performed after deleting the default external address chain link
```
RiccardoM, line 120: Why is it necessary to do this?

x/profiles/keeper/keeper_chain_links.go
```diff
@@ -95,3 +94,64 @@ func (k Keeper) DeleteAllUserChainLinks(ctx sdk.Context, user string) {
...
+	store := ctx.KVStore(k.storeKey)
+	return store.Has(types.DefaultExternalAddressKey(owner, chainName))
+}
+
+// isDefaultExternalAddress tells whether the given chain link is a default external address or not
+func (k Keeper) isDefaultExternalAddress(ctx sdk.Context, link types.ChainLink) bool {
+	store := ctx.KVStore(k.storeKey)
+	if !k.HasDefaultExternalAddress(ctx, link.User, link.ChainConfig.Name) {
+		return false
+	}
```
RiccardoM, line 154: I think these lines can be removed safely, as `string([]byte(nil))` will be `nil` and thus not equals to the link address itself

x/profiles/keeper/keeper_chain_links.go
```diff
@@ -95,3 +94,64 @@ func (k Keeper) DeleteAllUserChainLinks(ctx sdk.Context, user string) {
...
+		srcAddrData, err := types.UnpackAddressData(k.cdc, link.Address)
+		if err != nil {
+			panic(err)
+		}
+		k.SaveDefaultExternalAddress(ctx, owner, chainName, srcAddrData.GetValue())
+		return
+	}
+
+	// If owner has no chain link on the given chain name, then delete the key
+	store.Delete(types.DefaultExternalAddressKey(owner, chainName))
```
RiccardoM, line 134: I think it's better from a performance perspective putting the shorter branch inside an `if`: 

```suggestion
	if !found {
		// If the owner has no chain link on the given chain name, then delete the key
		store.Delete(types.DefaultExternalAddressKey(owner, chainName))
	}
	
	srcAddrData, err := types.UnpackAddressData(k.cdc, link.Address)
	if err != nil {
		panic(err)
	}
	k.SaveDefaultExternalAddress(ctx, owner, chainName, srcAddrData.GetValue())
```

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -25,6 +25,11 @@ func (k msgServer) LinkChainAccount(goCtx context.Context, msg *types.MsgLinkCha
 		return nil, err
 	}
 
+	// The first chain link of each chain for the owner will be set as default external address
+	if !k.HasDefaultExternalAddress(ctx, link.User, link.ChainConfig.Name) {
+		k.SaveDefaultExternalAddress(ctx, link.User, link.ChainConfig.Name, srcAddrData.GetValue())
+	}
```
RiccardoM, line 31: I think it's better to move this inside the keeper method that stores the chain link (`SaveChainLink`) instead of here

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -56,6 +61,10 @@ func (k msgServer) UnlinkChainAccount(goCtx context.Context, msg *types.MsgUnlin
 	// Delete the link
 	k.DeleteChainLink(ctx, link)
 
+	if k.isDefaultExternalAddress(ctx, link) {
+		k.updateOwnerDefaultExternalAddress(ctx, link.User, link.ChainConfig.Name)
+	}
```
RiccardoM, line 66: I think it's better to move this inside the keeper method `DeleteChainLink`

x/profiles/keeper/msg_server_chain_link.go
```diff
@@ -65,11 +74,40 @@ func (k msgServer) UnlinkChainAccount(goCtx context.Context, msg *types.MsgUnlin
...
-			sdk.NewAttribute(types.AttributeKeyChainLinkDestinationAddress, msg.Owner),
+			sdk.NewAttribute(types.AttributeKeyChainLinkExternalAddress, msg.Target),
+			sdk.NewAttribute(types.AttributeKeyChainLinkChainName, msg.ChainName),
+			sdk.NewAttribute(types.AttributeKeyChainLinkOwner, msg.Owner),
 		),
 	})
 
 	return &types.MsgUnlinkChainAccountResponse{}, nil
 }
+
```
bragaz, line 85: missing docs here

x/profiles/keeper/msg_server_chain_link.go
dadamu: From my side, `SaveDefaultExternalAddress` is different thing from `SaveChainLink`. 
Moving `SaveDefaultExternalAddress` to `SaveChainLink` will influence the behavior of `ImportGenesis`, say, the first chain link will be set as default even it is not `DefaultExnternalAddress`, so we need to update to the real one after.
As a result, I decided to let them separately.

x/profiles/keeper/msg_server_chain_link.go
dadamu: The same reason as moving `SaveDefaultExternalAddress` into `SaveChainLink`.

x/profiles/keeper/msg_server_chain_link.go
dadamu: From my side, SaveDefaultExternalAddress is different thing from SaveChainLink.
Moving SaveDefaultExternalAddress to SaveChainLink will influence the behavior of ImportGenesis, say, the first chain link will be set as default even it is not DefaultExnternalAddress, so we need to update to the real one after.
As a result, I decided to let them separately.

x/profiles/keeper/keeper_chain_links.go
dadamu: It updates the default address to the oldest chain link if the user unlink the default one. The idea is that the default address must exist if the profile has chain links.

x/profiles/keeper/msg_server_chain_link.go
RiccardoM: The problem with this approach (i.e. having the default address set inside the message handler) is that if we implement different ways of creating chain links (maybe using InterChain Accounts), then we will have to replicate these lines inside other parts of the code. I think it's much easier to: 
1. Move these lines inside the `SaveDefaultExternalAddress`
2. Make sure that inside `ImportGenesis` we first save the various chain links and *after that* we save the default links data. 

Thanks to simulation tests (particularly Import-Export tests) we should also be sure that the data is imported correctly

x/profiles/keeper/msg_server_chain_link.go
RiccardoM: I think we should move these lines inside the `DeleteChainLink` anyway. They are private functions that must be called each time a chain link is deleted. Leaving them here would make it hard to deal with any bug that might arise from calling the `DeleteChainLink` method from elsewhere (maybe in an IBC handler or something else). Also, with this feature we are not changing how messages are handled, so we shouldn't change the code here. We should change what is done under the hoods when a chain link is saved/deleted instead

RiccardoM (commented):
We should also update the `spec` to make sure we add: 
- the definition of a _Default Chain Link_
- the details of the store key used to save default chain links

x/profiles/keeper/msg_server_chain_link.go
dadamu: Got you, it makes more sense.

x/profiles/spec/07-client.md
```diff
@@ -116,6 +116,46 @@ pagination:
   total: "0"
 ```
 
+#### default-external-addresses
+The `default-external-addresses` command allows users to query for default external addresses in chain link optionally specifying an owner address and a chain name.
```
RiccardoM, line 120:
```suggestion
The `default-external-addresses` command allows users to query for default external addresses optionally specifying an owner address and a chain name.
```

x/profiles/spec/07-client.md
```diff
@@ -298,6 +338,18 @@ Example:
 desmos tx profiles unlink-chain "cosmos" cosmos18xnmlzqrqr6zt526pnczxe65zk3f4xgmndpxn2`
 ```
 
+#### set-default-external-address
+The `set-default-external-address` allows user to set a default external address.
```
RiccardoM, line 342:
```suggestion
The `set-default-external-address` command allows user to set a default external address.
```

x/profiles/spec/07-client.md
```diff
@@ -473,6 +525,46 @@ Example Output:
 }
 ```
 
+### DefaultExternalAddresses
+The `DefaultExternalAddresses` endpoint allows user to query for external default address optionally specifying an owner and chain name.
```
RiccardoM, line 529:
```suggestion
The `DefaultExternalAddresses` endpoint allows to query for external default addresses optionally specifying an owner and chain name.
```

x/profiles/spec/03-state.md
```diff
@@ -25,6 +25,11 @@ To make it possible to query chain links given a user address or given a chain n
 * Chain Link: `0x12 | User address | Chain name | External address | -> ProtocolBuffer(ChainLink)`
 * Chain Link Owner: `0x15 | ChainName | 0x00 | External address | 0x00 | User address | -> 0x01 `
 
+## Default External Address
+An external address to a spcified chain is stored using owner address and chain name as the key, which allows to easily query all the default addresses:
```
RiccardoM, line 29:
```suggestion
A chain external address is stored using the owner address and the chain name as the key:
```

x/profiles/spec/02-concepts.md
```diff
@@ -119,6 +119,9 @@ This will start an interactive prompt session allowing you to generate the prope
    This will effectively link your Desmos profile to the external chain address. The required argument is the (absolute) path to the file generated using the `create-chain-link-json` command.
 
 
+## Default External Address
+A default external address represents a default chain link which makes dealing with a single chain link per chain for a profile easier. Each chain must have a chain link as a default external address for interaction, it will pick up the oldest chain link as a default external address if the link of the default external address is unlinked.
```
RiccardoM, line 123:
```suggestion
Since each user can have multiple links to the same chain, we allow users to specify which address should be considered as the default one for each chain. By default, the first link created for each chain will be considered as the default one. Also, if a chain link was set as the default one and is later deleted, the oldest chain link for the same chain (if any) will be used as the default one.
```

RiccardoM (approved):
LGTM, great job!

x/profiles/spec/07-client.md
```diff
@@ -473,6 +525,46 @@ Example Output:
...
+      },
+      "creationTime": "2022-07-18T10:07:51.899288600Z"
+    }
+  ],
+  "pagination": {
+    "total": "1"
+  }
+}
+```
+
```
bragaz, line 567: Can you also add the REST endpoint inside the REST section?

Issue #890, opened 2022-06-06
Add the possibility of reacting to posts


RiccardoM:
## Feature description 
Currently the `x/posts` module has been merged into `master` so now we are ready to implement another fundamental piece of Desmos: reactions. 

## Implementation proposal
The new `x/reactions` module should allow users to react to post, and subspace owners to properly register custom reactions.

PR #898, opened 2022-06-08
feat: ADR-016: Reactions module

RiccardoM:
## Description
This PR implements ADR-016 adding the new `x/reactions` module. 

Closes: #890

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/reactions/types/genesis.go
```diff
@@ -0,0 +1,214 @@
...
+
+		err := reaction.Validate()
+		if err != nil {
+			return fmt.Errorf("invalid registered reaction: %s", err)
+		}
+	}
+
+	for _, entry := range data.PostsData {
+		if containsDuplicatedPostDataEntry(data.PostsData, entry) {
+			return fmt.Errorf("duplicated poost data entry: subspace id %d, post id %d", entry.SubspaceID, entry.PostID)
```
dadamu, line 71:
```suggestion
			return fmt.Errorf("duplicated post data entry: subspace id %d, post id %d", entry.SubspaceID, entry.PostID)
```

proto/desmos/reactions/v1/query.proto
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// QueryReactionsRequest is the request type for the Query/Reactions RPC method
+message QueryReactionsRequest {
+  // Id of the subspace to query the reactions for
+  uint64 subspace_id = 1;
+
+  // (optional) Post id to query the reactions for
+  uint64 post_id = 2;
+
```
dadamu, line 42: It would be useful to add `user` field that allows to filter by user.

x/reactions/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// GetCmdSetParams returns the command allowing to remove a reporting reason
```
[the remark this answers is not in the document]
RiccardoM:
```suggestion
// GetCmdSetParams returns the command allowing to set reactions parameters
```

proto/desmos/reactions/v1/query.proto
RiccardoM: This is extremely hard to implement as currently reactions are stored using the following key: 
```
ReactionPrefix | SubspaceID | PostID | ReactionID -> ProtocolBuffer(Reaction) 
```

This is done in order to allow a faster removal of one reaction from a post (which is going to be the most likely operation to be carried out). 

If we want to allow querying a reaction for a user, we would have to somehow store the user address inside the store key, to be able to iterate over them. This is going to later complicate the retrieval of a specific reaction. 

For this reasons, I believe we can simply let third party tools (e.g. DJuno) to deal with this.

bragaz (approved):

proto/desmos/reactions/v1/query.proto
dadamu: Instead of storing the user address, It can be easily implemented by `FilteredPaginate` in grpc_query, say:
```go
func (k Keeper) Reactions(ctx context.Context, request *types.QueryReactionsRequest) (*types.QueryReactionsResponse, error) {
        ...
	var reactions []types.Reaction
	pageRes, err := query.FilteredPaginate(reactionsStore, request.Pagination, func(key []byte, value []byte, accumulate bool) (bool, error) {
		var reaction types.Reaction
		if err := k.cdc.Unmarshal(value, &reaction); err != nil {
			return false, status.Error(codes.Internal, err.Error())
		}

		if request.SubspaceId != 0 && request.PostId != 0 && request.User != "" && request.User != reaction.Author {
			return false, nil
		}

		reactions = append(reactions, reaction)
                if accumulate {
			reactions = append(reactions, reaction)
		}
		return true, nil
	})
       ...
}
```
Although it has `O(N)` time complexity, we can limit the `user` field to be only available when `SubspaceID` and `PostID` are given to make sure the amount `N` no to be large.

x/reactions/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+	return cmd
+}
+
+// GetCmdEditRegisteredReaction returns the command allowing to edit a registered reaction
+func GetCmdEditRegisteredReaction() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "edit [subspace-id] [reason-id]",
+		Args:  cobra.ExactArgs(2),
+		Short: "Support a standard reporting reason",
+		Long:  "Add the support for the specific standard reporting reason inside the subspace",
```
dadamu, line 230:
```suggestion
		Long:  "Edit a existing reaction with reaction id inside the given subspace",
```

proto/desmos/reactions/v1/query.proto
RiccardoM: I didn't think of using `FilteredPaginate`, you're right! I've implemented it this way now by also requiring users to specify a non-zero post id (it was useless to get all reactions inside a subspace since they're strongly tied to a post anyway)

x/reactions/module.go
```diff
@@ -0,0 +1,227 @@
...
+func (AppModuleBasic) RegisterRESTRoutes(_ client.Context, _ *mux.Router) {}
+
+// RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the reactions module.
+func (AppModuleBasic) RegisterGRPCGatewayRoutes(clientCtx client.Context, mux *runtime.ServeMux) {
+	types.RegisterQueryHandlerClient(context.Background(), mux, types.NewQueryClient(clientCtx))
+}
+
+// GetTxCmd returns the root tx command for the reactions module.
+func (AppModuleBasic) GetTxCmd() *cobra.Command {
+	return nil
```
dadamu, line 84: Lost to add it.

x/reactions/module.go
```diff
@@ -0,0 +1,227 @@
...
+	// TODO: Implement this
+	// return cli.NewTxCmd()
+}
+
+// GetQueryCmd returns the root query command for the reactions module.
+func (AppModuleBasic) GetQueryCmd() *cobra.Command {
+	return nil
```
dadamu, line 91: as above

dadamu (changes requested):
Overall is good to me, just last thing that the cmd should be registered to module.

x/reactions/module.go
RiccardoM: Thanks, added it!

x/reactions/module.go
RiccardoM: Thanks, added it!

dadamu (approved):
Ready to go!

PR #968, opened 2022-07-18
feat: ADR-015: Improve chain link signature support

RiccardoM:
## Description

Closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


RiccardoM:
@dadamu @bragaz If you can review this it would be great. I will take care of updating the docs once everything is fine for you

bragaz (commented):
Awesome work! So this supports also the signatures made with Metamask right?

bragaz (approved):

RiccardoM:
> Awesome work! So this supports also the signatures made with Metamask right?

Yes, it does. It supports ETH `personal_sign` signature. The tests case present is actually testing against a signature generated with MetaMask

proto/desmos/profiles/v3/models_chain_links.proto
```diff
@@ -115,28 +119,87 @@ message HexAddress {
   string prefix = 2 [ (gogoproto.moretags) = "yaml:\"prefix\"" ];
 }
 
-// SingleSignatureData is the signature data for a single signer
-message SingleSignatureData {
+// --------------------------------------------------------------------------------------------------------------------
+
+// CosmosSingleSignature is the signature data for a single signer
+message CosmosSingleSignature {
```
RiccardoM, line 125: I'm thinking if there are better ways to name these type after what @dadamu said: 

> Luckily, thanks to [solana-wallet-adapter](https://github.com/solana-labs/wallet-adapter) providing the function signing a raw message(arbitrary string), we can use the same process as CosmosSingleSignature [...].

This means that what has been currently named `CosmosSingleSignature` is actually compatible with other ecosystems different from the Cosmos one. For this reason, I think a better name should be used instead. 

@dadamu Proposed `GeneralSingleSignature`, but maybe even a shorter `SingleSignature` can work. 

What do you think @bragaz @kwunyeung? 

**Note**  
By looking at the code, we know that the signing algorithm used to verify the signature will be based on the type of the private key used. The currently supported public key types are `Ed25519`, `Secp256r1` and `Secp256k1`. As @dadamu said, for Solana we should use `/cosmos.crypto.ed25519.PubKey` (hence why this signature also works for that ecosystem).

bragaz: I think we can use something like: `StandardSingleSignature` or `GeneralSingleSignature` is ok. Then we can add in the docs the supported chains of such signature

dadamu: @RiccardoM From my side, the `General` is compared to `EVM` or other unique spec signatures. Noting that `SingleSignature` is the general signature based on general private/public key signing should be enough as well.

x/profiles/legacy/v6/store.go
```diff
@@ -248,24 +249,48 @@ func convertChainLinkSignatureData(data *codectypes.Any, cdc codec.BinaryCodec)
...
+			panic(fmt.Sprintf("unsupported signing mode: %s", signature.Mode))
+		}
+
+		v6Signature := types.NewCosmosSingleSignature(signingMode, signature.Signature)
 		signatureAny, err = codectypes.NewAnyWithValue(v6Signature)
 		if err != nil {
 			panic(err)
 		}
 
 	case *v5types.MultiSignatureData:
```
dadamu, line 270: Can you add a test for `MultiSignatrueData` migrations?

proto/desmos/profiles/v3/models_chain_links.proto
RiccardoM: I ended up renaming `CosmosSingleSignature` to `SingleSignature`, and removing `EVMSignature` in favor or `SingleSignature` as well. Now EVM-compatible signatures can be specified by simply using the proper `SignatureValueEncoding` inside a `SingleSignature` itself.

x/profiles/legacy/v6/store.go
RiccardoM: Added it

proto/desmos/profiles/v3/models_chain_links.proto
```diff
@@ -115,28 +119,63 @@ message HexAddress {
...
+message SingleSignature {
   option (gogoproto.goproto_getters) = false;
   option (gogoproto.equal) = true;
-  option (cosmos_proto.implements_interface) = "SignatureData";
+  option (cosmos_proto.implements_interface) = "CosmosSignature";
 
-  // Mode is the signing mode of the single signer
-  cosmos.tx.signing.v1beta1.SignMode mode = 1;
+  // ValueEncoding represents how the value has been encoded before being signed
+  SignatureValueEncoding value_encoding = 1
```
dadamu, line 131: I would remain the naming `mode` with `SignMode` instead of `value_encoding` with `SignatureValueEncoding`.

RiccardoM: I thought about this too, but then I thought: _signature modes are things like Secp256k1, Ed25519, etc while in this case we want to specify how the signature value has been encoded before signing (Amino, Protobuf, PersonalSign, etc). So I think this makes more sense_. Does it make sense to you?

bragaz (approved):

proto/desmos/profiles/v3/models_chain_links.proto
dadamu: I think `PersonalSign` is not a encoding method since the plain text of it is raw text, it is the method to sign a raw text message into signature. Besides, `secp256k1` and `ed25519` algorithms are based on the keys in the [code base](https://github.com/cosmos/cosmos-sdk/tree/6f070623741fe0d6851d79ada41e6e2b1c67e236/proto/cosmos/crypto), since they use different curves to sign message with the key.

RiccardoM: Maybe we can just call it `SignatureType` then? And have `SIGNATURE_TYPE_COSMOS_AMINO`, `SIGNATURE_TYPE_EVM_PERSONAL_SIGN`, etc?

dadamu: @RiccardoM That is great! I agree with it.

RiccardoM: I ended up going to `SignatureValueType` which I think it's more explanatory than `SignatureType` which might be confused for something else

x/profiles/types/models_chain_links.go
```diff
@@ -159,84 +141,292 @@ func (p Proof) Verify(cdc codec.BinaryCodec, legacyAmino *codec.LegacyAmino, own
...
+
+	return nil
+}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+type CosmosSignature interface {
+	Signature
+	GetSignatureValueEncoding() (SignatureValueEncoding, error)
+}
```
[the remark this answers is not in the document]
RiccardoM: Yeah it makes sense. Done that

dadamu (approved):
Awesome job! Ready to ship~

PR #971, opened 2022-07-21
fix!: profile query returns nil if profile does not exist

dadamu:
## Description

Currently, querying a non-existing profile returns `nil` instead of error with `profile not found for address/dtag`.
This PR fixes the response of `Profile` behavior mentioned above. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/profiles/keeper/grpc_query.go
+++ b/x/profiles/keeper/grpc_query.go
@@ -47,7 +47,7 @@ func (k Keeper) Profile(ctx context.Context, request *types.QueryProfileRequest)
 	}
 
 	if !found {
-		return &types.QueryProfileResponse{Profile: nil}, nil
+		return nil, status.Errorf(codes.NotFound, "profile for dtag/address %s not found", dTagOrAddress)
 	}
 
 	accountAny, err := codectypes.NewAnyWithValue(account)
--- a/.changeset/entries/e8fdc65b9f41e85409395d7e7b317f59f6d7be1ff3f8a92ebb30dafeb3254d4e.yaml
+++ b/.changeset/entries/e8fdc65b9f41e85409395d7e7b317f59f6d7be1ff3f8a92ebb30dafeb3254d4e.yaml
@@ -1,0 +1,6 @@
+type: fix
+module: x/profiles
+pull_request: 971
+description: fix profile query returns nil to error if profile does not exist
+backward_compatible: true
+date: 2022-07-22T06:27:36.4265228Z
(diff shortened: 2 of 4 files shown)


x/profiles/keeper/grpc_query.go
```diff
@@ -47,7 +47,7 @@ func (k Keeper) Profile(ctx context.Context, request *types.QueryProfileRequest)
 	}
 
 	if !found {
-		return &types.QueryProfileResponse{Profile: nil}, nil
+		return nil, status.Errorf(codes.NotFound, "profile for dtag/address %s not found", dTagOrAddress)
```
RiccardoM, line 50: This is something I've thought about for a while, and honestly I don't remember why we chose to return `nil` instead of an error. However, I would like to take this chance to start a conversation about this. Particularly, I'm interested in knowing your opinion about the pros and cons that returning an error instead of `nil` might have. 

I would like to ask you guys (@dadamu, @bragaz, @manu0466) considering your experience in smart contract development, as well as @ale-mazz considering his experience in client-side development. Do you think that a missing profile should return an error or `null`, and why?

dadamu: This issue is found during building the `POAP-manager` contract. 

1. If it returns nil, the contract side need to handle the `Option` unwrap process like:
```rust
let profile_res = ProfilesQuerier::new(deps.querier.deref()).query_profile(user)?;
let profile = profile_res;
match profile {
    Some(profile) => Ok(operate(profile)),
    None => Err(some error)
}
```

2. If it returns an error, then the process can be simplified to be as follow since the `non found profile` case has been already handled by error:
```rust
let profile_res = ProfilesQuerier::new(deps.querier.deref()).query_profile(user)?;
let profile = profile_res;
operate(profile)
```
Besides, frontend also need a check if the profile exists or not if it returns `nil`. 
As a result, I would use returning error instead if the specific item does not exist.

ale-mazz: In my opinion, it would be more correct to return an error, so the frontend developer would have an advantage in relaunching the error in the UI or handling it in the way they prefer. Returning `null` might lead one to think of the presence of a possible backend error or otherwise something incorrect. I always consider error handling rather than returning `null`, `undefined`, etc. values to be superior.

RiccardoM: Ok thanks for the contributions. I think the best thing for clients then is to simply return `nil` so I will accept this PR as soon as @dadamu adds the proper changeset entry.

dadamu:
@RiccardoM Added.

RiccardoM (approved):

PR #1057, opened 2022-12-12
feat!: implement ADR-017 subspace grant

dadamu:
## Description

Closes: #XXXX
This PR implements ADR-017 subspace grant feature.

---

### Author Checklist

I have...

- [x] added `!` to the type prefix if API or client breaking change

### Reviewers Checklist

I have...


x/subspaces/types/keys.go
```diff
@@ -205,3 +211,35 @@ func SplitUserAddressPermissionKey(key []byte) (subspaceID uint64, sectionID uin
 	user = GetAddressFromBytes(key[lenSubspaceID+lenSectionID:])
 	return subspaceID, sectionID, user
 }
+
+// --------------------------------------------------------------------------------------------------------------------
+
```
dadamu, line 216: `ExpiringAllowanceQueuePrefix` for removing expired allowances is impossible to implement in 0.45.11 version of cosmos sdk because `FeeAllowanceI` misses the method called `ExpireAt`. We can implement the feature after updating the cosmos-sdk version over 0.46.

Reference: [0.46.0 FeeAllowanceI](https://github.com/cosmos/cosmos-sdk/blob/v0.46.0/x/feegrant/fees.go#L29) and [0.45.11 FeeAllowanceI](https://github.com/cosmos/cosmos-sdk/blob/v0.45.11/x/feegrant/fees.go#L9)

x/subspaces/types/keys.go
```diff
@@ -41,6 +45,8 @@ var (
 	UserPermissionsStorePrefix = []byte{0x05}
 	SectionIDPrefix            = []byte{0x06}
 	SectionsPrefix             = []byte{0x07}
+	UserAllowancePrefix        = []byte{0x08}
+	GroupAllowancePrefix       = []byte{0x09}
```
dadamu, line 49: `ExpiringAllowanceQueuePrefix` for removing expired allowances is impossible to implement in 0.45.11 version of cosmos sdk because current `FeeAllowanceI` misses the method called `ExpireAt`. We can implement the feature after updating the cosmos-sdk version over 0.46.

Reference: [0.46.0 FeeAllowanceI](https://github.com/cosmos/cosmos-sdk/blob/v0.46.0/x/feegrant/fees.go#L29) and [0.45.11 FeeAllowanceI](https://github.com/cosmos/cosmos-sdk/blob/v0.45.11/x/feegrant/fees.go#L9)

x/subspaces/keeper/groups.go
```diff
@@ -106,6 +113,16 @@ func (k Keeper) DeleteUserGroup(ctx sdk.Context, subspaceID uint64, groupID uint
...
+	// Create account if user does not exist.
+	userAcc, err := sdk.AccAddressFromBech32(user)
+	if err != nil {
+		panic(err)
+	}
+	if !k.ak.HasAccount(ctx, userAcc) {
+		defer telemetry.IncrCounter(1, "new", "account")
+		k.ak.SetAccount(ctx, k.ak.NewAccountWithAddress(ctx, userAcc))
+	}
+
```
dadamu, line 125: Users in group now are required to have an account, it is useful when users have a group grant.

x/subspaces/keeper/feegrant.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+// SaveUserGrant saves the given user grant
+func (k Keeper) SaveUserGrant(ctx sdk.Context, grant types.UserGrant) {
+	granteeAddr, err := sdk.AccAddressFromBech32(grant.Grantee)
+	if err != nil {
+		panic(err)
+	}
+	if !k.ak.HasAccount(ctx, granteeAddr) {
+		k.ak.SetAccount(ctx, authtypes.NewBaseAccountWithAddress(granteeAddr))
```
RiccardoM, line 18: Wer should add the following line to properly keep track of the account numbers: 
```
defer telemetry.IncrCounter(1, "new", "account")
```

We could even move these two lines in a new function named `createAccount` so that it's easier to manage. We would need this when adding a user to a group as well

x/subspaces/keeper/msg_server_feegrant.go
```diff
@@ -0,0 +1,148 @@
...
+		),
+	})
+	return &types.MsgGrantUserAllowanceResponse{}, nil
+}
+
+// RevokeUserAllowance defines a rpc method for MsgRevokeUserAllowance
+func (k msgServer) RevokeUserAllowance(goCtx context.Context, msg *types.MsgRevokeUserAllowance) (*types.MsgRevokeUserAllowanceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if !k.HasUserGrant(ctx, msg.SubspaceID, msg.Granter, msg.Grantee) {
```
RiccardoM, line 57: I think we should also check the permission to manage allowances here

x/subspaces/keeper/msg_server_feegrant.go
```diff
@@ -0,0 +1,148 @@
...
+		),
+	})
+	return &types.MsgGrantGroupAllowanceResponse{}, nil
+}
+
+// RevokeGroupAllowance defines a rpc method for MsgRevokeGroupAllowance
+func (k msgServer) RevokeGroupAllowance(goCtx context.Context, msg *types.MsgRevokeGroupAllowance) (*types.MsgRevokeGroupAllowanceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if !k.HasGroupGrant(ctx, msg.SubspaceID, msg.Granter, msg.GroupID) {
```
RiccardoM, line 127: I think we should check the permission to manage allowances here as well

x/subspaces/keeper/subspaces.go
```diff
@@ -100,6 +106,12 @@ func (k Keeper) DeleteSubspace(ctx sdk.Context, subspaceID uint64) {
 		return false
 	})
 
+	// Delete all sections
+	k.IterateSubspaceSections(ctx, subspaceID, func(section types.Section) (stop bool) {
+		k.DeleteSection(ctx, section.SubspaceID, section.ID)
+		return false
+	})
```
RiccardoM, line 113: I think this should be moved as first thing

x/subspaces/keeper/subspaces.go
```diff
@@ -88,9 +88,15 @@ func (k Keeper) DeleteSubspace(ctx sdk.Context, subspaceID uint64) {
...
+	k.IterateSubspaceUserGrants(ctx, subspaceID, func(grant types.UserGrant) (stop bool) {
+		k.DeleteUserGrant(ctx, grant.SubspaceID, grant.Granter, grant.Grantee)
+		return false
+	})
+
+	// Delete all the group grants
+	k.IterateSubspaceGroupGrants(ctx, subspaceID, func(grant types.GroupGrant) (stop bool) {
+		k.DeleteGroupGrant(ctx, grant.SubspaceID, grant.Granter, grant.GroupID)
 		return false
 	})
```
RiccardoM, line 101: Ideall, this is not necessary. Deleting the sections will cause to delete the groups, which will trigger the deletion of their  grants as well.

x/subspaces/legacy/v5/store.go
```diff
@@ -0,0 +1,36 @@
...
+	for ; iterator.Valid(); iterator.Next() {
+		_, _, user := types.SplitGroupMemberStoreKey(append(types.GroupsMembersPrefix, iterator.Key()...))
+		userAcc, err := sdk.AccAddressFromBech32(user)
+		if err != nil {
+			return err
+		}
+
+		accExists := accountKeeper.HasAccount(ctx, userAcc)
+		if !accExists {
+			accountKeeper.SetAccount(ctx, accountKeeper.NewAccountWithAddress(ctx, userAcc))
```
RiccardoM, line 32: Should we track this creation using the `telemetry` package as well?

x/subspaces/types/keys.go
RiccardoM: Is there any other way we can have this? Maybe backporting the changes to the `FeeAllowanceI` interface inside our Cosmos SDK fork?

x/subspaces/ante/common_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	ante ante.DeductFeeDecorator
+}
+
+func TestAnteTestSuite(t *testing.T) {
+	suite.Run(t, new(AnteTestSuite))
+}
+
+func (suite *AnteTestSuite) SetupTest() {
+	ctrl := gomock.NewController(suite.T())
```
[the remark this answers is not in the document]
dadamu: Yes, it is used inside the Cosmos SDK, they are using it to decouple the dependencies of the module now.
See https://github.com/cosmos/cosmos-sdk/blob/main/x/feegrant/keeper/keeper_test.go#L20

x/subspaces/types/keys.go
dadamu: Yes, backporting the new version `FeeAllowanceI`  inside our Cosmos SDK fork can solve the issue.

x/subspaces/keeper/msg_server_feegrant.go
dadamu: I finally decided not to check the permission in the `Revoke` since it makes the granter could not remove the grants granted when he had the permission.

x/profiles/types/expected_keeper.go
```diff
@@ -6,21 +6,10 @@ import (
...
 // DONTCOVER
 
-// SubspacesKeeper represents the expected keeper used to interact with subspaces
-type SubspacesKeeper interface {
-	// HasSubspace tells if the subspace with the given id exists
-	HasSubspace(ctx sdk.Context, subspaceID uint64) bool
-
-	// GetAllSubspaces returns all the subspaces stored
-	GetAllSubspaces(ctx sdk.Context) []subspacestypes.Subspace
-}
```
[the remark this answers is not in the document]
dadamu: `x/profiles` is not required to have `x/subspaces` now. It remains since `x/profiles` used to include `x/relationships` which couple with `x/subspaces`.

x/subspaces/keeper/msg_server_feegrant.go
dadamu: As above, it will make the granter could not remove the grants granted when he had the permission.

x/subspaces/keeper/subspaces.go
dadamu: Follow this idea, `k.IterateSubspaceUserGroups` can be removed as well since it is included in the `k.IterateSubspaceSections`. From my side, using subspace id to remove `UserGroups` and `UserGrants` will have better performance since they are indexed with `SubspaceID` instead of `SectionID` and `GroupID`.

x/subspaces/keeper/feegrant.go
```diff
@@ -0,0 +1,166 @@
...
+	store := ctx.KVStore(k.storeKey)
+	key := types.UserAllowanceKey(subspaceID, granter, grantee)
+	store.Delete(key)
+}
+
+// GetUserGrant returns the grant having the granter and grantee from the provided subspace.
+// If there is no grant associated with the info the function will return an error.
+func (k Keeper) GetUserGrant(ctx sdk.Context, subspaceID uint64, granter, grantee string) (types.UserGrant, error) {
+	if !k.HasUserGrant(ctx, subspaceID, granter, grantee) {
+		return types.UserGrant{}, fmt.Errorf("user grant does not exist: subspace id %d, granter %s, grantee %s", subspaceID, granter, grantee)
```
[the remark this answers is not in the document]
dadamu: Yes, you are right! I got lost when making decision the returnings of `get` method.

proto/desmos/subspaces/v3/models.proto
```diff
@@ -110,4 +112,49 @@ message UserPermission {
   uint32 section_id = 2 [ (gogoproto.customname) = "SectionID" ];
   string user = 3;
   repeated string permissions = 4 [ (gogoproto.castrepeated) = "Permissions" ];
+}
+
+// UserGrant represents a grant to a user
+message UserGrant {
```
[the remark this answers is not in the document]
dadamu: Never thought it. Such a great idea!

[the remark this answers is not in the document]
dadamu: Never thought it, such a great idea!

x/profiles/types/expected_keeper.go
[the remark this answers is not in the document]
RiccardoM: Perfect, thanks for the explanation

x/subspaces/ante/fee.go
```diff
@@ -0,0 +1,117 @@
...
+	if addr := dfd.ak.GetModuleAddress(authtypes.FeeCollectorName); addr == nil {
+		return ctx, false, fmt.Errorf("fee collector module account (%s) has not been set", authtypes.FeeCollectorName)
+	}
+
+	used := dfd.sk.UseGrantedFees(ctx, subspaceID, feeGranter, feePayer, fee, tx.GetMsgs())
+	if !used {
+		return ctx, false, nil
+	}
+
+	deductFeesFrom = feeGranter
```
RiccardoM, line 94: I just realized another thing, and I would like to start a discussion about this. 

Currently, this way of managing the grants allows one user to grant another user (or user group) the permission to use their funds to pay for fees. This is very similar to the current `x/feegrant` implementation, we're just adding the support for user groups and the spefication of working only for a single subspace. 

However, I imagined the thing to work different for subspaces. In particular, what I think would be more valuable for subspace admins is allowing users to pay for fees using the subspace `treasury` funds. The way I envision this working is: 
1. subspace admins grant a fee allowance to a user or a group 
2. when the user (or group) performs a transaction, if they have a fee allowance, the fee payer will become the subspace treasury account

This way we are not requiring admins to pay with their own balance but, instead, we are allowing all admins that have a permission, to create grant on behalf of the treasury account. Which should be the one effectively paying for the various fees

dadamu: It is interesting idea, but it will allow unexpected behaviour, which is, currently the subspace manager(s) can assign any accounts as `treasury`, then the victim account will pay the fees for the subspace. 

In order to solve the issue mentioned above, we need a another proof that the subspace has the access to control the treasury account, it would another implementation something like chain link but for subspace. The process will be like:
1. treasury account sign a proof whose target is subspace id
2. create a post with the proof
3. make a subspace link for the account
But it might have a another issue making contract as `granter` impossible since no one has the contract private key to build a link then grant users fee allowances.

Therefore, I prefer the current implementation to make sure the granter can control the account to pay for users.

x/subspaces/types/codec.go
```diff
@@ -28,9 +28,25 @@ func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
...
 
 	cdc.RegisterConcrete(&MsgSetUserPermissions{}, "desmos/MsgSetUserPermissions", nil)
+
+	cdc.RegisterConcrete(&MsgGrantUserAllowance{}, "desmos/MsgGrantUserAllowance", nil)
+	cdc.RegisterConcrete(&MsgRevokeUserAllowance{}, "desmos/MsgRevokeUserAllowance", nil)
+	cdc.RegisterConcrete(&MsgGrantGroupAllowance{}, "desmos/MsgGrantGroupAllowance", nil)
+	cdc.RegisterConcrete(&MsgRevokeGroupAllowance{}, "desmos/MsgRevokeGroupAllowance", nil)
+
+	cdc.RegisterInterface((*GrantTarget)(nil), nil)
+	cdc.RegisterConcrete(&UserTarget{}, "desmos/subspaces/UserTarget", nil)
```
dadamu, line 38: In order to avoid the name conflict with UserTarget in `x/reports`, I added the modules namespace.

dadamu:
@RiccardoM Updated the implementation of generic `Grant`.

x/subspaces/ante/fee.go
RiccardoM: I see, that's right. 

What if, instead of allowing any account to be the threasury, we generate the treasury account using the `authtypes.NewModuleAddress` method? This should allow to generate a unique address based on the subspace id theoretically. We then just need to delete the ability to later edit the `treasury` field of a subspace, and we would be sure that account is 100% controlled by the subspace. 

Then, to allow spending money from the treasury, we implement a method that allows the subspace *owner* to issue grants on behalf of the treasury to perform various operations (i.e. `SpendAuthorization`).

x/subspaces/types/codec.go
RiccardoM: I think it's better to rename this `UserGrantee` probably, so we can avoid any name problems. We then rename the field to be `grantee` instead of `target`

x/subspaces/ante/fee.go
dadamu: Yeah, it would be the best solution. In conclusion, we need a new treasury account which has the following features:
1. A unique address generated by subspace id
2. To allow granted manager(s) have permission to spend it

I think we should pause this subspace grant ADR implementation then creating a new ADR for the new treasury account.

dadamu: I made a PR to work with the new feature. https://github.com/desmos-labs/desmos/pull/1062

RiccardoM: Completely agree with starting another ADR to work on a better treasury address, and then coming back to this. I would mark this as _Draft_ in the meanwhile

dadamu:
This PR has too many conflicts to current master branch, so moving to #1071

PR #1091, opened 2023-03-10
chore(tests): improve reactions tests using go mock

dadamu:
## Description

Closes: #XXXX
This PR improves the x/reactions tests using gomock.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/posts/keeper/polls.go
```diff
@@ -113,9 +135,15 @@ func (k Keeper) GetUserAnswer(ctx sdk.Context, subspaceID uint64, postID uint64,
...
+	if !k.HasUserAnswer(ctx, subspaceID, postID, pollID, user) {
+		return
+	}
+
 	store.Delete(types.PollAnswerStoreKey(subspaceID, postID, pollID, user))
 
-	k.AfterPollAnswerDeleted(ctx, subspaceID, postID, pollID, user)
+	if !isTally {
+		k.AfterPollAnswerDeleted(ctx, subspaceID, postID, pollID, user)
+	}
```
dadamu, line 148: AfterPollAnswerDeleted should not be called if the deletion is from tally operation.

x/posts/keeper/polls.go
```diff
@@ -62,6 +63,27 @@ func (k Keeper) Tally(ctx sdk.Context, subspaceID uint64, postID uint64, pollID
...
+	if err != nil {
+		panic(err)
+	}
+	poll.Content = contentAny
+
+	k.SaveAttachment(ctx, poll)
+	k.RemoveFromActivePollQueue(ctx, poll)
+
+	k.AfterPollVotingPeriodEnded(ctx, poll.SubspaceID, poll.PostID, poll.ID)
+}
```
dadamu, line 85: To allow to test `AfterPollVotingPeriodEnded` inside the keeper, it is moved here from abci.

x/posts/keeper/hooks_test.go
```diff
@@ -0,0 +1,350 @@
...
+				tc.store(ctx)
+			}
+
+			hooks := newMockHooks()
+			suite.k.SetHooks(types.NewMultiPostsHooks(hooks))
+
+			suite.k.EndPoll(ctx, tc.poll)
+
+			suite.Require().True(hooks.CalledMap["AfterPollVotingPeriodEnded"])
+			suite.Require().False(hooks.CalledMap["AfterPollAnswerDeleted"])
```
dadamu, line 347: `AfterPollVotingPeriodEnded` should not call `AfterPollAnswerDeleted`.

x/posts/abci_test.go
```diff
@@ -56,11 +43,8 @@ func TestEndBlocker(t *testing.T) {
 	ctx := sdk.NewContext(ms, tmproto.Header{ChainID: "test-chain"}, false, log.NewNopLogger())
 	cdc, legacyAmino := app.MakeCodecs()
 	pk := paramskeeper.NewKeeper(cdc, legacyAmino, keys[paramstypes.StoreKey], tKeys[paramstypes.TStoreKey])
-	sk := subspaceskeeper.NewKeeper(cdc, keys[subspacestypes.StoreKey], nil, nil)
-	rk := relationshipskeeper.NewKeeper(cdc, keys[relationshipstypes.StoreKey], sk)
-	authKeeper := authkeeper.NewAccountKeeper(cdc, keys[authtypes.StoreKey], pk.Subspace(authtypes.ModuleName), authtypes.ProtoBaseAccount, app.GetMaccPerms())
-	profilesKeeper := profileskeeper.NewKeeper(cdc, legacyAmino, keys[profilestypes.StoreKey], pk.Subspace(profilestypes.DefaultParamsSpace), authKeeper, rk, nil, nil, nil)
-	keeper := postskeeper.NewKeeper(cdc, keys[poststypes.StoreKey], pk.Subspace(types.DefaultParamsSpace), profilesKeeper, sk, rk)
+
+	keeper := postskeeper.NewKeeper(cdc, keys[poststypes.StoreKey], pk.Subspace(types.DefaultParamsSpace), nil, nil, nil)
```
dadamu, line 47: Removed since abci test does not depend on these module.

x/reactions/keeper/alias_functions.go
```diff
@@ -172,7 +172,7 @@ func (k Keeper) IteratePostReactions(ctx sdk.Context, subspaceID uint64, postID
 func (k Keeper) HasReacted(ctx sdk.Context, subspaceID uint64, postID uint64, user string, value types.ReactionValue) bool {
 	found := false
 	k.IteratePostReactions(ctx, subspaceID, postID, func(reaction types.Reaction) (stop bool) {
-		if reaction.Author == user && reaction.Value.GetCachedValue().(types.ReactionValue) == value {
+		if reaction.Author == user && reaction.Value.GetCachedValue().(types.ReactionValue).Equal(value) {
```
dadamu, line 175: Using "==" always returns false here since `types.ReactionValue` is a interface

x/reactions/keeper/alias_functions_test.go
```diff
@@ -116,7 +116,24 @@ func (suite *KeeperTestSuite) TestKeeper_HasReacted() {
 			postID:     1,
 			user:       "cosmos1efa8l9h4p6hmkps6vk8lu7nxydr46npr8qtg5f",
 			value:      types.NewRegisteredReactionValue(1),
-			expResult:  false,
+			expResult:  true,
```
dadamu, line 119: Fix to true

x/reactions/keeper/alias_functions.go
RiccardoM: Does this fix mean we might have double reactions stored on chain? Should be create a migration to make sure those are deleted if that's the case?

dadamu: It is, added the migration scripts and its test.

manu0466 (approved):
LGTM!

RiccardoM:
@dadamu Would you mind fixing the conflicts that are now present please?

dadamu:
@RiccardoM Fixed

x/posts/keeper/polls.go
RiccardoM: Why was this decided? I believe we can call it safely in every occasion: if we are removing a user answer from the store, we should notify it anyway. It will be the listener's job to decide whether to handle it or not. And since the tally emits a custom event, it should be easy enough to distinguish the two cases if the listener wants

x/reactions/legacy/v3/store.go
```diff
@@ -0,0 +1,54 @@
...
+		var reaction types.Reaction
+		err := cdc.Unmarshal(iter.Value(), &reaction)
+		if err != nil {
+			return err
+		}
+		reactions = append(reactions, reaction)
+
+		if containsDuplicatedReaction(reactions, reaction) {
+			duplicatedReactions = append(duplicatedReactions, reaction)
+		}
```
RiccardoM, line 32: I believe this can be done in another way that might make the code more readable: 

```go
if contains(reactions, reaction) {
  duplicatedReactions = append(duplicatedReactions, reaction)
}

reactions = append(reactions, reacion)
```

What do you think?

x/posts/keeper/polls.go
RiccardoM: Why was this decided? I believe we can call it safely in every occasion: if we are removing a user answer from the store, we should notify it anyway. It will be the listener's job to decide whether to handle it or not. And since the tally emits a custom event, it should be easy enough to distinguish the two cases if the listener wants

x/posts/keeper/polls.go
RiccardoM: Why was this decided? I believe we can call it safely in every occasion: if we are removing a user answer from the store, we should notify it anyway. It will be the listener's job to decide whether to handle it or not. And since the tally emits a custom event, it should be easy enough to distinguish the two cases if the listener wants

x/reactions/legacy/v3/store.go
dadamu: Agree with it, improved

x/posts/keeper/polls.go
RiccardoM: @dadamu I only have this comment left. Once this is solved, we can merge this

dadamu: From my side, we can avoid from calling redundant `AfterPollAnswerDeleted` hooks since `AfterPollVotingPeriodEnded` handles the everything.

For example, when calling `AfterPollAnswerDeleted`, the listener can remove a user answer, like `DELETE FROM user_answer WHERE user_address = "cosmos...."`. In the other hand, `AfterPollVotingPeriodEnded` can remove all the answers with `DELETE FROM user_answer WHERE poll_id = 1`.

Additionally, it is better performance if the listener module wants to handle the tallied poll only and no executions with answers.

dadamu: The listener module handling the tallied poll only and no execution with answers in the current way will be like:
```go
func (h Hooks) AfterPollAnswerDeleted(sdk.Context, subspaceID uint64, postID uint64, pollID uint32, user string) {
    if !IsTalliedPoll(ctx, subspaceID, postID, pollID) {
        handleAnswer(ctx, subspaceID, postID, pollID)
    }
}
```
We need to check the poll is tallied or not in every answer deleted action, which costs O(N) redundantly where N = the amount of answers in the poll.

dadamu: @RiccardoM `Tally` function should not call `AfterPollAnswerDeleted` from my side. The main thing is to avoid passing `isTally` to `DeleteUserAnswer`, right?
If so, I think the best way is to have a new function `TallyUserAnswer`, instead of calling `DeleteUserAnswer`.

RiccardoM: @dadamu I'm honestly not convinced of this change to remove the call of `AfterPollAnswerDeleted` from `DeleteUserAnswer`. I think it's not logically correct: we are removing a user answer, so why not emit the event? Also, I think this is out of scope for this PR. I would like to ask you to revert this change and open another PR where we can have a discussion about this without blocking this one (that should only be related to tests and not breaking changes - like this one)

dadamu: Agree with you, open another PR. Reveted.

RiccardoM (approved):

PR #1117, opened 2023-04-20
docs(adr): ADR-19: Subspace post migration

dadamu:
## Description

This PR represents the ADR of a new feature which will allow users to move their own posts between subspaces.

[DCD-323](https://forbole.atlassian.net/browse/DCD-323)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

[DCD-323]: https://forbole.atlassian.net/browse/DCD-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/architecture/adr-019-subspace-post-moving.md
+++ b/docs/architecture/adr-019-subspace-post-moving.md
@@ -3,6 +3,7 @@
 ## Changelog
 - April 20th, 2023: First draft;
 - April 24th, 2023: First review;
+- May 2th, 2023: Second review;
 
 ## Status
 
@@ -18,9 +19,16 @@ Currently, when a user creates a post, they are assigned the author role and bec
 
 ## Decision
 
-We will implement a new handler that allows users to move their own posts through a message. Here are the upcoming changes:
+In order to move the post properly, we will implement an operation that involves the following functions:
+
+1. delete all reactions, reports and references that are incompatible with the new subspace of the post
+2. reset the moved post `PostReferences` to be __empty__
+3. reset the moved post `Conversation` ID to be __0__
+4. set the proper `PostID` within target subspace
 
 ### `Msg` Service
+
+We will implement a new handler that allows users to move their own posts through a message.
 
 ```proto
 service Msg {
--- a/docs/architecture/adr-019-subspace-post-moving.md
+++ b/docs/architecture/adr-019-subspace-post-moving.md
@@ -1,4 +1,4 @@
-# ADR 019: Post moving
+# ADR 019: Subspace post migration
 
 ## Changelog
 - April 20th, 2023: First draft;
@@ -9,7 +9,7 @@ Proposed
 
 ## Abstract
 
-This ADR introduces a new feature that enables post owners to move their own content to a different subspaces.
+This ADR introduces a new feature that enables post owners to move their own content between subspaces.
 
 ## Context
 
@@ -21,11 +21,9 @@ This process is overly complicated and time-consuming because it requires the us
 
 ## Decision
 
-We will implement a new handler that allows users to move their own posts through a message.
+We will implement a new handler that allows users to move their own posts through a message. Here are the upcoming changes:
 
-## `Msg` Service
-
-In order to enable users to move their own posts, we will need to implement a new message. 
+### `Msg` Service
 
 ```proto
 service Msg {
--- a/docs/architecture/adr-019-subspace-post-moving.md
+++ b/docs/architecture/adr-019-subspace-post-moving.md
@@ -2,6 +2,7 @@
 
 ## Changelog
 - April 20th, 2023: First draft;
+- April 24th, 2023: First review;
 
 ## Status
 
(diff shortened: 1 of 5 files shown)


docs/architecture/adr-019-subspace-post-moving.md
```diff
@@ -0,0 +1,77 @@
...
+
+This ADR introduces a new feature that enables post owners to move their own content between subspaces.
+
+## Context
+
+Currently, the only way to move a post to another subspace is by following process:
+1. Create an identical post as the original post in the target subspace
+2. Delete the original post on source subspace
+
+This process is overly complicated and time-consuming because it requires the user to perform two steps and manually recreate the post in the new location. Additionally, it can lead to inconsistencies if the user forgets to include certain details. A more user-friendly process would be beneficial for improving efficiency and reducing the risk of inconsistencies.
```
RiccardoM, line 20: I think that, as the context of this decision, we can instead insert more detailed information about why we want this feature in the first place. We idea of the _Context_ section is to make sure that any developer reading this document can understand why we the ADR was created in the first place. I don't think that only because something is _complex_ to do, then it's worth implementing a feature for it. 

Instead, we could use the following motivation as the context: 

> Currently, when a user creates a post their address is assigned the `author` role and they become the only user that can later edit or manage such post. Even though this currently works, it does not support all use cases. For example, currently we do not allow users to transfer the ownership of such post to another user. This might be particularly useful in contexts where an employee of a company might want to draft a post and later transfer the ownership of such post to their company.

docs/architecture/adr-019-subspace-post-moving.md
```diff
@@ -0,0 +1,77 @@
...
+## Decision
+
+We will implement a new handler that allows users to move their own posts through a message. Here are the upcoming changes:
+
+### `Msg` Service
+
+```proto
+service Msg {
+    // MovePost allows users to move their own posts to another subspace
+    rpc MovePost(MsgMovePost) returns (MsgMovePostResponse);
```
RiccardoM, line 31: Should we use a new `MsgMovePost` message, or should we instead use the existing `MsgEditPost`? I think there are pro/cons for each one but I'm not sure if there's a clear winner on this matter. What do you think?

docs/architecture/adr-019-subspace-post-moving.md
```diff
@@ -0,0 +1,77 @@
...
+
+Currently, the only way to move a post to another subspace is by following process:
+1. Create an identical post as the original post in the target subspace
+2. Delete the original post on source subspace
+
+This process is overly complicated and time-consuming because it requires the user to perform two steps and manually recreate the post in the new location. Additionally, it can lead to inconsistencies if the user forgets to include certain details. A more user-friendly process would be beneficial for improving efficiency and reducing the risk of inconsistencies.
+
+## Decision
+
+We will implement a new handler that allows users to move their own posts through a message. Here are the upcoming changes:
```
RiccardoM, line 24: There is one thing we need to consider: implementing this new message, when moving a post to another subspace, should we 
1. migrate all the reports, reactions, and everything else that references that post as well? OR
2. should we just delete everything in the original subspace instead?

docs/architecture/adr-019-subspace-post-moving.md
dadamu: I think defining a new message can have the benefit as follows:
1. User can avoid from migrating post to an another subspace accidentally when editing a post, and vice versa
2. `MsgEditPost` related methods inside desmos-bindings, desmjs can stay away from breaking change
3. Tracker can easily track the migration history of a post

As result, I decided to define a new message.

docs/architecture/adr-019-subspace-post-moving.md
dadamu: We should deleted everything, except for reference posts inside the original subspace since:
1. Supported reactions is not compatible between subspaces, for instance, `A` subspace registered reaction id 1 can be different from `B` subspace one.
2. Supported report reasons is not compatible between subspaces, for instance, `A` subspace supported reason id 1 can be different from `B` reason id

Also, tracking a post reactions and reports history can be done with finding `MsgMovePost` event to get the original post id, then check all the actions related to the post.

Therefore, I think it would be no problem to delete all the references except for posts.

dadamu: Updated to ADR

docs/architecture/adr-019-subspace-post-moving.md
dadamu: Updated to ADR

docs/architecture/adr-019-subspace-post-moving.md
RiccardoM: Should we move comments and everything else, when moving a post? Because if yes, then it would mean that the owner of a post owns also all of its comments, which might not be want we want. 

If I comment a post inside Scripta, and then the owner moves it to anohter subspace, I don't want that person to control my comment and move it to the subspace as well. It's my content and I should always be in control. If I created it inside Scripta, I would prefer it being deleted instead of migrated without my consent. 

I think we should only migrate the post and delete all the references to it.

dadamu: Agree with you, we should remain the references (reactions, reports, and etc.), except for reference posts (reply, reference). The moved target will be only the post controlled by the mover.

RiccardoM: @dadamu Actually, I think we should *delete* everything from the origin subspace. 

Suppose we are moving a post from `Subspace A` to `Subspace B`. Then we should:

1. delete all reactions and reports that reference that post, within `Subspace A`
2. reset the post `PostReferences` field to be empty 
3. reset the post `ConversationID` to be `0` 
4. set the proper `PostID` within `Subspace B`

In short, after moving the post to `Subspace B` it should appears as if it was a post that was just created. Nothing referencing it should stay inside `Subspace A`, and the post should not have any other posts' references inside it.

dadamu: @RiccardoM My bad, I made a typo that should be __remove__ instead of __remain__, I wanted to say that we should remove all the references, and only remains the comments and posts which refer to the moved post, then delete all the incompatible stuff. 😅

Totally agree with you, I will update ADR tomorrow.

dadamu: @RiccardoM Updated

RiccardoM (approved):

PR #1119, opened 2023-04-24
docs(adr): ADR-21: Subspace specific custom fee tokens

dadamu:
## Description

Closes: #XXXX
This PR introduces the subspace specific custom fee tokens by the ADR.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
+++ b/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
@@ -53,7 +53,7 @@ We will implement the operation that allows subspace admins to update the list o
 
 ```proto
 service Msg {
-  // UpdateSubspaceFeeTokens allows subspace admins to update the allowed tokens to be fee tokens inside the subspace by governance
+  // UpdateSubspaceFeeTokens allows subspace admins to update the list of fee tokens inside the subspace via a governance proposal
   rpc UpdateSubspaceFeeTokens(MsgUpdateSubspaceFeeTokens) returns (MsgUpdateSubspaceFeeTokensResponse);
 }
 
@@ -95,8 +95,7 @@ For this reason, we have decided to remove the `x/fees` module from Desmos.
 
 ### Backwards Compatibility
 
-The solution outlined above is **not** backwards compatible and will require a migration script to update all existing subspaces to the new version. This script will handle the following tasks:
-- migrate all subspaces to have a default allowed fee tokens list.
+The solution outlined above is **not** backward compatible, and it requires a upgrade to delete `x/fees` stores.
 
 ### Positive
 
--- a/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
+++ b/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
@@ -7,7 +7,7 @@
 
 ## Status
 
-Proposed
+ACCEPTED Not Implemented
 
 ## Abstract
 
--- a/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
+++ b/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
@@ -3,6 +3,7 @@
 ## Changelog
 
 - April 24th, 2023: First draft;
+- May 8th, 2023: First review;
 
 ## Status
 
@@ -24,25 +25,14 @@ To address the issue mentioned above, we propose implementing a mechanism that a
 
 Ideally, the flow that will lead one subspace to accept an additional fee token is going to be the following: 
 
-1. The subspace owner creates an on-chain governance proposal asking validators if they are fine in receiving fees paid in the new token denomination with minimum gas price.
+1. The subspace owner creates an on-chain governance proposal asking validators if they are fine in receiving fees paid in the new token denomination when validating transactions related to that subspace.
 2. Validators will agree or reject the proposal through on-chain voting. 
-3. If the proposal is accepted, the new token denom with minimum gas price will be added to the list of accepted  minimum gas prices that can be used to pay for fees for content-related transactions, then subspace managers can add the token denom to allowed fee token list.
-4. If the proposal is rejected, the new token denom will not be added to the list of accepted minimum gas prices that can be used to pay for fees for content-related transactions.
-
+3. If the proposal is accepted, the new token denom will be added to the list of additional fee token denoms that can be used to pay for fees within that subspace.
+4. If the proposal is rejected, the new token denom will not be added to the list of additional fee token denoms.
 
 ### Types
 
-We will define a new `Params` type to show the `x/subspaces` parameters.
-
-```proto
-// Params contains the module parameters
-message Params {
-  // List of the minimum gas prices to be accepted by validators
-  repeated Coin min_gas_prices = 2;
-}
-```
-
-In order to represent the list of fee token denoms that are supported by a subspace, we will add a new field to the current `Subspace` structure.
+To represent the list of supported fee token denoms and their default minimum prices within a subspace, we will add a new field `allowed_fee_tokens` to the existing Subspace structure.
 
 ```proto
 message Subspace {
@@ -52,82 +42,51 @@ message Subspace {
   // the creation time of the subspace
   google.protobuf.Timestamp creation_time = 7;
   
-  // List of fee token denoms allowed inside the subspace
-  repeated string allowed_fee_tokens = 8;
+  // List of fee token denoms with default minimum gas prices allowed inside the subspace
+  repeated Coin allowed_fee_tokens = 8;
 }
 ```
 
 ### Msgs
 
-We will allow the the following operations:
-1. update subspace allowed fee tokens list by managers;
-2. update `x/subspaces` parameters by governance.
+We will implement the operation that allows subspace admins to updates subspace allowed fee tokens list by governance;
 
 ```proto
 service Msg {
-  // UpdateSubspaceFeeTokens allows subspace managers to update the allowed tokens to be fee tokens inside the subspace
+  // UpdateSubspaceFeeTokens allows subspace admins to update the allowed tokens to be fee tokens inside the subspace by governance
   rpc UpdateSubspaceFeeTokens(MsgUpdateSubspaceFeeTokens) returns (MsgUpdateSubspaceFeeTokensResponse);
-    
-  // UpdateParams defines a (governance) operation for updating the module
-  rpc UpdateParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
 }
 
-// MsgUpdateSubspaceFeeTokens represents the message to be used to update a subspace fee tokens
+// MsgUpdateSubspaceFeeTokens represents the message to be used to update a subspace fee tokens by governance
 message MsgUpdateSubspaceFeeTokens {
 
   // Id of the subspace where the list of allowed fee tokens will be updated
   uint64 subspace_id = 1;
     
-  // List of the allowed tokens to be fee token inside the subspace
-  repeated string allowed_fee_tokens = 2;
+  // List of the allowed tokens to be fee token inside the subspace along with their default minimum prices
+  repeated Coin allowed_fee_tokens = 2;
     
-  // Address of the sender/signer
-  string signer = 3;
+  // authority is the address that controls the module (defaults to x/gov unless overwritten).
+  string authority = 3;
 }
 
 // MsgUpdateSubspaceFeeTokensResponse represents the Msg/UpdateSubspaceFeeTokens response type
 message MsgUpdateSubspaceFeeTokensResponse {}
-
-
-// MsgUpdateParams is the Msg/UpdateParams request type.
-message MsgUpdateParams {
-  string authority = 1;
-  Params params = 2;
-}
-
-// MsgUpdateParamsResponse defines the response structure for executing a
-// MsgUpdateParams message.
-message MsgUpdateParamsResponse {}
-```
-
-### Query
-
-We will also provide a new query endpoint that enables users to check `x/subspaces` parameters.
-
-```proto
-service Query {
-  // Params allows to query the module parameters
-  rpc Params(QueryParamsRequest) returns (QueryParamsResponses) {
-    option (google.api.http).get = "/desmos/subspaces/v3/params";
-  };
-}
-
-// QueryParamsRequest is the request type for Query/Params RPC method
-message QueryParamsRequest {}
-
-// QueryParamsResponse is the response type for Query/Params RPC method
-message QueryParamsResponse {
-  Params params = 1;
-}
 ```
 
 ### Custom TxFeeChecker
 
-We will build a new `TxFeeChecker` based on the existing one that acts as follows:
-1. check if the provided fee tokens are included in the list of subspace allowed fee tokens;
-2. perform the following operation:
-  1. if the provide fee tokens are all included subspace allowed fee tokens list, run custom `x/subspaces` TxFeeChecker;
-  2. otherwise, run the default fee checker `CheckTxFeeWithValidatorMinGasPrices`.
+To make it easier for validators to manage the minimum prices of fee tokens allowed within a subspace, we will develop a new `TxFeeChecker` based on the existing one, which will function as follows:
+- Combine the list of minimum gas prices in the validator's local configuration with the list of allowed fee tokens and their minimum prices within the subspace.
+- Follow the same process as the existing `TxFeeChecker`.
+
+### Update `x/fees`
+
+Since `x/fees` module may conflict with the custom fee tokens feature, as demonstrated in the following scenario:
+- Governance decides to change the minimum fees of MsgCreatePost to `10dsm` using `x/fees`.
+- Governance accepts the custom token `minttoken` as a fee within subspace 1 using `MsgUpdateSubspaceFeeTokens`.
+
+Due to this conflict, `minttoken` will not be accepted as a fee within the subspace 1. Therefore, we decided to update the `CheckFees` logic of `x/fees` to ensure compatibility with this feature.
 
 ## Consequences
 
--- a/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
+++ b/docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
@@ -10,7 +10,7 @@ Proposed
 
 ## Abstract
 
-This ADR introduces a mechanism that allows users to pay fees within a designated subspace using their existing service tokens.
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+- April 24th, 2023: First draft;
+
+## Status
+
+Proposed
+
+## Abstract
+
+This ADR introduces a mechanism that allows users to pay fees within a designated subspace using their existing service tokens.
```
RiccardoM, line 13:
```suggestion
This ADR enables subspace owners to let content creators pay for transaction fees using different token denominations when creating contents within their subspace.
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Context
+
+Desmos uses subspaces to represent virtual applications built on top of our protocol. Currently, subspaces are mainly used to store content such as posts, reports, and reactions. While this is sufficient for developing DApps on Desmos, it may not be enough to convince existing applications to migrate to our platform since Desmos is lack of a mechanism that enables users to pay fees with existing service tokens.
+
+For instance, consider Keplr's on-chain governance discussions feature, which is currently implemented using a custom backend that stores data in a centralized way. Although Keplr could easily use Desmos as their backend, they are not willing to do so because it offers no advantage to them. Migrating to Desmos would simply transfer their existing users to our chain without any benefits for them.
```
RiccardoM, line 19:
```suggestion
Desmos uses subspaces to represent virtual applications built on top of our protocol. Currently, subspaces are mainly used to store content such as posts, reports, and reactions. While these features are sufficient to develop new applications on top of Desmos, they may not be enough to convince existing applications to migrate to our platform.

For instance, let's consider an exiting application that uses a centralized database in order to store users discussions. Although such application might easily use Desmos as its backend, they most likely are not going to do so because it offers no advantage to them. Migrating to Desmos would simply transfer their existing users to our chain without any benefits for them.
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Decision
+
+To address the issue mentioned above, we propose implementing a mechanism that allows users to pay fees using the existing service tokens within the specified subspace. The mechanism must have the following features:
+- Defining a minimum gas prices for non-DSM tokens that validators are willing to accept via governance proposal;
+- Allowing the subspace manager(s) to update the list of fee tokens allowed in the subspace;
+- Checking the on-chain minimum gas price requirement for non-DSM fees using a custom `TxChecker`.
```
RiccardoM, line 26:
```suggestion
To address the issue mentioned above, we propose implementing a mechanism that allows users to pay content-related fees using additional token denominations within the specified subspace. 

Ideally, the flow that will lead one subspace to accept an additional fee token is going to be the following: 

1. The subspace owner creates an on-chain governance proposal asking validators if they are fine in receiving fees paid in the new token denomination when validating transactions related to that subspace.
2. Validators will agree or reject the proposal through on-chain voting. 
3. If the proposal is accepted, the new token denom will be added to the list of additional fee token denoms that can be used to pay for fees within that subspace.
4. If the proposal is rejected, the new token denom will not be added to the list of additional fee token denoms.
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+The upcoming changes are as follows:
```
RiccardoM, line 28:
```suggestion
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+We will define a new `Params` type to show the `x/subspaces` parameters.
+
+```proto
+// Params contains the module parameters
+message Params {
+  // List of the minimum gas prices to be accepted by validators
+  repeated Coin min_gas_prices = 2;
+}
+```
```
RiccardoM, line 40: Why is this needed? Each subspace will have its own `allowed_fee_tokens` list anyway. There's no need to have a module-wide list

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+In addition, we will add a __allowed_fee_tokens__ field to current `Subspace` structure.
```
RiccardoM, line 42:
```suggestion
In order to represent the list of fee token denoms that are supported by a subspace, we will add a new field to the current `Subspace` structure.
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+```proto
+message Subspace {
+  // Unique id that identifies the subspace
+  uint64 id = 1;
+
+  // Human-readable name of the subspace
+  string name = 2 [ (gogoproto.moretags) = "yaml:\"name\"" ];
```
RiccardoM, line 50: No need to put theses. Thye can be included inside the `...skip` part

```suggestion
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+
+  ...skip
+
+  // the creation time of the subspace
+  google.protobuf.Timestamp creation_time = 7;
+  
+  // List of allowed fee tokens inside the subspace
```
RiccardoM, line 57:
```suggestion
  // List of fee token denoms allowed inside the subspace
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,152 @@
...
+  repeated string allowed_fee_tokens = 8;
+}
+```
+
+### Msgs
+
+We will allow the the following operations:
+1. update subspace allowed fee tokens list by managers;
+2. update `x/subspaces` parameters by governance.
```
RiccardoM, line 66: Why is this needed? It would be much simpler to require subspace owners to go through a governance-gated update of the subspace `allowed_fee_tokens` list instead of having two different message types for this. Just: 

1. The subspace owner creates an on-chain proposal to update the list of tokens 
2. The governance votes yes/no 
3. Based on the result, the list is updated or not.

No need for multiple message types or anything like that

RiccardoM (commented):
I have some doubts on the proposed implementation. I think there's no need for multiple messages and steps. Everything could be simplified to use on-chain proposals instead.

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
dadamu: Splitting the message into two parts would make it easier for validators to manage min-gas-prices. 

For example, let's say Subspace A and B both allow a min-gas-price of 0.1 minttoken. If validators want to update the mini-gas-price for minttoken to 0.2 minttoken after a market crash, they would need to perform the following operations:
1. Propose a new min-gas-price of 0.2 minttoken for Subspace A
2. Propose a new min-gas-price of 0.2 minttoken for Subspace B
3. Propose multiple proposals to update the min-gas-price for all subspaces that allow minttoken

However, this approach can make it difficult for validators to manage min-gas-prices. To address this, I decided to create a separate list for minimum gas prices.

RiccardoM: @dadamu This is extremely complicated to understand and wrap my head around. I honestly cannot understand: 

1. why there's the need to have two different messages.  
   If both subspace managers *and* any other user must use *the same* governance proposal to update the subspace fee tokens, why do we need to have two different messages? 
2. why can't we use an implementation of the `Content` interface, rather than custom messages?  
   Why can't we have something as simple as other governance proposal types such as `SpendCommunityProposal` but instead we need to have this custom message? 


I honestly don't think it's necessary to have this extremely complicated mechanism with two different messages types. Instead, we can simply use a single `Content` implementation type and rely on that

dadamu: @RiccardoM `Content` interface is gov v1beta1 feature, it will be deprecated in the future. Now, cosmos-sdk applies proposal message to have spend community proposal. See:
https://github.com/cosmos/cosmos-sdk/blob/v0.47.2/proto/cosmos/distribution/v1beta1/tx.proto#L155

RiccardoM: @dadamu Ok, I see what you mean about the `Content` interface. 

For the other discussion, about two messages, it makes little to no sense to have such an incredible complex procedure. Instead, the procedure should be as follow: 

1. The subspace admin creates a governance proposal asking validators if it's OK for them to accept fees paid in the `TokenA` inside their subspace.
2. If the proposal passes, the list of fee tokens **only within that subspace** is updated. 

This guarantees that: 

1. Subspace admins **always** have to ask the governance its permission to accept a new token 
2. The governance can agree to accept `TokenA` as a fee token inside one subspace, but reject it for another subspace (for whatever reason)
3. There is a lower chance of introducing bugs: with a simplified procedure and less operations, fewer things can go wrong

To do this, we can simply implement a `MsgUpdateSubspaceFeeTokens` that accepts: 
- a `SubspaceID`
- a list of coin denominations 

It should be executed only by the governance module, making it de-facto as much similar as the old `Content` interface.

dadamu: @RiccardoM I see what you want to do is reduce the potential bugs, while this procedure is lack of the minimum gas prices management from validators since only Subspace `admin` can create a proposal to modify the minimum gas prices of subspaces. I believe validators would like to fix the minimum gas prices after market crash.

For example, multiple subspaces passed proposals to use stable coin `USDM`. `USDM` depegs and become 0.001USD after market crash. Then validator must ask all the subspaces managers to modify their own fee token list to remove `USDM`, or validators will keep receiving valueless `USDM` as fees.

Can you give me a sample that as a validator how to modify the minimum prices of `USDM` inside each subspace in a one proposal?

RiccardoM: @dadamu If validators want to modify the minimum gas prices they accept, they should be able to do it by editing the `min-gas-price` inside their node configuration. We could simply read that file and if the value written there is less than the value inside the subspace, we should make sure the transaction is not added to the validator's mem pool. 

This way: 
- subspace admins are the only ones that can touch the fee denoms accepted within their subspace;
- if validators want to independently update their min gas prices, they can do that without a problem;
- if no validator sets a min-gas-price for a specific fee denom, then the default one for that subspace will be used.

Do you think that's technically possible?

dadamu: Yes, it is doable and solves my concerns. I will update ADR to include everything we discussed.

dadamu: @RiccardoM Updated.

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -52,82 +42,51 @@ message Subspace {
...
+To make it easier for validators to manage the minimum prices of fee tokens allowed within a subspace, we will develop a new `TxFeeChecker` based on the existing one, which will function as follows:
+- Combine the list of minimum gas prices in the validator's local configuration with the list of allowed fee tokens and their minimum prices within the subspace.
+- Follow the same process as the existing `TxFeeChecker`.
 
-// QueryParamsResponse is the response type for Query/Params RPC method
-message QueryParamsResponse {
-  Params params = 1;
-}
-```
+### Update `x/fees`
```
dadamu, line 83: Since `x/fees` may conflict with subspace fee token checks, I think we can update the logic of `CheckFees`. What do you think? @RiccardoM

dadamu, line 83: Since x/fees may conflict with subspace fee token checks, I think we can update the logic of CheckFees. What do you think? @RiccardoM @manu0466

RiccardoM: I think we should completely drop the `x/fees` module instead. It's never been used anyway

dadamu: @RiccardoM Updated docs.

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,109 @@
+# ADR 021: Subspace specific custom fee tokens
+
+## Changelog
+
+- April 24th, 2023: First draft;
+- May 8th, 2023: First review;
+
+## Status
+
+Proposed
```
RiccardoM, line 10:
```suggestion
ACCEPTED Not Implemented
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+  google.protobuf.Timestamp creation_time = 7;
+  
+  // List of fee token denoms with default minimum gas prices allowed inside the subspace
+  repeated Coin allowed_fee_tokens = 8;
+}
+```
+
+### Msgs
+
+We will implement the operation that allows subspace admins to updates subspace allowed fee tokens list by governance;
```
RiccardoM, line 52:
```suggestion
We will implement the operation that allows subspace admins to update the list of allowed fee tokens list within their subspace via a governance proposal:
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+
+```proto
+service Msg {
+  // UpdateSubspaceFeeTokens allows subspace admins to update the allowed tokens to be fee tokens inside the subspace by governance
```
RiccardoM, line 56:
```suggestion
  // UpdateSubspaceFeeTokens allows subspace admins to update the list of fee tokens inside the subspace via a  governance proposal
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+- Combine the list of minimum gas prices in the validator's local configuration with the list of allowed fee tokens and their minimum prices within the subspace.
+- Follow the same process as the existing `TxFeeChecker`.
+
+### Deprecated `x/fees`
+
+Since `x/fees` module may conflict with the custom fee tokens feature, as demonstrated in the following scenario:
+- Governance decides to change the minimum fees of MsgCreatePost to `10dsm` using `x/fees`.
+- Governance accepts the custom token `minttoken` as a fee within subspace 1 using `MsgUpdateSubspaceFeeTokens`.
+
+Due to this conflict, `minttoken` will not be accepted as a fee within the subspace 1. Therefore, we decided to drop `x/fees` from Desmos.
```
RiccardoM, line 89:
```suggestion
By implementing the mechanism described above, there might be a problem with the `x/fees` module. Here's an example scenario to illustrate this:

1. the governance decides to use `x/fees` to change the minimum fees of `MsgCreatePost` to `10dsm`.
2. the governance also wants to use a custom token called `TokenA` as a fee token within subspace `1` using `MsgUpdateSubspaceFeeTokens`.

Because `x/fees` takes priority over anything, `TokenA` will not be accepted as a fee token within subspace `1`. 

For this reason, we have decided to remove the `x/fees` module from Desmos.
```

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Consequences
+
+### Backwards Compatibility
+
+The solution outlined above is **not** backwards compatible and will require a migration script to update all existing subspaces to the new version. This script will handle the following tasks:
+- migrate all subspaces to have a default allowed fee tokens list.
```
RiccardoM, line 96:
```suggestion
The solution outlined above is **not** backward compatible, and it requires a migration script to update all existing subspaces to the new version. The script should take care of migrating all subspaces to have a default allowed fee tokens list.
```

RiccardoM: Is this actually true though? If we mark the Proto field as [`optional`](https://protobuf.dev/programming-guides/proto3/#specifying-rules), do we need to migrate everything?

dadamu: Now proto3 fields are all optional now. You are right, adding new field does not cause the backward compatibility issues. I just learned from the official documentation. Thanks!

> https://protobuf.dev/programming-guides/proto3/#updating
If you add new fields, any messages serialized by code using your “old” message format can still be parsed by your new generated code. You should keep in mind the [default values](https://protobuf.dev/programming-guides/proto3/#default) for these elements so that new code can properly interact with messages generated by old code. Similarly, messages created by your new code can be parsed by your old code: old binaries simply ignore the new field when parsing. See the [Unknown Fields](https://protobuf.dev/programming-guides/proto3/#unknowns) section for details.

docs/architecture/adr-021-subspace-specific-custom-fee-tokens.md
```diff
@@ -0,0 +1,111 @@
...
+
+// MsgUpdateSubspaceFeeTokensResponse represents the Msg/UpdateSubspaceFeeTokens response type
+message MsgUpdateSubspaceFeeTokensResponse {}
+```
+
+### Custom TxFeeChecker
+
+To make it easier for validators to manage the minimum prices of fee tokens allowed within a subspace, we will develop a new `TxFeeChecker` based on the existing one, which will function as follows:
+- Combine the list of minimum gas prices in the validator's local configuration with the list of allowed fee tokens and their minimum prices within the subspace.
+- Follow the same process as the existing `TxFeeChecker`.
```
RiccardoM, line 81:
```suggestion
1. combine the list of minimum gas prices in the validator's local configuration with the list of allowed fee tokens and their minimum prices within the subspace;
2. follow the same process used inside the existing `TxFeeChecker`.
```

RiccardoM (approved):

PR #1138, opened 2023-05-10
feat!: implement ADR 21:  Subspace specific custom fee tokens

dadamu:
## Description

This PR implements ADR-21: Subspace specific custom fee tokens feature.

Closes: [DCD-312](https://desmoslabs.atlassian.net/browse/DCD-312)

~~Depends-On: #1135~~
~~Depends-On: #1139~~

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

[DCD-312]: https://forbole.atlassian.net/browse/DCD-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dadamu:
@RiccardoM @manu0466 Rebased

x/subspaces/ante/tx_checker.go
```diff
@@ -0,0 +1,48 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	antetypes "github.com/desmos-labs/desmos/v5/x/subspaces/ante/types"
+)
+
+// CheckTxFeeWithSubspaceMinPrices returns the tx checker that including the subspace allowed tokens into minimum prices list
+func CheckTxFeeWithSubspaceMinPrices(txFeeChecker ante.TxFeeChecker, sk antetypes.SubspacesKeeper) ante.TxFeeChecker {
+	return func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
+		subspaceID, isSubspaceTx := GetTxSubspaceID(tx)
+		if !isSubspaceTx {
```
dadamu, line 13: @RiccardoM
Do you think we should exclude subspace management messages ( `MsgEditSubspace`, `MsgAddUserToUserGroup` and etc.), forcing admins mange subspace by using `DSM`?

proto/desmos/subspaces/v3/msgs.proto
```diff
@@ -492,3 +500,33 @@ message MsgSetUserPermissions {
 // MsgSetUserPermissionsResponse defines the Msg/SetPermissionsResponse
 // response type
 message MsgSetUserPermissionsResponse {}
+
+// --------------------------------------------------------------------------------------------------------------------
+
+// MsgUpdateSubspaceFeeTokens represents the message to be used to update a subspace fee tokens by governance
```
RiccardoM, line 506:
```suggestion
// MsgUpdateSubspaceFeeTokens represents the message to be used to update the accepted fee tokens
// inside a given subspace, using an on-chain governance proposal
```

proto/desmos/subspaces/v3/msgs.proto
```diff
@@ -492,3 +500,33 @@ message MsgSetUserPermissions {
...
+message MsgUpdateSubspaceFeeTokens {
+  option (cosmos.msg.v1.signer) = "authority";
+
+  // Id of the subspace where the list of allowed fee tokens will be updated
+  uint64 subspace_id = 1 [
+    (gogoproto.customname) = "SubspaceID",
+    (gogoproto.moretags) = "yaml:\"subspace_id\""
+  ];
+    
+  // List of the allowed tokens to be fee token inside the subspace along with their default minimum prices
```
RiccardoM, line 518:
```suggestion
  // List of tokens allowed to be fee tokens inside the given subspace, represented as their gas prices
```

x/subspaces/ante/tx_checker.go
RiccardoM: Yes, I believe we should *not* allow those messages to be paid using non-DSM tokens.

x/subspaces/ante/tx_checker.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		subspace, found := sk.GetSubspace(ctx, subspaceID)
+		if !found {
+			return txFeeChecker(ctx, tx)
+		}
+
+		newMinPrices := MergeMinPrices(ctx.MinGasPrices(), sdk.NewDecCoinsFromCoins(subspace.AllowedFeeTokens...))
+		newCtx := ctx.WithMinGasPrices(newMinPrices)
+		return txFeeChecker(newCtx, tx)
```
RiccardoM, line 24: I'm not super familiar with on-chain gas prices, but I thought they were treated using the `and` logic. So if I have a min gas price of `0.01udsm,0.01ustar` then to properly broadcast my transaction I have to pay fees in *both* tokens, and not *either one* of the two. Do you have any link for this to be checked out that we can paste here to remember this in the future?

x/subspaces/client/cli/cli_test.go
```diff
@@ -195,7 +195,7 @@ func (s *IntegrationTestSuite) TestCmdQuerySubspace() {
 					"cosmos1s0he0z3g92zwsxdj83h0ky9w463sx7gq9mqtgn",
 					"cosmos1s0he0z3g92zwsxdj83h0ky9w463sx7gq9mqtgn",
 					time.Date(2020, 1, 1, 12, 00, 00, 000, time.UTC),
-				),
+				).SetAllowedFeeTokens(sdk.Coins{}),
```
RiccardoM, line 198: Why is this needed?

x/subspaces/keeper/msg_server.go
```diff
@@ -831,3 +832,44 @@ func (k msgServer) SetUserPermissions(goCtx context.Context, msg *types.MsgSetUs
 
 	return &types.MsgSetUserPermissionsResponse{}, nil
 }
+
+// UpdateSubspaceFeeTokens defines a rpc method for MsgUpdateSubspaceFeeTokens
+func (k msgServer) UpdateSubspaceFeeTokens(goCtx context.Context, msg *types.MsgUpdateSubspaceFeeTokens) (*types.MsgUpdateSubspaceFeeTokensResponse, error) {
+	authority := k.authority
+	if authority != msg.Authority {
+		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", authority, msg.Authority)
+	}
```
RiccardoM, line 841: No need to delcare a variable here if we're not going to use it anywhere outside this check
```suggestion
	if msg.Authority != k.authority {
		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", k.authority, msg.Authority)
	}
```

x/subspaces/keeper/msg_server_test.go
```diff
@@ -3059,3 +3060,125 @@ func (suite *KeeperTestSuite) TestMsgServer_SetUserPermissions() {
...
+			},
+			msg: types.NewMsgUpdateSubspaceFeeTokens(
+				1,
+				sdk.Coins{{Denom: "minttoken", Amount: sdk.NewInt(-10)}},
+				"cosmos1x5pjlvufs4znnhhkwe8v4tw3kz30f3lxgwza53",
+			),
+			shouldErr: true,
+		},
+		{
+			name: "subspace are updated correctly",
```
RiccardoM, line 3112:
```suggestion
			name: "subspace is updated correctly",
```

x/subspaces/types/models.go
```diff
@@ -84,9 +84,22 @@ func (sub Subspace) Validate() error {
...
+		return fmt.Errorf("invalid allowed fee tokens: %s", sub.AllowedFeeTokens)
+	}
+
 	return nil
 }
 
+// Set allowed fee tokens without validating it.
+// Before storing the updated subspace, a validation with Validate() should
+// be performed.
+func (sub Subspace) SetAllowedFeeTokens(feeTokens sdk.Coins) Subspace {
```
RiccardoM, line 98: I'm not a big fan of this approach. Instead I think it would be better to: 

1. Add another param inside the `NewSubspace` function to specify the fee tokens when creating the subspace. This is going to be useful when exporting/importing the genesis as well as inside tests. 
2. Rename this method `Update` as we have done with the `Post` and `Attachment` type, to keep everything consistent.

Also, I'm doubting whether `AllowedFeeTokens` is the proper name. I think `AdditionalFeeTokens` might be better instead. The name `AllowedFeeTokens` might make people think that if I specify `0.01ustar` then the `star` token is going to be then only one allowed, while the `udsm` should always be allowed anyway. Using `AdditionalFeeTokens` might make everything more easy to understand.

x/subspaces/ante/tx_checker.go
dadamu: They are treated as `or`,  so one of tokens reaches the min fees can broadcast properly, say, to send the message with 100 gas via the node with `0.01udsm, 0.01ustar` min gas prices, `1udsm` or `1ustar` can pass the check.

Here is the reference:
https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/auth/ante/validator_tx_fee.go#L38

x/subspaces/types/models.go
dadamu: I see, I use the `SetAllowedFeeTokens` since adding another param inside the `NewSubspace` will break hundreds of existing test codes, but yeah, it is better to have an additional field.

strongly agree with `AdditionalFeeTokens` is better.

dadamu: Finally, using `sed` to update the new field quickly:
```bash
#!/bin/bash

dir="./x"

find "$dir" -type f -name "*.go" | while read -r file; do
    sed -i -E ':a;N;$!ba;s/NewSubspace\([^)]*blockTime,/&\nnil,/g' "$file"
    echo "$file"
done
```

x/subspaces/client/cli/cli_test.go
dadamu: It is added by mistake, removed.

dadamu: It is added to avoid from returning `nil` which actually expected to be `empty` slice.
It can be solved by adding new parameter to `NewSubspace`.

x/subspaces/ante/tx_checker.go
RiccardoM: I think it's best to do the opposite: adding a `ManageSubspaceMsg` interface that identifies only the one that should *not* be included. This is faster and easier to maintain for future features: it's going to be less likely to add new management operations rather than social operations

x/subspaces/types/models.go
RiccardoM: I honestly don't know of any other tool that can easily update the code. I would have done it by hand :sweat: Great job using `sed` :100:

x/subspaces/ante/tx_checker.go
dadamu: Great idea, Implemented `ManageSubspaceMsg`, instead.

RiccardoM (approved):
Great job :100:

PR #1157, opened 2023-05-23
feat: implement ADR-019: Subspace post migration

dadamu:
## Description

This PR implements ADR-019: Subspace post migration.

Closes: [DCD-324](https://desmoslabs.atlassian.net/browse/DCD-324)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

[DCD-324]: https://forbole.atlassian.net/browse/DCD-324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,88 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+			k.InsertActivePollQueue(ctx, updatedAttachment)
+		}
+		return false
+	})
+
+	// Update the id for the next attachment
+	k.SetNextAttachmentID(ctx, msg.TargetSubspaceID, newPostID, newAttachmentID+1)
+
+	// Delete the post
+	k.Keeper.DeletePost(ctx, msg.SubspaceID, msg.PostID)
```
dadamu, line 566: `DeletePost` calls the `PostDeletedHook`, which removes all the incompatibility references managed by other modules already, so no need to implement the [`PostMovedHook`](https://github.com/desmos-labs/desmos/blob/master/docs/architecture/adr-019-subspace-post-moving.md#hooks).

proto/desmos/posts/v3/msgs.proto
```diff
@@ -41,6 +41,9 @@ service Msg {
   //
   // Since: Desmos 5.0.0
   rpc UpdateParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
+
+   // MovePost allows users to move their own posts to another subspace
```
RiccardoM, line 45: We should probably add a `Since: Desmos 5.1.0` here

proto/desmos/posts/v3/msgs.proto
```diff
@@ -303,4 +306,46 @@ message MsgUpdateParams {
 // MsgUpdateParams message.
 //
 // Since: Desmos 5.0.0
-message MsgUpdateParamsResponse {}
\ No newline at end of file
+message MsgUpdateParamsResponse {}
+
+// MsgMovePost move a post to another subspace
```
RiccardoM, line 311:
```suggestion
// MsgMovePost moves a post to another subspace
```

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,88 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+
+	// Check if the target section exists
+	if !k.HasSection(ctx, msg.TargetSubspaceID, msg.TargetSectionID) {
+		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace section with id %d not found", msg.TargetSectionID)
+	}
+
+	// Check the permission to move the post
+	canMove := post.Author == msg.Owner && k.HasPermission(ctx, msg.TargetSubspaceID, msg.TargetSectionID, msg.Owner, types.PermissionWrite)
+	if !canMove {
+		return nil, errors.Wrap(subspacestypes.ErrPermissionDenied, "you cannot move this post")
```
RiccardoM, line 525:
```suggestion
		return nil, errors.Wrap(subspacestypes.ErrPermissionDenied, "you don't have write permission on the target section")
```

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,88 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace with id %d not found", msg.TargetSubspaceID)
+	}
+
+	// Check if the target section exists
+	if !k.HasSection(ctx, msg.TargetSubspaceID, msg.TargetSectionID) {
+		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "subspace section with id %d not found", msg.TargetSectionID)
+	}
+
+	// Check the permission to move the post
+	canMove := post.Author == msg.Owner && k.HasPermission(ctx, msg.TargetSubspaceID, msg.TargetSectionID, msg.Owner, types.PermissionWrite)
```
RiccardoM, line 523: Probably the `post.Author == msg.Owner` check should be split, so that the user can know exactly what's wrong (they are not they author of the post)

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,88 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+	// Get the next post id
+	newPostID, err := k.GetNextPostID(ctx, msg.TargetSubspaceID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the post and validate it
+	updateTime := ctx.BlockTime()
+	postMove := types.NewPostMove(msg.TargetSubspaceID, msg.TargetSectionID, newPostID, updateTime)
+	updatedPost := postMove.Update(post)
```
RiccardoM, line 537: Why is the `PostMove` type defined? Why can't we simply use the `Post` type and update the fields?

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,88 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+
+	// Update the id for the next post
+	k.SetNextPostID(ctx, msg.TargetSubspaceID, updatedPost.ID+1)
+
+	// Move all post attachments
+	newAttachmentID := uint32(0)
+	k.IteratePostAttachments(ctx, post.SubspaceID, post.ID, func(attachment types.Attachment) (stop bool) {
+		newAttachmentID++
+		attachmentMove := types.NewAttachmentMove(msg.TargetSubspaceID, newPostID, newAttachmentID)
+		updatedAttachment := attachmentMove.Update(attachment)
```
RiccardoM, line 553: Again, I don't understand why there's the need to define a new `AttachmentMove` type here. Can't we simply edit the `Attachment` fields?

x/posts/keeper/msg_server.go
dadamu: I refer the [`EditPost`](https://github.com/desmos-labs/desmos/blob/master/x/posts/keeper/msg_server.go#L151) update section to have update object to update fields since I think it brings the benefit as follows:

#### 1. msg server can be focus on operating store rather than updating post fields
If we simply updates the fields in this section, the codes will make the section longer and not only be focus on store operation as follows:
```go
...checking section

// Update the post and validate it
updateTime := ctx.BlockTime()
post.SubspaceID = msg.TargetSubspaceID
post.ID = newPostID
post.Conversation = 0
post.References = nil
// ...other upcoming new fields update

err = k.ValidatePost(ctx, updatedPost)
if err != nil {
return nil, err
}

k.SavePost(ctx, post)

...other store operations
```

#### 2. Having [specific unit test](https://github.com/desmos-labs/desmos/blob/paul/DCD-324/impl-post-transfer/x/posts/types/models_test.go#L1041) on `PostMove.Update` can have another protection for fields update

`TestPostMove` tests all the possibility on the fields update of the moved post, it can exclude the missing cases inside msg server test. As a result, we can be focus on the store operation inside msg server since the unit test guarantees fields update would be not wrong.

proto/desmos/posts/v3/msgs.proto
dadamu: Maybe Desmos 6.0.0 is better since it would be Desmos v6 feature on roadmap

x/posts/client/cli/tx.go
```diff
@@ -347,3 +348,52 @@ Then, the "Cat" answer has index 0 and and the "Dog" answer has index 1.
...
+			if err != nil {
+				return err
+			}
+
+			targetSubspaceID, err := subspacestypes.ParseSubspaceID(args[2])
+			if err != nil {
+				return err
+			}
+
+			targetSectionID, err := subspacestypes.ParseSectionID(args[2])
```
RiccardoM, line 380: This should be `args[3]`
```suggestion
			targetSectionID, err := subspacestypes.ParseSectionID(args[3])
```

x/posts/client/cli/cli_test.go
```diff
@@ -879,3 +879,77 @@ func (s *IntegrationTestSuite) TestCmdAnswerPoll() {
...
+	testCases := []struct {
+		name      string
+		args      []string
+		shouldErr bool
+		respType  proto.Message
+	}{
+		{
+			name: "invalid subspace id returns error",
+			args: []string{
+				"", "1", "2", "1",
```
RiccardoM, line 894: We should add the `--from` flag to all these tests in order to make sure that they don't fail due to the invalid signer address. Otherwise we will not catch errors like the one highlighted above

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,92 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+	k.Keeper.SavePost(ctx, updatedPost)
+
+	// Update the id for the next post
+	k.SetNextPostID(ctx, msg.TargetSubspaceID, newPostID+1)
+
+	// Move all post attachments
+	newAttachmentID := uint32(0)
+	k.IteratePostAttachments(ctx, msg.SubspaceID, msg.PostID, func(attachment types.Attachment) (stop bool) {
+		newAttachmentID++
+		attachmentMove := types.NewAttachmentMove(msg.TargetSubspaceID, newPostID, newAttachmentID)
```
RiccardoM, line 556: To make this more clear, I think we should use the following notation instead:
```suggestion
		attachmentMove := types.NewAttachmentMove(updatedPost.SubspaceID, updatedPost.ID, newAttachmentID)
```

x/posts/types/models.go
```diff
@@ -228,6 +228,53 @@ func (p Post) Update(update PostUpdate) Post {
...
+// PostMove contains data related to a post that can be updated after it has been moved.
+type PostMove struct {
+	// Move's subspace id will always replace the existing ones
+	SubspaceID uint64
+
+	// Move's section id will always replace the existing ones
+	SectionID uint32
+
+	// Move's post id will always replace the existing ones
+	PostID uint64
```
RiccardoM, line 240:
```suggestion
	// ID of the subspace where the post should be moved to
	SubspaceID uint64

	// ID of the section where the post should be moved to
	SectionID uint32

	// ID of the post within the new subspace
	PostID uint64
```

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,92 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+	if post.Author != msg.Owner {
+		return nil, errors.Wrapf(sdkerrors.ErrInvalidRequest, "you are not the owner of the post")
+	}
+
+	// Check the permission to move the post
+	if !k.HasPermission(ctx, msg.TargetSubspaceID, msg.TargetSectionID, msg.Owner, types.PermissionWrite) {
+		return nil, errors.Wrap(subspacestypes.ErrPermissionDenied, "you don't have write permission on the target section")
+	}
+
+	// Get the next post id
```
RiccardoM, line 532:
```suggestion
	// Get the post id within the new subspace
```

x/posts/types/models.go
```diff
@@ -449,6 +496,39 @@ func (a *Attachment) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
...
+// AttachmentMove contains data related to a attachment that can be updated after it has been moved.
+type AttachmentMove struct {
+	// Move's subspace id will always replace the existing ones
+	SubspaceID uint64
+
+	// Move's post id will always replace the existing ones
+	PostID uint64
+
+	// Move's id will always replace the existing ones
+	AttachmentID uint32
```
RiccardoM, line 508:
```suggestion
	// ID of the subspace inside which the attachment should be moved
	SubspaceID uint64

	// ID of the post to which the attachment should be associated within the new subspace 
	PostID uint64

	// ID of the attachment within the new subspace
	AttachmentID uint32
```

x/posts/keeper/msg_server.go
```diff
@@ -498,3 +498,92 @@ func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParam
...
+		newAttachmentID++
+		attachmentMove := types.NewAttachmentMove(msg.TargetSubspaceID, newPostID, newAttachmentID)
+		updatedAttachment := attachmentMove.Update(attachment)
+		k.SaveAttachment(ctx, updatedAttachment)
+
+		if types.IsActivePoll(updatedAttachment) {
+			k.InsertActivePollQueue(ctx, updatedAttachment)
+		}
+		return false
+	})
```
RiccardoM, line 564: Why is this being done? Why can't we preserve the attachment ids that were used inside the old post instead? They are unique to the post itself anyway. If we preserve them we have less writes (which means less gas used) and we can simply set the `NextAttachmentID` to be `max(post attachment ids) + 1` which is way simpler. This way we would also make it easier for external tools that could simply update the post id instead of having to delete the attachments and re-storing them as well

x/posts/types/models.go
```diff
@@ -584,6 +664,15 @@ func IsPoll(attachment Attachment) bool {
 	return ok
 }
 
+// IsActivePoll tells whether the given attachment represents a active poll or not
+func IsActivePoll(attachment Attachment) bool {
```
RiccardoM, line 668: Instead of relying on a check of `FinalTallyResulsts` I think it's best to check the storage instead, running through the list of active polls. This would make sure that if there is any bug with the setting of the tally results we do not end up with a bug here as well

x/posts/keeper/msg_server.go
dadamu: Agree with you, we should preserve the original attachment id is correct, and if anything wrong invariant will alert when simulation test. I was confused that we should reset id or apply the original id here, finally I decided to reset the id because it will be error if ids are wrongly set.

By the way, I don't think it brings less writes since it actually adds some keys for moved attachments then deleted the original ones. The storage side is performing the same operations like:
```go
store.Set(movedAttachmentKey, movedAttachment)
store.Delete(oldAttachmentKey, oldAttachment)
```

x/posts/keeper/msg_server.go
RiccardoM: Note to self: `DeletePost` calls `DeleteAttachment` which removes the poll from the active queue (if it was there).

RiccardoM (approved):
Great job! :100:

PR #1270, opened 2023-12-01
fix: register missing interfaces to x/profiles module codec

dadamu:
## Description

Closes: #XXXX

This PR fixes the registry with empty interfaces, it causes the `LinkChainAccountPacketData` can not be parsed properly with the error: 
```
unable to resolve type URL /cosmos.crypto.secp256k1.PubKey
unable to resolve type URL /desmos.profiles.v3.Bech32Address
```

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/profiles/types/codec.go
+++ b/x/profiles/types/codec.go
@@ -5,6 +5,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/codec/legacy"
 	"github.com/cosmos/cosmos-sdk/codec/types"
 	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/msgservice"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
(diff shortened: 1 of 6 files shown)


x/profiles/types/codec.go
```diff
@@ -83,18 +88,30 @@ func RegisterInterfaces(registry types.InterfaceRegistry) {
 	msgservice.RegisterMsgServiceDesc(registry, &_Msg_serviceDesc)
 }
 
+// ModuleCdc references the global x/profiles protobuf codec. Note, the codec should
+// ONLY be used in certain instances of tests and for JSON encoding.
+//
+// The actual codec used for serialization should be provided to x/profiles and
+// defined at the application level.
+func ModuleCdc() *codec.ProtoCodec {
```
RiccardoM, line 96: Why was this changed from a variable to a method? If we do this, each time we wall `ModuleCdc` we are de-facto re-registering all the interfaces and returning a new coded. Which might impact performance issues. Why wasn't this kept as a variable and an `init` method used instead if needed?

dadamu: The `init` of [models.pb.go ](https://github.com/desmos-labs/desmos/blob/master/x/profiles/types/models_chain_links.pb.go#L420) runs after the codec.go `init`. This causes 
error that the interfaces registration of `x/profiles` is before proto message name registration so the typeURL of `x/profiles` types are always `/`:
```
panic: concrete type *types.Base58Address has already been registered under typeURL /, cannot register *types.HexAddress under same typeURL. This usually means that there are conflicting modules registering different concrete types for a same interface implementation
```

dadamu: I realize that the current app codec is already proto codec which registered all the required modules, so we can use it directly without any problems instead of maintaining another one for `x/profiles`

RiccardoM (approved):

PR #1289, opened 2024-01-16
docs: ADR-024: Decentralized identifiers resolver

dadamu:
## Description

Closes: #XXXX
This PR writes the documentation of how to integrate DIDs registry to Desmos.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

## Summary by CodeRabbit

- **New Features**
	- Integrated Decentralized Identifiers (DIDs) into Desmos, enhancing identity management with the addition of a `DidDoc` query method for resolving Desmos Profiles into DID documents.


docs/architecture/adr-024-decentralized-identifiers-registry.md
```diff
@@ -0,0 +1,396 @@
...
+    
+    // External controller in DID syntax who can manage the verification method (optional).
+    // Documentation: https://w3id.org/security#controller
+    string controller = 3;
+
+    // The assertion that two or more DIDs or other types of UIR refer to the same DID.
+    // Documentation: https://www.w3.org/TR/did-core/#also-known-as
+    string also_known_as = 4;
+
+    // Keys of verification methods for verifying digital signature, must have at least one associated to Desmos account.
```
RiccardoM, line 60: > must have at least one associated to Desmos account.

What does this mean?

docs/architecture/adr-024-decentralized-identifiers-registry.md
```diff
@@ -0,0 +1,396 @@
...
+
+    // DeactivateDidDoc allows to deactivate DID document.
+    rpc DeactivateDidDoc(MsgsDeactivateDidDoc) returns(MsgsDeactivateDidDocResponse);
+
+    // EditVerificationMethod allows to edit an existing verification method inside DID document.
+    rpc EditVerificationMethod(MsgEditVerificationMethod) returns(MsgEditVerificationMethodResponse);
+}
+
+// MsgCreateDidDoc represents the message to be used to create a new DID document.
+message MsgCreateDidDoc {
```
RiccardoM, line 197: I see that inside this message there are some fields that contain some data that can ideally be faked by the user. For example I have the following questions:

1. How can we make sure that the data provided by the user is correct (e.g. the verification methods are all valid)? 
2. Should we allow the user to set the ID? Wouldn't it be better to generate this one instead? 
3. Should we allow the user to set the Controller ID?

docs/architecture/adr-024-decentralized-identifiers-registry.md
```diff
@@ -0,0 +1,396 @@
...
+
+### Types
+
+#### DID document
+
+`DidDoc` represents the DID document which includes full context with the standard published by W3C DID core.
+
+The store keys of DID document inside KVStore will be as follows:
+
+* 0x1 | {id} | {version_id} -> ProtocolBuffer(DidDoc)
```
RiccardoM, line 33: I was thinking that the only DID that we will use is going to be the one related to a profile, so all these data could be auto-filled by us when a user wants to query a profile as a DID, directly inside the query response. 

I don't believe we would want other methods to get a DID of a post or something else. So why do we need to store DIDs on the chain, and we cannot simlpy generate them on-the-fly when a user requests one using a particular query?

docs/architecture/adr-024-decentralized-identifiers-registry.md
dadamu: > How can we make sure that the data provided by the user is correct (e.g. the verification methods are all valid)?
We can only check the public keys is the valid format or not. On the other hand, if user provide the non-owned public key to these fields, then user can not provide the proper signature to verifier, it means that we can recognize that user does not have that DID document.

> Should we allow the user to set the ID? Wouldn't it be better to generate this one instead?
Yeah, you are right. Generating on chain would be better for this case. I will update the docs.

> Should we allow the user to set the Controller ID?
Controller is a spec that allowing other DID to update the document. It gives more flexibility to DID document control, like group control. To have the flexibility, I think we can have it.
https://www.w3.org/TR/did-core/#the-relationship-between-did-controllers-and-did-subjects

docs/architecture/adr-024-decentralized-identifiers-registry.md
dadamu: Do you mean that having a resolver translating to DID from Desmos address? By this way, we will lose a lot of information and flexibility defined by w3c, like `service`, `alsoKnownAs` and the possibility other keys to control DID.

For example, the auto generated DID document will be like:
```json
{
  "id": "did:desmos:<desmos-address>"
  "verificationMethods": [{ "id": "did:desmos:<desmos-address>#did-key", "type": "Secp256k1", publicKeyMultibase: "<base58-encoded-key>" }]
  "assertionMethod": ["did:desmos:<desmos-address>#did-key"],
  "authentication": ["did:desmos:<desmos-address>#did-key"],
  "capabilityInvocation": ["did:desmos:<desmos-address>#did-key"]
  "keyAgreement": ["did:desmos:<desmos-address>#did-key"]
}
```

RiccardoM: Do we need all the things that you have listed? 

1. Do we really need the user to allow specifying a `service` or `alsoKnownAs`? 
2. What would it mean for a Desmos account, which is strictly tied to a Desmos private key, that other keys can control the DID? Since the DID is representing the Desmos account/profile, and it's not technically possible to have other keys controlling such profile, what would it mean if a user set other keys to control such DID? Wouldn't it imply that the profile is controled by someone external and thus does not represent the identity anymore?

dadamu: > Do we really need the user to allow specifying a service or alsoKnownAs?

They are optional information defined by users for their custom application usage. For example, on the service field side, user can set a bank service where this DID document especially use. The application can read this service field to quickly recognize this DID document is for bank service. There is also a resolution query method which is `did:desmos:<id>#<service-name>`: https://www.w3.org/TR/did-core/#fragment

On the other hand, `alsoKnownAs` is also used in application situation, to define the DID document related on other external ID but no guaranteed. For instance, setting `paul_chen_tw` on X on this field, requiring another process to link ID on X and my DID document. So, I think alsoKnownAs field is not necessary and can be removed.

> What would it mean for a Desmos account, which is strictly tied to a Desmos private key, that other keys can control the DID? Since the DID is representing the Desmos account/profile, and it's not technically possible to have other keys controlling such profile, what would it mean if a user set other keys to control such DID? Wouldn't it imply that the profile is controled by someone external and thus does not represent the identity anymore?

DID designed by W3C allows users to authorize other public keys to control their DID document by adding the key into capabilityInvocation field. The DID identity is owned by the public key owner, so user should add the keys they really owns. So I don't think DID document for general purpose can be seen as a DID document.

dadamu: I think I am misunderstanding something. I thought what we will build is a general DID document registry like cheqd. After this discussion, I think your decision is resolving Desmos profile as DID document rather than building a registry. Do I understand correctly?

RiccardoM: Yes, exactly that was my idea. I don't think that having a generic DID registry might be good for Desmos. But resolving Desmos profiles as DIDs it might be indeed. I would like to ask @kwunyeung his thoughts though as I might be the one misinterpreting here.

kwunyeung: @RiccardoM what you have mentioned is the same as what I originally thought of implementing DID. This feature is to provide a web standard compatible representation of Desmos Profile. We can generate all data of the profile as the DID structure which can be easily adapted by any applications that are going to make use of the DID w3c standardiaztion.

Even for fields like `service` and `alsoKnownAs` should be generated from what data has been stored on chain with the `chain-link` or `app-link`. E.g., we may see

`did:ens:forbole`
`did:ethr:0x1234123123123123123123`

The applications that supports those DIDs should support those resolvers from their ends.

It's still early but in the long run, the `service` field could be useful. E.g., a university is providing a service for providing educational qualification proofs to an identity and the service is registered on Desmos. If that specific DID is linked with the Desmos Profile, then in the Desmos Profile DID, we can have a service that is pointing to that DID with the qualification verification service.

dadamu: @kwunyeung After the discussion, I now understand that our DID will tie to the user profile in Desmos, in the other word, the starting point is Desmos profile. 
I agree with we can resolve `chain-link` and `application-link` in `alsoKnownAs` field. However, `service` may not be useful since it is a hint by user that tells where the DID can be used.  In our case, our DID should be use for all the apps inside Desmos ecosystem.

dadamu:
@RiccardoM Updated ADR to DID resolver from DID registry.

docs/architecture/adr-024-decentralized-identifiers-resolver.md
```diff
@@ -0,0 +1,143 @@
+# ADR 024: Decentralized Identifiers resolver
+
+## Changelog
+
+- Jan 16th, 2024: First draft;
+- Jan 23th, 2024: First review;
+
+## Status
+
+PROPOSED
```
RiccardoM, line 10:
```suggestion
ACCEPTED Not Implemented
```

RiccardoM (commented):
Looks great to me. Clear and effective :100: We can accept it and start with the implementation.

